PR #13, opened 2017-05-20
fix: Clean build directory before running build script

rkostrzewski:
Build directory is not cleaned before each successive build. This causes redundant file leftovers between builds.

As title describes - this fixes that by using rimraf via prebuild hook in created package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,6 +117,7 @@
     "raw-loader": "^0.5.1",
     "recursive-copy": "^2.0.6",
     "replace-bundle-webpack-plugin": "^1.0.0",
+    "rimraf": "^2.6.1",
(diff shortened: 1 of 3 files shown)


developit:
Thought: what about cleaning the directory in `src/commands/build.js`? Can still use rimraf, just that way `preact build` handles it for you without the extra dep?

rkostrzewski:
Makes sense - I always rant about lots of dependencies yet I always end up hoarding them 😄 . Another benefit is that when ```build``` changes to ```dist``` there won't be a problem.

I've also changed it so that only for production builds the dir is cleaned - might help those with slow HDDs in watch mode.

developit:
Merging - I might move this into the `build` command to simplify things but otherwise we're good to go!

PR #14, opened 2017-05-20
fix: Throw error for unknown command

rkostrzewski:
Currenty runing ```preact whatever``` won't report error. This is kinda unexpected behaviour and might cause some harm, e.g. in CI environments where script like ```preact built``` will return exit code 0.

This PR fixes that. I've also took liberty in changing default test script (coz ```preact test``` won't work now and people might get confused).

Pictures coz a picture is worth a thousand words.
Before:
screen shot 2017-05-20 at 16 17 50

After:
screen shot 2017-05-20 at 16 17 53

### DIFF
--- a/src/commands/create.js
+++ b/src/commands/create.js
@@ -92,7 +92,7 @@ export default asyncCommand({
 			build: 'preact build',
 			serve: 'preact build && preact serve',
 			dev: 'preact watch',
-			test: 'eslint src && echo "Error: no test specified" && exit 1'
+			test: 'eslint src && preact test'
 		};
 
 		pkg.eslintConfig = {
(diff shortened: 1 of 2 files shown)


developit:
Wow, thanks a ton @rkostrzewski, this is great!

src/commands/create.js
```diff
@@ -92,7 +92,7 @@ export default asyncCommand({
 			build: 'preact build',
 			serve: 'preact build && preact serve',
 			dev: 'preact watch',
-			test: 'eslint src && preact test'
+			test: 'eslint src && echo "Error: no test specified" && exit 1'
```
developit: This one we actually probably need to keep as `preact test` - that command isn't available yet, but it will just spawn a preconfigured jest. I wanted to get the setup into generated package.json's since we can't retroactively update them later. 

Would you be all right removing this one line? I think we can get the `test` command released really soon.

rkostrzewski: There will be ```preact test```? ❤️

PR #19, opened 2017-05-20
Install eslint-config-synacor as devDependency when running create

### LINKED ISSUE #18: Is it intentional that the eslint-config-synacor is referenced in the package.json but not in devDependencies?
Visual Code throws an error if an eslint-config is not present in the node_modules directory.

maoberlehner:
Fixes #18


developit:
Awesome. I'm also open to other suggestions for lint configs if people want them.

PR #20, opened 2017-05-20
Add NamedModulesPlugin for Dev

### LINKED ISSUE #6: Consider using the NamedModulesPlugin for module names.
Consider using the NamedModulesPlugin for module names. This option lists names of the modules that are hot reloaded vs the ids as displayed currently:

image

lukeed:
Fixes #6 

Also pulled `env.production` to a `let`... cuz that's how you roll.

### DIFF
--- a/src/lib/webpack-config.js
+++ b/src/lib/webpack-config.js
@@ -261,11 +261,6 @@ export default env => {
 			}
 		}),
 
-		// dev-only plugins
-		!isProd && addPlugins([
-			new webpack.NamedModulesPlugin()
-		]),
-
 		// copy any static files
 		addPlugins([
 			new CopyWebpackPlugin([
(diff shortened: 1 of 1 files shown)


src/lib/webpack-config.js
```diff
@@ -260,6 +261,11 @@ export default env => {
 			}
 		}),
 
+		// dev-only plugins
```
developit: thought: maybe drop this into the [`development()` group](https://github.com/developit/preact-cli/blob/2ed133bd9c5fa92b5ee4375628fe6c67b1749775/src/lib/webpack-config.js#L326) instead?

lukeed: Ah, right. My hesitation was that it wasn't modifying `plugins` but, ofc, that's what `addPlugins` is doing 😅

PR #21, opened 2017-05-20
Allow preinstall of plugins via --less / --sass

developit:


developit:
ah - just realized these should really be devDependencies.

PR #22, opened 2017-05-20
Fix `test:*` scripts

lukeed:
I have to leave my desk, but I'll get back to this later.

The `--cwd` flag was throwing errors on all scripts. Babel did not accept it as a valid argument.

As per @developit's suggestion, I tried passing it behind a `--` stop, but doing so set `env.src` equal to `'--cwd'`, which ofc broke the [`src` helper](https://github.com/developit/preact-cli/blob/master/src/lib/webpack-config.js#L42).

Currently, the `test:watch` script now works. However, all other commands fail because `test:build` fails with:

```
Your current PostCSS version is 5.2.17, but postcss-modules-resolve-imports uses 6.0.1. Perhaps this is the source of the error below.

  Total precache size is about 173 kB for 15 resources.

Template execution failed: TypeError: Cannot set property 'parent' of undefined
  TypeError: Cannot set property 'parent' of undefined

  - container.es6:442 Root.removeChild
    [preact-cli]/[postcss]/lib/container.es6:442:9

  - root.es6:29 Root.removeChild
    [preact-cli]/[postcss]/lib/root.es6:29:16

  - container.es6:611
    [preact-cli]/[postcss-modules-resolve-imports]/[postcss]/lib/container.es6:611:38

  - Array.map

  - container.es6:608 Root.normalize
    [preact-cli]/[postcss-modules-resolve-imports]/[postcss]/lib/container.es6:608:31

  - root.es6:32 Root.normalize
    [preact-cli]/[postcss-modules-resolve-imports]/[postcss]/lib/root.es6:32:21

  - container.es6:315 Root.append
    [preact-cli]/[postcss-modules-resolve-imports]/[postcss]/lib/container.es6:315:30

  - resolveDeps.js:98 resolveDeps
    [preact-cli]/[postcss-modules-resolve-imports]/lib/resolveDeps.js:98:12

  - sync.js:38 resolveImports
    [preact-cli]/[postcss-modules-resolve-imports]/sync.js:38:5

  - lazy-result.es6:321 LazyResult.run
    [preact-cli]/[postcss]/lib/lazy-result.es6:321:20

  - lazy-result.es6:307 LazyResult.sync
    [preact-cli]/[postcss]/lib/lazy-result.es6:307:32

  - lazy-result.es6:157 LazyResult.warnings
    [preact-cli]/[postcss]/lib/lazy-result.es6:157:21

  - index.js:99 fetch
    [preact-cli]/[css-modules-require-hook]/lib/index.js:99:16

  - index.js:134
    [preact-cli]/[css-modules-require-hook]/lib/index.js:134:14

  - attachHook.js:15 Object.cssModulesHook [as .css]
    [preact-cli]/[css-modules-require-hook]/lib/attachHook.js:15:20

  - module.js:20 require
    internal/module.js:20:19

  - index.js:1 Object.<anonymous>
    /Users/lukee/repos/oss/preact/preact-cli/examples/root/index.js:1:1
```

Following the suggestion (at the top), `postcss-less` imports `5.2.x` which causes the mismatch.

---

I'll get back to this later tonight if no one else already knows the solution!


developit:
this is weird - it's working locally for me. the `--cwd` flag is supposed to represent the equivalent of a repo root, which is slightly different than `src`, which is the entrypoint.  There must be a way to have cwd be passed to the app rather than babel itself... that's what it's been doing on my machine at least.

lukeed:
Right? It looked like it should have worked as it was, but I fresh-cloned three times and was the same every time. Have you done a fresh install of all deps recently? I can only guess that a package patch-bump introduced a regression of some kind.

developit:
I'll check, was thinking the same thing.  I don't dare attempt an `npm install` over this wifi though :P

rkostrzewski:
FYI: Getting same error using yarn to install dependencies. Basically running
```
rm -rf node_modules
yarn
```
creates same error.

developit:
thinking it's related to module flattening then.

lukeed:
Maybe. But before you reinstall everything, can you shrink wrap your current setup and send that to me? I don't mind chasing down bumped dependencies to see what's changed.

rkostrzewski:
Ok, so I've run docker images installing creating an app and running build (both npm & yarn). Sometimes npm fails and sometimes yarn doesn't fail. Latest npm ```v4.6.1``` and latest yarn ```v0.24.4``` were used to test this.

The issue occurs because *postcss-modules-resolve-imports* has peer dependency on *postcss* set to *>= 5* and it doesn't work with *postcss@^6.0.0*. Some packages import *postcss@^5.0.0* others *postcss@^6.0.0*. Now here's the funky part - it seems that both yarn and npm are nondeterministic and different versions are placed in different places.

Anyway I've createa a PR to fix this at *postcss-modules-resolve-imports* css-modules/postcss-modules-resolve-imports#17

P.S.
Running ```test:build``` on the branch does not fail for me.

lukeed:
All scripts now work thanks to @rkostrzewski's change in #31 & @developit's latest `master` changes.

developit:
@rkostrzewski you're a genius

PR #25, opened 2017-05-21
Minor tweaks to our readme

addyosmani:
Some minor improvements to our readme now that we've launched:

- Break up SW into two words and link our to a primer for those that are new to the feature
- Link up `async!` to an example of where it is showed in the examples directory
- Expand SSR


developit:
good calls!

PR #31, opened 2017-05-21
feature: Installing dependencies with yarn

rkostrzewski:
Changes ```npm install``` to ```yarn add``` when creating app & yarn is available as global command (similar to how create-react-app works). Faster install, progressive enhancement, yada, yada, yada. 

Your next Preact PWA starts in **less than** 30 seconds. :wink:

Cross-platform compatibility verified on MacOS & Windows 10.


rkostrzewski:
Good call.

Also npm-install-loader possibly could use yarn not end up with missing dependencies in yarn.lock file.

src/lib/run-webpack.js
```diff
@@ -66,13 +66,12 @@ export function writeJsonStats(stats) {
 	jsonStats.modules.forEach(normalizeModule)
 	jsonStats.chunks.forEach(c => c.modules.forEach(normalizeModule))
 
-	return fs.writeFile(outputPath, JSON.stringify(jsonStats))
```
rkostrzewski: Sneaky refactor 👣

rkostrzewski:
@lukeed removed ```npm``` but this should still be on hold as [postcss-modules-resolve-imports](https://github.com/css-modules/postcss-modules-resolve-imports) has just upgraded to *postcss@6* and yarn installs it as peer dependency and #22 happens all over again. 

I've looked into ```npm-install-loader``` and I have no idea what does it do.
Firstly, it does not install dependencies (aka other loaders) when other are not present and webpack returns error - I guess that's why #21 was made.
Furthermore, returning early from loader with error doesn't break the build & isn't shown earlier. I've tried following:
```
let callback = this.async();
callback("Hey I'm an error in a loader!")
```

What's the point of the loader?

package.json
```diff
@@ -94,7 +94,7 @@
     "babel-register": "^6.24.1",
     "copy-webpack-plugin": "^4.0.1",
     "cross-spawn-promise": "^0.10.1",
-    "css-modules-require-hook": "^4.1.0-beta",
+    "css-modules-require-hook": "^4.0.6",
```
rkostrzewski: This is the culprit of #22 and issues with yarn. I don't know why beta version was used - last stable version works fine for prerendering (at least for the examples). Don't know if that's ok.

lukeed: Ooh cool. I'll check this out!

thangngoc89:
I finished review this PR, read several articles on hacker news and it still installing devDeps. So +1 +1

package.json
developit: heads up - i merged @lukeed's PR without seeing this first.

src/commands/create.js
```diff
@@ -83,7 +84,7 @@ export default asyncCommand({
 
 		spinner.text = 'Initializing project';
 
-		await npm(target, ['init', '-y']);
+		await spawn('npm', ['init', '-y'], { cwd: target, stdio: 'ignore' })
```
developit: lol this is what it used to be 😆

thangngoc89: @developit yes it does

src/lib/run-webpack.js
developit: well-earned, I'd say ;)

src/commands/create.js
```diff
@@ -150,5 +149,27 @@ export default asyncCommand({
...
+	const isDev = env === 'dev' ? true : false
+	const isYarnAvailable = await isCommandAvailable('yarn')
 
-const npm = (cwd, args) => spawn('npm', args, { cwd, stdio: 'ignore' });
+	if(isYarnAvailable) {
+		const args = ['add']
+		if(isDev) {
+			args.push('-D')
+		}
+
```
developit: We can slim this down by moving the `.filter(Boolean)` into install, but I'll just slip that in post-merge :)

src/commands/create.js
developit: future thing to do might be to move the `which` check up and use yarn here too.  For now anyone running yarn would have npm installed anyway.

developit:
@rkostrzewski regarding `npm-install-loader` - it's less automatic than you're thinking.  All it does is install missing dependencies from a manually specified list before continuing to the next loader.  If the deps are installed it's a no-op, otherwise it will shell out to install them and then continue.

developit:
@rkostrzewski should we move `install()` into its own module and import that for use in `npm-install-loader`?  I kept that loader internal for changes like these - we could rename it to `dependency-install-loader` and have it use npm/yarn same as this PR does.

rkostrzewski:
@developit I don't think it works at all. The less example might work because less-loader is included in preact-cli deps.
Running:
```
preact create preact-app && cd preact-app
mv ./style/index.css ./style/index.scss
npm run build
```
Yields:
```

> preact-app@1.0.0 build /Users/rkostrzewski/GitHub/preact-app
> preact build

./index.js
Module not found: Error: Can't resolve 'sass-loader' in '/Users/rkostrzewski/GitHub/preact-app'
resolve 'sass-loader' in '/Users/rkostrzewski/GitHub/preact-app'
  Parsed request is a module
  using description file: /Users/rkostrzewski/GitHub/preact-app/package.json (relative path: .)
  after using description file: /Users/rkostrzewski/GitHub/preact-app/package.json (relative path: .)
    resolve as module
      looking for modules in /Users/rkostrzewski/GitHub/preact-app/node_modules/preact-cli/node_modules
        using description file: /Users/rkostrzewski/GitHub/preact-app/node_modules/preact-cli/package.json (relative path: ./node_modules)
        after using description file: /Users/rkostrzewski/GitHub/preact-app/node_modules/preact-cli/package.json (relative path: ./node_modules)
          using description file: /Users/rkostrzewski/GitHub/preact-app/node_modules/preact-cli/package.json (relative path: ./node_modules/sass-loader)
            as directory
              /Users/rkostrzewski/GitHub/preact-app/node_modules/preact-cli/node_modules/sass-loader doesn't exist
            no extension
              /Users/rkostrzewski/GitHub/preact-app/node_modules/preact-cli/node_modules/sass-loader doesn't exist
            .js
              /Users/rkostrzewski/GitHub/preact-app/node_modules/preact-cli/node_modules/sass-loader.js doesn't exist
            .json
              /Users/rkostrzewski/GitHub/preact-app/node_modules/preact-cli/node_modules/sass-loader.json doesn't exist
      looking for modules in /Users/rkostrzewski/GitHub/preact-app/node_modules
        using description file: /Users/rkostrzewski/GitHub/preact-app/package.json (relative path: ./node_modules)
        after using description file: /Users/rkostrzewski/GitHub/preact-app/package.json (relative path: ./node_modules)
          using description file: /Users/rkostrzewski/GitHub/preact-app/package.json (relative path: ./node_modules/sass-loader)
            as directory
              /Users/rkostrzewski/GitHub/preact-app/node_modules/sass-loader doesn't exist
            no extension
              /Users/rkostrzewski/GitHub/preact-app/node_modules/sass-loader doesn't exist
            .js
              /Users/rkostrzewski/GitHub/preact-app/node_modules/sass-loader.js doesn't exist
            .json
              /Users/rkostrzewski/GitHub/preact-app/node_modules/sass-loader.json doesn't exist
[/Users/rkostrzewski/GitHub/preact-app/node_modules/preact-cli/node_modules/sass-loader]
[/Users/rkostrzewski/GitHub/preact-app/node_modules/preact-cli/node_modules/sass-loader]
[/Users/rkostrzewski/GitHub/preact-app/node_modules/preact-cli/node_modules/sass-loader.js]
[/Users/rkostrzewski/GitHub/preact-app/node_modules/preact-cli/node_modules/sass-loader.json]
[/Users/rkostrzewski/GitHub/preact-app/node_modules/sass-loader]
[/Users/rkostrzewski/GitHub/preact-app/node_modules/sass-loader]
[/Users/rkostrzewski/GitHub/preact-app/node_modules/sass-loader.js]
[/Users/rkostrzewski/GitHub/preact-app/node_modules/sass-loader.json]
 @ ./index.js 2:0-17
 @ ./~/preact-cli/lib/lib/entry.js
 @ multi ./~/preact-cli/lib/lib/entry
```

Even tough sass-loader & node-sass are included in [npm-install-loader config](https://github.com/developit/preact-cli/blob/master/src/lib/webpack-config.js#L162). That's why I was asking.

I'll try to fix it along while adding yarn to it.

rkostrzewski:
@developit Yeah, I think it won't work and if I'm not wrong there's no way to fix this.

Long story short, webpack verifies if all loaders are present before attempting to compile files. This happens in [NormalModuleFactory](https://webpack.github.io/docs/plugins.html#normal-module-factory) and occurs before compiler starts compiling & any loaders are run ([source](https://webpack.github.io/docs/plugins.html#normal-module-factory) + empirical tests). Seemed like a cool feature tough :cry:

developit:
Darn. I had figured pitching phase would circumvent that.

rkostrzewski:
Yeah, proxying the loader would work for normal loaders. However it would be hard to make it work with pitch loaders and the solution looks very hacky.

hassanbazzi:
@rkostrzewski I think you can just fix that in webpack by passing a `resolveLoader` object. I had the same issue happen to me before

rkostrzewski:
@hassanbazzi I don't quite get how to use `resolveLoader` in this scenario. If I alias loader to another module which exists won't that be used by webpack later on?

lukeed:
`rkostrzewski dismissed lukeed’s review` I guess now we're even 😆

reznord:
@rkostrzewski Travis tests are failing because of an eslint error. Can I go ahead and fix that error?
It is because `const npm = ...` [here](https://github.com/rkostrzewski/preact-cli/blob/a6bcbecd3425ac224a62c6459a3f6580728eb9e0/src/commands/create.js#L173) is not being used anywhere !!

But there is this error - https://travis-ci.org/developit/preact-cli/jobs/247829570#L1925-L1926 which I didn't understand.

What is this about?

rkostrzewski:
@reznord sure go ahead. As for the other error I think there's some problem with async commands. I think about spiking the tesrs with console.log to determine where the error is (seems like output folder doesn't exist sometimes on first test)

rkostrzewski:
Also @lukeed has proposed installing with tarn using a switch as new npm has same capabilities as yarn with which I totally agree.

thangngoc89:
> as new npm has same capabilities as yarn 
@rkostrzewski I'm not agree with this. Yarn offers much more besides speed and lock file. IMHO, if someone installed yarn it mean that they want to use it.

How about a interactive question to ask which dependency manager to use?

reznord:
@rkostrzewski you have to make the changes, since it is a branch in your local repo :P

reznord:
> Also @lukeed has proposed installing with tarn using a switch as new npm has same capabilities as yarn with which I totally agree.

Well, consider this case where you have enabled `npm install` for running tests. That would just take an eternity to install the node_modules. Considering the same situation with yarn, it caches all the packages and it hardly takes a minute to install the packages.

thangngoc89:
> Well, consider this case where you have enabled npm install for running tests. That would just take an eternity to install the node_modules. Considering the same situation with yarn, it caches all the packages and it hardly takes a minute to install the packages.

@reznord try npm@5 ;)

reznord:
@thangngoc89 I did try, I have to say that it is faster than previous versions but still not close to yarn.

thangngoc89:
@reznord It's faster than yarn in my exeperience. Maybe your npm@5 global cache is empty? But anyway, see my comment above for my opinion

lukeed:
Chiming in: Yes, I think `yarn` usage should _not_ be used by default if it is found. Suppose @thangngoc89 had once been a Yarn user, switched to `npm@5` but forgot to uninstall Yarn. With this setup, we would find his (old,outdated) Yarn binary and use it.

We should use a `--yarn` flag because (1) it's technically not an official package manager, (2) it may not always be wanted, (3) presence of a flag is faster & easier to check, plus (4) when the flag _is_ present, we don't have to make any assumptions about whether or not to use it.

Just my 2¢ 😉

thangngoc89:
@lukeed That seems ok to me

reznord:
Agreed on this !!

developit:
Oooh very good point @lukeed, I'm for the `--yarn` flag if we want to support installing via it.

rkostrzewski:
@reznord what should I do with this in light of #155? Should I close this?

lukeed:
@rkostrzewski Unless I'm hugely mistaken, it doesn't look like that PR is going with a `--yarn` flag either.

rkostrzewski:
@lukeed yeah I've meant continuing as in developing `--yarn` flag there

reznord:
A fresh new PR available for `yarn support` #225 Closing this.

PR #35, opened 2017-05-21
initialize new projects with `git init` if git present

### LINKED ISSUE #28: No git repo initiated with project
When creating a new project no git repo was initiated at all. Not sure what would be the standard, but would be nice if a git repo was started, with a .gitignore setup inside as well.

fivetanley:
fixes #28 
i may have also extracted that `trimLeft` functionality


rkostrzewski:
I would add several more patterns to gitignore:
- ignoring logs:
```
*.log
npm-debug.log*
yarn-debug.log*
yarn-error.log*
```
- ignoring coverage (preact-cli will be using jest AFAIK which uses istanbul)
```
coverage
```

The [default gitignore for node](https://github.com/rkostrzewski/gitignore-test/blob/master/.gitignore) might have some more goodies but frankly I haven't seen half of those files in any node project.

lukeed:
^^ With `*.log` in place, no need for these:

```
npm-debug.log*
yarn-debug.log*
yarn-error.log*
```

rkostrzewski:
I remember npm generating timestamps in log file names but it doesn't do so anymore (at least I don't remember how it is done) so probably they are not needed. :)

src/commands/create.js
```diff
@@ -146,9 +149,31 @@ export default asyncCommand({
...
 })
 
+const trimLeft = (string) => string.trim().replace(/^\t+/gm, '');
 
 const npm = (cwd, args) => spawn('npm', args, { cwd, stdio: 'ignore' });
+
+// Initializes the folder using `git init` and a proper `.gitignore` file
+// if `git` is present in the $PATH.
+const initializeVersionControl = async function(target) {
+	if (which.sync('git')) {
```
rkostrzewski: which.sync throws an error if command is not found according to the docs. Is the exception swallowed?

developit: This should be using the async version of `which`.  Here's how:

```js
import promisify from 'es6-promisify';

// ...

async function initializeVersionControl(target) {
  let git;
  try {
    git = await promisify(which)('git');
  }
  catch (err) {}
  if (!git) {
    // ... etc
  }
}
```

fivetanley: Why does this have to be async? It would throw in either case when using async/await.

src/commands/create.js
```diff
@@ -146,9 +149,31 @@ export default asyncCommand({
...
+		`) + '\n';
+		const gitignorePath = path.resolve(target, '.gitignore');
+		await fs.writeFile(gitignorePath, gitignore);
+
+		const cwd = target;
+
+		await spawn('git', ['init'], { cwd });
+		await spawn('git', ['add', '-A'], { cwd });
+
+		const gitUser = 'Preact CLI<developit@users.noreply.github.com>';
```
developit: heh - I am wondering if this will spam me with notifications 😆

developit:
Overall this is great! Thanks for doing all the work, haha.

Only thing I'm keen to add to this aside from the little sync fix would be a `git` option that defaults to `true`, so someone can do `preact create --no-git` to avoid initializing a repo if they wish.  I'm happy to add that after merging if it's not of interest :)

developit:
@lukeed agreed, it would be nice to have `commandExists('yarn')` that just returns a Boolean without the try/catch.

rkostrzewski:
@developit @lukeed it's here - https://github.com/developit/preact-cli/pull/31/files#diff-3d03cf8df0979b4050f3a105cfd861c5R168 .
We can do this after merges.

developit:
wow I even subconsciously stole the name

src/commands/create.js
developit: Just to avoid blocking (nothing else is blocking in the CLI).

developit:
I think this makes more sense to be paired up with #56 as a minor or even major.

src/commands/create.js
reznord: of dear !! That sure will. Good luck with that !!

PR #48, opened 2017-05-23
Run whole repo through ImageOptim

krawaller:
Not a biggie since users shouldn't really ship with the default icons, but it doesn't hurt if they're properly optimized to set a good example either 😺

screen shot 2017-05-23 at 15 54 03


prateekbh:
LGTM

PR #52, opened 2017-05-24
Don't limit zooming

robdodson:
This'll help low vision users who need to zoom to read smaller print.


prateekbh:
@developit any specific reason why we limited it to 5?

robdodson:
we were chatting about it here: https://github.com/developit/preact-boilerplate/issues/192#issuecomment-303413164

> Nope! Seems arbitrary, could just remove it.

PR #54, opened 2017-05-24
adding 200.html

### LINKED ISSUE #53: 'Empty cache and hard reload' in Chrome crashes demo site
https://cli-demo.preactjs.com/profile/john 
I don't know if this is an issue/bug I am curious why it happens (I know think it is a bug because if I just opens this link in private mode it crashes):
(in Chrome, have the developer tools open and then right click on the reload button)
image

prateekbh:
this fixes #53 
this re routes all the other URLs to 200.html


developit:
Works for me. I wonder if we could just have specified `200.html` instead of `index.html` or added an option?

PR #58, opened 2017-05-24
Kiricon template

### LINKED ISSUE #51: Not Being able to edit or configure index.html
I don't know if this is intentional but, I think that would be nice to have the index.html open to the CLI user, this way we can configure some stuff.

Kiricon:
Fixes #51 

Places `template.html` in the root of the generated preact project. This file is read and used for building the project, however if missing then the template defaults to the build in template included in the npm package.

This will allow uses to modify language settings, meta tags and asynchronous JavaScript dependencies that exist outside of preact.

### DIFF
--- a/examples/empty/template.html
+++ b/examples/empty/template.html
@@ -1,0 +1,28 @@
+<!DOCTYPE html>
+<html lang="en">
+	<head>
+		<meta charset="utf-8">
+		<title><%= htmlWebpackPlugin.options.title %></title>
+		<meta name="viewport" content="width=device-width,initial-scale=1">
+		<meta name="mobile-web-app-capable" content="yes">
+		<meta name="apple-mobile-web-app-capable" content="yes">
+		<link rel="manifest" href="/manifest.json">
+		<% if (htmlWebpackPlugin.options.manifest.theme_color) { %>
+			<meta name="theme-color" content="<%= htmlWebpackPlugin.options.manifest.theme_color %>">
+		<% } %>
+		<% for (var chunk of webpack.chunks) { %>
+			<% for (var file of chunk.files) { %>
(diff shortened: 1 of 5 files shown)


developit:
I think it might be exposing a bit too much surface area to generate this into each project for everyone - it would make updating preact-cli very very hard since everyone would have custom templates right out of the box.  There was some discussion in #56 about allowing custom templates without requiring them (using the built-in one if none exists), which I'm more amenable to.  Perhaps we can use this in _one_ of the `preact create` templates in order to get people started if they really do want a custom index file?

thangngoc89:
@developit to prevent that from happening, perhaps we can provide hook like: `head.top`, `head.bottom`, `body.top`, `body.bottom`. That's pretty much everything you'll need

Kiricon:
@developit Continuing on your comments from PR #58 perhaps we should stead of including `template.html/index.html` in each create project, make it included in a new create/single create.
Despite that uses would still have the option to add a template file to their existing project if they would like.

This way the templates are not automatically included in each project, preventing everyone from having a custom template when they don't need it, and still giving people with an interest in having a template the ability to use one.

Making a single `create` have a template file would be an excellent example for user's looking to add that functionality to their project.

developit:
Open to options there.  We could also have preact-cli parse the HTML template and inject things appropriately (`html-webpack-plugin` does this internally)

rkostrzewski:
I think the best approach would be to create a template based on variables like this:
```
<html>
   <head>
        {{ head.top }}
        {{ injectStyles() }}
        {{ injectPreload() }}
        {{ head.bottom }}
   </head>
   <body>
        {{ body.content }}
        {{ injectScripts() }}
        {{ body.bottom }}
   </body>
</html>
```

This way we could provide easy to use customization of templates and use it on server side if we go that way.

thangngoc89:
Hey customize is in preact-cli for a while, closing this

Issue #62, opened 2017-05-27
Prerender fails when using React component


thangngoc89:
This is yet another SSR bug :D 

Reproduce: 
- Include any 3rd party component and run build with prerender
- Error message: 

```
Template execution failed: Error: Cannot find module 'react'
  Error: Cannot find module 'react'
  
  - prerender.js:43 Function.Module._resolveFilename
    [frontend]/[preact-cli]/lib/lib/prerender.js:43:21
  
  - module.js:20 require
    internal/module.js:20:19
  
  - index.jsx:1 Object.<anonymous>
    /home/khoa/web/tooth/frontend/components/Search/index.jsx:1:1
  
  - node.js:144 loader
    [frontend]/[babel-register]/lib/node.js:144:5
  
  - node.js:154 Object.require.extensions.(anonymous function) [as .jsx]
    [frontend]/[babel-register]/lib/node.js:154:7
  
  - module.js:20 require
    internal/module.js:20:19
  
  - index.js:7 Object.<anonymous>
    /home/khoa/web/tooth/frontend/index.js:7:1
  
  - node.js:144 loader
    [frontend]/[babel-register]/lib/node.js:144:5
  
  - node.js:154 Object.require.extensions.(anonymous function) [as .js]
    [frontend]/[babel-register]/lib/node.js:154:7
```

developit:
Ah weird, [module-alias](https://npm.im/module-alias) was actually never set up.  We need to add it to prerender.js.

kurtextrem:
couldn't be a webpack build be used for the prerender? That way webpack's alias could be used and code functionality wouldn't somewhat get duplicated.

developit:
Yes. Looks like @rkostrzewski is working on it because he's a genius.

kurtextrem:
Awesome! LMK if you need / want help.

(and just for the overview - maybe add the SSR label to this issue and to the PR?)

jangerhofer:
@developit: Do I understand correctly that the module-alias redirects `react` and `react-dom` includes to the corresponding `preact` packages, as described in the [`preact-compat`](https://github.com/developit/preact-compat) config?

I'm wondering because I'd like to use preact (and the convenience of the CLI!) with React packages, but `preact-compat` requires access to the `webpack.config`.

developit:
Label added :)

[...]

Issue #43, opened 2017-05-23
preact build error


nikhilbelchada:
Hello,
Getting SyntaxError on `preact build` while `npm start` works properly

```
Build [==================  ] 91% (1.9s) additional asset processing
  Total precache size is about 180 kB for 15 resources.

Template execution failed: SyntaxError: Unexpected token {
  SyntaxError: Unexpected token {
  
  - node.js:152 Object.require.extensions.(anonymous function) [as .js]
    [lib]/[preact-cli]/[babel-register]/lib/node.js:152:7
  
  - module.js:12 require
    internal/module.js:12:17
  
  - index.js:3 Object.<anonymous>
    [lib]/[preact-cli]/[postcss-modules-resolve-imports]/index.js:3:18
  
  - node.js:152 Object.require.extensions.(anonymous function) [as .js]
    [lib]/[preact-cli]/[babel-register]/lib/node.js:152:7
  
  - module.js:12 require
    internal/module.js:12:17
  
  - index.js:24 Object.<anonymous>
    [lib]/[preact-cli]/[css-modules-require-hook]/lib/index.js:24:22
  
  - node.js:152 Object.require.extensions.(anonymous function) [as .js]
    [lib]/[preact-cli]/[babel-register]/lib/node.js:152:7
```

developit:
@nikhilbelchada You can circumvent this using `preact build --no-prerender`, but I'd like to figure out why that's happening. Are you using less or sass?

juanpasolano:
I'm having the same issue. No less or sass but my error is 
```
bundle.js from UglifyJs
Unexpected token: punc ()) [bundle.js:5084,22]
```

`preact build --no-prerender` does not work either

```
$ node -v
v4.1.0
$ npm -v
4.4.1
```

nikhilbelchada:
@developit I am not using less or sass, just trying to run the default app.

Steps I followed:
preact create my-great-app
cd my-great-app/
npm start
npm run build [error..]

tiansijie:
seems like using the same node version of your preact-cli installed will solve the problem

developit:
@nikhilbelchada are you also running Node v4?  I think most of our testing has been done on 6+.

nikhilbelchada:
@developit yes i was on v4, now i updated to v6 and its working, so issue is specific to v4.

lukeed:
[...]

Issue #39, opened 2017-05-23
Error when require a PNG image


thangngoc89:
This only happens on build step, it doesn't happen on dev mode. I guests it's node/webpack environment

Stack trace:

```
Template execution failed: SyntaxError: /home/khoa/web/tooth/frontend/components/search/trans.png: Unexpected character '�' (1:0)
  SyntaxError: /home/khoa/web/tooth/frontend/components/search/trans.png: Unexpected character   '�' (1:0)
  > 1 | �PNG
      | ^
    2 | ?
    3 | [22m
    4 | IHDR  m[41  1mQ�^a[22m  2mtEXtSoftwareAdobe ImageReadyq�e  m<  qiTXtXML:com.adobe.xmp  <?xpacket begin=" " id="W5M0MpCehiHzreSzNTczkc9d"?> <x:xmpmeta xmlns:x="adobe:ns:meta/" x:xmptk  3m="Adobe XMP Core 5.5-c014 79.151481, 2013/03/13-12:09:15        ">  m <rdf:RDF xmlns:rdf="http:  //www.w3.org/1999/02/22-rdf-syntax-ns#"> <rdf:[33mDescription rdf:about="" xmlns:xmpMM=[39m"http://ns.adobe.com/xap/1.0/mm/" xmlns:stRef="http:/  /ns.adobe.com/xap/1.0/sType/ResourceRef#" xmlns:xmp="http://ns  .adobe.com/xap/1.0/" xmpMM:OriginalDocumentID="xmp.d  id:d3a7c906-7b1b-1749-be6c-7b7c0df0881c" xmpMM:DocumentID=  m"xmp.did:FC0A9B24DE7111E4A312C636C492E7C2" xmpMM:InstanceID="xmp.iid:FC0A9B23DE7111E4A312C636C492E7C2" xmp:CreatorTool  ="Adobe Photoshop CC (Windows)"> <xmpMM  9m:DerivedFrom stRef:instanceID="xmp.iid:f76eb8  cd-94ec-114f-a62f-7f548c5ac408" stRef:documentID="xmp.did:d3a7  c906-7b1b-1749-be6c-7b7c0df0881c"/> </rdf[39m:Description> </rdf:  9mRDF> </x:xmpmeta> <?xpacket end="r"?>t`��ID  ATxڼ��
  
  - index.js:4452 Parser.pp$5.raise
    [frontend]/[babylon]/lib/index.js:4452:13
  
[...]

Issue #42, opened 2017-05-23
Build error with preact-material-components


cedric-marcone:
Hello,

Linking a component from preact-material-components works like a charm in dev mode but when trying to build for production it fails.

**Steps to reproduce**

1) preact create test
2) cd test
3) npm install --save preact-material-components
4) edit routes/home/index.js
5) paste this code :

```js
import { h, Component } from 'preact'
import style from './style'

import FormField from 'preact-material-components/FormField'
import Radio from 'preact-material-components/Radio'

import 'preact-material-components/FormField/style.css'
import 'preact-material-components/Radio/style.css'


export default class Home extends Component {
  render() {
    return (
      <div class={style.home}>
        <h1>Home</h1>
        <FormField>
          <Radio id="r1" name='opts'></Radio>
          <label for="r1">Radio 1</label>
          <Radio id="r2" name='opts'></Radio>
          <label for="r2">Radio 2</label>
        </FormField>
      </div>
    )
  }
}
```

=> npm start : **OK**

=> npm run build : **Fails**

```
Template execution failed: SyntaxError: Unexpected token import
  Error: /Users/cedric/Desktop/test/node_modules/preact-material-components/FormField/index.js:1
  (function (exports, require, module, __filename, __dirname) { import { h } from "preact";
                                                                ^^^^^^
  SyntaxError: Unexpected token import
  
  - node.js:152 Object.require.extensions.(anonymous function) [as .js]
    [test]/[babel-register]/lib/node.js:152:7
  
  - module.js:20 require
    internal/module.js:20:19
  
  - index.js:4 Object.<anonymous>
    /Users/cedric/Desktop/test/routes/home/index.js:4:1
  
  - node.js:144 loader
    [test]/[babel-register]/lib/node.js:144:5
  
  - node.js:154 Object.require.extensions.(anonymous function) [as .js]
    [test]/[babel-register]/lib/node.js:154:7
  
  - module.js:20 require
    internal/module.js:20:19
  
  - index.js:7 Object.<anonymous>
    /Users/cedric/Desktop/test/index.js:7:1
  
  - node.js:144 loader
    [test]/[babel-register]/lib/node.js:144:5
```

Any idea ?

cedric-marcone:
Oops, BTW, 
preact --version
**1.0.1**

lukeed:
Thanks! 

[...]

PR #71, opened 2017-05-29
feature: Prerender using webpack

rkostrzewski:
Prerender fails when webpack specific stuff is in js code (described here #43, here #39, here #62 and possibly other places).

This fixes that by creating separate webpack bundle just for prerender with commonjs2 target.

This adds some overhead when building, but I couldn't use webpack's multiple compiler with two configs because some plugins failed. :cry:

WIP because I must fully test this & wanted to hear early feedback.


developit:
A thought I had:  could we produce a UMD bundle that would be usable in both the client and the server?

src/lib/run-webpack.js
```diff
@@ -32,11 +31,56 @@ export default (watch=false, config, onprogress) => new Promise( (resolve, rejec
...
+			target: 'node',
+			entry: path.resolve(env.cwd, env.src || 'src', 'index.js'),
+			plugins: config.plugins.filter(c => !(c.constructor && c.constructor.name === 'HtmlWebpackPlugin')),
+		});
+
+		newConfig.output.path = ssrOutputPath;
+		newConfig.output.libraryTarget = 'commonjs2';
+
+		let asyncLoaderIndex = newConfig.module.loaders
+			.findIndex(l => l.loader === path.resolve(__dirname, 'async-component-loader'));
```
developit: might be able to just alias this to a dummy / passthrough loader

developit:
If you don't mind I might also take a stab at this with the UMD approach.  Might not work at all so I don't want to waste your time :D

src/lib/run-webpack.js
rkostrzewski: Yeah it would be better + whenever someone uses ```import Component from 'async!./components/...'``` currently it won't work

rkostrzewski:
@developit go ahead! :smile:

Also, I'm thinking about using some in memory file system ([memfs](https://www.npmjs.com/package/memfs) for example) for webpack builds so that there won't be some magic folder called in npm module (+using normal fs breaks concurrency) :smile:

rkostrzewski:
@developit I didn't see your first comment.

> A thought I had: could we produce a UMD bundle that would be usable in both the client and the server? Definitely interested in any way to avoid the overhead of 2 concurrent builds. I think the closest prior art I know of here would be static-site-generator-webpack-plugin - maybe their technique would work. Or maybe a webpack bundle that can detect if its running under node/cjs and do a module.exports, otherwise performs a client render.

Technically possible. Passing two configs to webpack with two entries (sth like ```webpack([{ entry: 'index.js', rules: ... }, { entry: 'App.js', rules: ... }]```) doesn't add overhead and webpack doesn't seem to process same file twice (at least based on time it takes to compile). However, there is a problem with some plugins (just a guess) that are applied twice and throw errors. I want to investigate that and possibly exclude those plugins in SSR build.

rkostrzewski:
I've been able to pinpoint location of the erroneous plugin :smile:, this: goldhand/sw-precache-webpack-plugin#78 will enable us to use single webpack compiler pass & reduce overhead. :tada:

rkostrzewski:
All wrapped up & tested.
Chosen not to use any memory filesystem (```require``` doesn't work) and webpack creates ```ssr-build``` folder with bundle for SSR - this will be convenient when we will be working for server-side integration of rendering routes.

@developit if you haven't started yet & have some ideas for further work on this, feel free to work on this branch. :smile:

I'm thinking about writing some tests that would be able to verify the CLI works correctly (not only running builds, but running built bundles & watch mode).

lukeed:
@rkostrzewski I think we should bump the sw-precache dependency now that your PR there has been merged.

kurtextrem:
I haven't tried the code yet, but is polyfilling `history` and `location` enough? Where do the other things Preact uses (such as `document.createElement`) come from?

developit:
@kurtextrem During SSR, only `h()` is actually invoked from Preact. It knows to avoid any DOM mutation :)

src/lib/webpack/webpack-client-config.js
```diff
@@ -0,0 +1,188 @@
+import { resolve } from 'path';
```
developit: I wonder if there is a way to avoid duplicating this?

src/lib/webpack/webpack-base-config.js
```diff
@@ -57,26 +42,19 @@ export default env => {
...
-			filename: 'bundle.js',
-			chunkFilename: '[name].chunk.[chunkhash:5].js'
-		}),
 
 		customConfig({
 			resolve: {
 				modules: [
 					'node_modules',
-					resolve(__dirname, '../../node_modules')
+					resolve(__dirname, '../../../node_modules')
```
hassanbazzi: We should just be using NODE_PATH for this wherever possible. Actually, anywhere we have node_modules references for resolving, we should add NODE_PATH. That'll make us future proof.

ajoslin:
How can I help move this PR forward? I'm looking to use prerendering with Sass.

prateekbh:
just an fyi, i just debugged 
#42 is also being caused due to prerender, i am trying to dig deeper, will post if any finding comes up

hassanbazzi:
I just tried this branch on my site and I got the following error:

```
(undefined) Template execution failed: Error: Cannot find module '/Users/mypath/build/ssr-build/ssr-bundle.js'
(undefined)   Error: Cannot find module '/Users/mypath/build/ssr-build/ssr-bundle.js'
  
  - module.js:20 require
    internal/module.js:20:19
  
  - prerender.js:17 prerender
    [website]/[preact-cli]/lib/lib/webpack/prerender.js:17:10
  
  - webpack-client-config.js:156 Object.ssr
    [website]/[preact-cli]/lib/lib/webpack/webpack-client-config.js:156:54
  
  - template.html:109 EM6I.module.exports
    [website]/[preact-cli]/lib/resources/template.html:109:37
  
  - index.js:265 
    [website]/[html-webpack-plugin]/index.js:265:16
  
  - util.js:16 tryCatcher
    [website]/[html-webpack-plugin]/[bluebird]/js/release/util.js:16:23
  
  - promise.js:512 Promise._settlePromiseFromHandler
    [website]/[html-webpack-plugin]/[bluebird]/js/release/promise.js:512:31
  
  - promise.js:569 Promise._settlePromise
    [website]/[html-webpack-plugin]/[bluebird]/js/release/promise.js:569:18
  
  - promise.js:606 Promise._settlePromiseCtx
    [website]/[html-webpack-plugin]/[bluebird]/js/release/promise.js:606:10
  
  - async.js:138 Async._drainQueue
    [website]/[html-webpack-plugin]/[bluebird]/js/release/async.js:138:12
  
  - async.js:143 Async._drainQueues
    [website]/[html-webpack-plugin]/[bluebird]/js/release/async.js:143:10
  
  - async.js:17 Immediate.Async.drainQueues
    [website]/[html-webpack-plugin]/[bluebird]/js/release/async.js:17:14
```

hassanbazzi:
LGTM! I just went through and tested this super thoroughly. Also tried using the few config options differently for templates and such. :)

hassanbazzi:
teaaaam let's put this in master yo

ethanroday:
@hassanbazzi +1 but looks like tests are failing?

rkostrzewski:
Seems like snapshots have different sizes on CI.

src/lib/webpack/webpack-base-config.js
rkostrzewski: As a variable inside the file `const NODE_PATH = resolve(__dirname, '../../node_modules')` or via CLI?

developit: I thought `NODE_PATH` pointed to the global `node_modules` directory?

hassanbazzi: @developit Correct, which is what you want. It's basically used when people decide not to put node_modules inside their root dir but rather somewhere else.

developit: right but that assumes global - preact-cli installs itself as a local dep and this is looking there. I had wondered if there was a way to get to that path... maybe something like:

```js
path.resolve(require.resolve('preact-cli'), 'node_modules/foo')
```

tests/build.test.js
```diff
@@ -25,13 +25,22 @@ test('preact build - before', async () => {
...
+});
+
 test(`preact build - should use custom .babelrc.`, options, async t => {
 	// app with custom .babelrc enabling async functions
 	let app = await fromSubject('custom-babelrc');
 
 	// UglifyJS throws error when generator is encountered
-	// TODO: Remove '--no-prederender' once #71 is merged - babel-register problem
-	await build(app, ['--no-prerender']);
+	await build(app);
```
hassanbazzi: Nice! Got it to work with uglify?

hassanbazzi:
LGTM :)

PR #82, opened 2017-05-31
Promoting a little bit of design consistency 💅

davi-mbatista:
bkgdiff

Hi, this is just a little fix (although it bugs me a lot 😅 ) on the background color of profile / Me views. They have a different background color that is slightly darker than the home. So I just removed the style rule in favor design consistency.


developit:
I can't remember why I had them as different backgrounds. I think I was testing routing or something haha.

PR #83, opened 2017-05-31
add alias for components for more flexibility

hassanbazzi:


developit:
Hmm - I'm not sure it's advisable to have this be built-in and recommended by the templates.  Perhaps it'd be better to respect a [`_moduleAliases` field in the `package.json`](https://www.npmjs.com/package/module-alias#usage) if it exists?  That way it's configurable and not prescribed.

hassanbazzi:
I totally agree as I mentioned on Slack. If we think about the users of preact-cli, there will be a lot of people who want to do as least work as possible. That's why I added it.

developit:
Agreed. Also I'm kinda 50/50 here - I don't want everyone to end up with webpack config customizations, so if enough people need aliasing as a default then it should probably be baked in.  That said, if we can just bake it in via the package.json, that makes it nice and visible (and editable).

PR #86, opened 2017-06-01
fix: Build returns non-zero exit code for failed builds.

### LINKED ISSUE #61: `preact build` doesn't exit with proper code on error
`preact build` always exit with code = 0 where there are errors or not. And this is pretty annoying because you can't catch this on CI.

P/s: After a quick digging, a try ...catch block is missing here https://github.com/developit/preact-cli/blob/35fd84a86fe6b6c8abd60aeb721494771930c4ff/src/commands/build.js#L41 . 
I think it's the root of this evil

rkostrzewski:
This time I'm bringing even more conflicts on other branches 😭  Fixes #61


developit:
Looks good to me. Maybe one more set of eyes?

Issue #70, opened 2017-05-29
Custom server support


robinvdvleuten:
Instead of having a static production-like server when building for production. It would definitely be awesome to have some kind of "custom server" type of output like [next.js](https://github.com/zeit/next.js#custom-server-and-routing) has.

thangngoc89:
@robinvdvleuten I think preact-cli and next.js are 2 different beasts. You can serve the static build with any server. So no need for support from preact-cli. Just build your own :D

robinvdvleuten:
@thangngoc89 I am aware of that, but the preact-cli also has supports for prerendering. It would be nice if there is some sort of exposed API for this so that you still can rely on the webpack / babel configuration of this project :)

thangngoc89:
@robinvdvleuten I think the prerendering is done in build step. Check out `index.html`, you'll see the rendered code of your index route

developit:
If we end up with a UMD or CommonJS bundle output from something like #71, you could use preact-cli as an opaque compiler within a Node server.  I'm not sure it'd fit perfectly into `preact-cli` itself, but certainly I could see supporting something like this:

```js
import { h } from 'preact';
import express from 'express';
import { build } from 'preact-cli';
import renderToString from 'preact-render-to-string';

const app = build('./client', { format: 'umd' }).then( output => require(output.main) );

const app = express();
app.get( (req, res) => {
  app.then( App => {
    let body = renderToString(<App url={req.path} />);
    res.send(`<!doctype html><html lang="en">${body}`);
  });
});
app.listen(process.env.PORT);
```

robinvdvleuten:
@developit that's exactly the kind of interface I meant! I do not expect that preact-cli becomes some kind of prext.js, but it would be great if it provides some kind of interface / functionality to build this kind of tooling upon. The only thing the preact-cli should do extra I guess is to expose the build process. Is there any way I can help with this?

developit:
Any experimentation you have time to do would be appreciated.  The API we can solidify later, but I'd love to get some experiments done around returning a reusable "preact compiler" instance.

robinvdvleuten:
@developit I've created a PR for exposing a compiler instance, let's continue the discussion there.

PR #88, opened 2017-06-02
Expose compiler object responsible for compiling the preact application

robinvdvleuten:
As mentioned by #70, I started creating a reusable compiler instance responsible for building or watching a preact application. The commands are now using this compiler to generate their output. The compiler is exposed as main entry so one can use it in their own project like this;

```js
const createCompiler = require('preact-cli');

let compiler = createCompiler(/* any options can be passed here */);

// Compile the application and optionally serve it through the dev server by passing a bool.
await compiler.compile(true);

// Return a prerendered string for the given params
compiler.prerender(/* any params passed to prerender */)
```

I guess this can be seen as a first step in exposing an api to built upon; like exposing the prerender functionality to render compiled preact strings in a node.js app, replacing the dev server with dev middleware so you can mount the preact-cli on your own server, etc.

### DIFF
--- a/src/compiler.js
+++ b/src/compiler.js
@@ -1,0 +1,36 @@
+import { resolve } from 'path';
+import promisify from 'es6-promisify';
+import rimraf from 'rimraf';
+import prerender from './lib/prerender';
+import webpackConfig from './lib/webpack-config';
+import runWebpack, { showStats } from './lib/run-webpack';
+
+class Compiler {
+	constructor(env = {}) {
+		this.env = env;
+		this.config = webpackConfig(env);
+		this.stats = null;
+	}
+
+	async clean() {
+		let dest = resolve(this.env.cwd || process.cwd(), this.env.dest || 'build');
+		await promisify(rimraf)(dest);
+	}
+
+	async compile(watch = false) {
+		this.stats = await runWebpack(watch, this.config, showStats);
+		showStats(this.stats);
+	}
+
+	prerender(params) {
+		return prerender(this.env, params);
+	}
+
+	async stats() {
+		await writeJsonStats(stats);
+	}
+}
+
+export default env => {
+	return new Compiler(env);
+}
--- a/src/compiler.js
+++ b/src/compiler.js
@@ -3,34 +3,30 @@ import promisify from 'es6-promisify';
 import rimraf from 'rimraf';
 import prerender from './lib/prerender';
 import webpackConfig from './lib/webpack-config';
-import runWebpack, { showStats } from './lib/run-webpack';
-
-class Compiler {
-	constructor(env = {}) {
-		this.env = env;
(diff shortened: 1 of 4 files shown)


robinvdvleuten:
@developit did you manage to find some time to look at this PR? It would be great to have some sort of extendability built into this project

developit:
Not yet, no. Hopefully someone else can review before me!

src/compiler.js
```diff
@@ -0,0 +1,36 @@
+import { resolve } from 'path';
+import promisify from 'es6-promisify';
+import rimraf from 'rimraf';
+import prerender from './lib/prerender';
+import webpackConfig from './lib/webpack-config';
+import runWebpack, { showStats } from './lib/run-webpack';
+
+class Compiler {
```
developit: Curious about everyone's opinions on this, but I wondered if this could have been done more directly via a closure + exported object (just a POJO factory).  If everyone is good with classes I don't have a strong opinion here, just it's the first class into this codebase that isn't specifically for Webpack, so I thought I'd ask :)

fforres: (Can anyone make an opinion?)
The implementation looks awesome, but as a personal choice, I'd prefer non-class approaches 👍  
I guess it depends on the project... are classes used in `preact-cli`?
If so, it could be good to keep consistency with pojo/factory approach 😃

robinvdvleuten: yeah maybe a pojo is a better idea, I'll convert it!

src/compiler.js
```diff
@@ -0,0 +1,32 @@
...
+import prerender from './lib/prerender';
+import webpackConfig from './lib/webpack-config';
+import runWebpack, { showStats, writeJsonStats } from './lib/run-webpack';
+
+export default env => {
+	let config = webpackConfig(env)
+
+	return {
+		async clean() {
+			let dest = resolve(env.cwd || process.cwd(), env.dest || 'build');
```
rkostrzewski: In retrospective this could have been `config.output.path` - if this lands before #56 I'll change it there

ForsakenHarmony:
Is this still relevant with the ssr-bundle being produced now?

ForsakenHarmony:
given that there has been no activity I'm gonna close this

we're working towards modularising the cli which will also export the config

Download:
Why was this closed? 
I still cannot find a good answer to the question: 
How to take a preact-cli app and make it server-side rendering? 
I was hoping this PR was the answer...

lukeed:
@Download You can require the `build/ssr-build/ssr-bundle.js` file directly and pass in whatever URLs you want to as `props.url`.

It's the same approach the CLI is using internally to produce the "prerendered" outputs -- except that you can do it at runtime.

I have a full example with this approach & am using it on a couple sites: https://github.com/lukeed/preact-cli-ssr

PR #104, opened 2017-06-08
Changed —type option name to match CLI

davi-mbatista:
Reflecting the CLI change of 'full' to 'root' of --type options in the README.


developit:
good catch!

PR #106, opened 2017-06-10
tests: Preact CLI

rkostrzewski:
Basic tests for every command CLI exposes :tada:

Stuff used:
- Headless chrome (which requires Chrome > 59)
- Tape - Jest had problems with running Chrome, Ava had problems with spawning - tests are kinda slow without concurrency (~1m on my laptop).


rkostrzewski:
Wrong PR direction 😅

PR #107, opened 2017-06-10
test: Preact CLI tests

rkostrzewski:
Basic tests for every command CLI exposes :tada:

Stuff used:
- Headless chrome (which requires Chrome > 59)
- Tape - Jest had problems with running Chrome, Ava had problems with spawning - tests are kinda slow without concurrency (~1m on my laptop).

Stuff changed:
- Added option to `preact create` not to install dependencies


tests/create.test.js
```diff
@@ -0,0 +1,34 @@
...
+import { resolve } from 'path';
+import { create } from './lib/cli';
+import lsr from './lib/lsr';
+import { setup, clean } from './lib/output';
+
+const listTemplate = async dir => await lsr(resolve(__dirname, '../examples', dir), ['.gitkeep', 'package.json']);
+const listOutput = async dir => await lsr(dir, ['.gitkeep', 'package.json']);
+
+test('preact create - before', async () => {
+	await setup();
```
lukeed: I'd also prefer to move all the `setup()`s and `clean()`s into the test itself, since the presence of `async` will ensure that those prep/teardowns happen accordingly.

tests/lib/output.js
```diff
@@ -0,0 +1,16 @@
+import { resolve } from 'path';
+import rimraf from 'rimraf';
+import promisify from 'es6-promisify';
+import mkdirp from 'mkdirp';
+
+const rm = promisify(rimraf);
+
+export const outputPath = resolve(__dirname, '../output');
+
+export const setup = async () => {
```
lukeed: Both of these can just be:

```js
export const name = () => helper(foo);
```

since they are returning Promises & will be/are `await`ed in the tests.

rkostrzewski:
@lukeed I was afraid of `DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.`.  Since tape doesn't care about returned promises `async-test` takes care of that. 

`opts.timeout` looks good - I'll definitely use it!

tests/build.test.js
```diff
@@ -0,0 +1,105 @@
...
+		}
+	},
+	'polyfills.chunk.*.js': { size: 4066 },
+	'push-manifest.json': { size: 303 },
+	'favicon.ico': { size: 15086 },
+	'manifest.json': { size: 426 },
+	'sw.js': { size: 3905 }
+};
+
+const expectedOutputs = prepare({
```
rkostrzewski: Yeah, you're right - just being lazy 😛 I was thinking about some snapshots but the only package i found didn't work

tests/create.test.js
rkostrzewski: I separated them because it the 'main' test fails then they will be called for sure.

It doesn't matter for `create` or `build` that much (they could be handled inside tests and in `test.onFinish` hook), but stuff with chrome get's trickier as we must ensure service worker is unregistered after tests because on non-CI environment service workers could provide false positives (I've tried using profile flag to ensure clean state to no avail).

I thought that consistency between all tests would be easier to maintain & extend.

lukeed:
Where does that `DeprecationWarning` show up?

tests/lib/output.js
rkostrzewski: That's what you get for writing in language with native `async` & `await` for years. 😅

rkostrzewski:
It comes right out of node in console when async test fails using regular tape test.

lukeed:
Oh sorry, I meant a specific test as an example 😃

rkostrzewski:
Then any test hopefully never 😄

src/commands/create.js
```diff
@@ -112,34 +117,36 @@ export default asyncCommand({
 
 		spinner.text = 'Installing dev dependencies';
```
developit: super nitpick (feel free to ignore) - seems like the dev dep message and "done!" message should be inside the if here.

developit:
@rkostrzewski @lukeed think we're good to merge?  It'd be great to have these running on Travis.  To that end, I just enabled travis for the preact-cli repo.  I think updating this branch will trigger a build and we can check that everything is peachy there.

rkostrzewski:
@developit hold on I'll work on setting up travis on this branch - I've got a feeling it won't be so easy (I bet timeouts will have to be increased 😝 )

src/commands/create.js
lukeed: Oh, I misread the diff last time. Yeah, ^^ what I was trying to say last time. 😴

rkostrzewski:
...many commits later....

It's green! 🎉  😀 I've restarted the build several times to make sure it is reliable.

@lukeed assuming you mean sth like `test('Name', 500, async t =>` I'll leave it like it is - I think current approach with async test having same API as tape test is more versatile.

@developit I think we're ready to merge! 😀 

BTW we can install yarn on travis after yarn PR makes it way and use it to create app - maybe the `install` switch could be removed

developit:
That's awesome :D

src/commands/create.js
rkostrzewski: And I wanted to change it like that! Anyway, should be good now 😆

rkostrzewski:
| rkostrzewski dismissed stale reviews from @developit and @lukeed via 7800035 9 days ago
😢 any of you mind doing it again? :pray:

tests/serve.test.js
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	let { DOM, DOMDebugger } = chrome;
+	let { root: document} = await DOM.getDocument();
+	let a = await DOM.querySelector({ selector: 'a', nodeId: document.nodeId });
+	let { object } = await DOM.resolveNode({ nodeId: a.nodeId });
+	let { listeners } = await DOMDebugger.getEventListeners({ objectId: object.objectId });
+
+	if (!listeners.some(l => l.type === 'click')) {
+		await pageIsInteractive(chrome, retryCount - 1, retryInterval);
```
developit: I think this might be missing an `await sleep(retryInterval)`?

rkostrzewski: Preact sure is fast as it didn't ever fail on CI AFAIR. :smile: I'll add the delay with smaller interval

rkostrzewski: Done so [url](https://github.com/developit/preact-cli/pull/107/commits/59e1293508b12f7a4ec1ea2ae71148e486186310).

Is there any trick not to dismiss reviews when committing?

developit: not sure, maybe it's a setting..

developit: yup, just turned that thing off. sorry!

lukeed:
🙌 🕺

PR #111, opened 2017-06-11
Adding Support for video files on preact-cli

fforres:
I tried creating a fullscreen background video and found out preact-cli  did not webpack support for local video loading.
(This PR addresses that 😸)

I only added `.mp4`, `.webm` and `.ogv` formats


developit:
Perhaps we should use file-loader as a fallback for all unrecognized file formats?

lukeed:
@developit That makes sense to me 👍

PR #119, opened 2017-06-17
fix: Cors headers

rkostrzewski:
@developit this change allows only `https://localhost:port` origin to access served app.


src/commands/serve.js
```diff
@@ -179,11 +179,12 @@ const SERVERS = {
 			options.cwd = persistencePath('preact-cli');
 			process.stderr.write(`Falling back to shared directory + simplehttp2server.\n(dir: ${options.cwd})\n`);
 		}
+		let port = options.port || 8080;
 		return [
 			simplehttp2server,
-			'-cors', '*',
+			'-cors', `https://localhost:${port}`,
```
developit: Can we allow an `options.cors` that gets appended to this as a CSV?

rkostrzewski: Here you go. FYI simplehttp2server is able to add only one origin.

rkostrzewski:
@developit this is a breaking change - as it can break stuff on things like now.sh

developit:
@rkostrzewski maybe we should check if there is an env var we can watch for?

rkostrzewski:
@developit we can check for `NOW` variable on now.sh, but I don't really think it's worth the trouble as far more providers exist out there. Furthermore setting CORS to * even on hosting provider doesn't seem like a good idea.

reznord:
Looks like it passed all the tests. Can I merge this?

rkostrzewski:
I'd wait for @developit since IMHO this is a major version bump and maybe something smaller will come along.

developit:
I think I overestimated the breakage this would cause initially - really the only time someone would run into an issue is if they are trying to fetch assets from their preact-cli output on a page with a differing origin. I think that's a fairly small set of people - I've had such a usecase before, but I would expect to have to muck around with hostnames/headers a bit to make things work.

Should we wait for a bit of breakage to accumulate and make a 2.0.0, or merge and cross our fingers? haha

lukeed:
TBH, I'm not fully aware of the implications, but I have a feeling that (in development) that would be a tiny tiny subset of people. 

In production, I'd assume that the requesting origins would be known & specified... or just reverted back to `"*"`.

developit:
The last time I ran into this was proxying my initial bundle, but leaving it to reference its associated webpack assets from a CDN.  The thing is, I had to manually configure Webpack to set the public path to the CDN.

Maybe we could generate a list of allowed origins using some more information:

- the user's own ip (or maybe `192.*`?)
- any domain found in `publicPath`
- the `host` / `hosts` CLI arg
- `$HOSTNAME`

The combination of these would handle a whole bunch of cases, really leaving only a few where there's not enough information to determine safety.

rkostrzewski:
Simple http2server doesn't support multiple origins :cry:

developit:
(continuing from slack lol) When you suggested setting the header from `firebase.json` - do you know if that works with simplehttp2server?

PR #125, opened 2017-06-20
Respect "browserslist" config

### LINKED ISSUE #124: Honor browserslist in package.json for Babel, Autoprefixer, etc.
Rather than hard-code a list of browsers in babel and webpack configs, it would be great if the CLI just added a default `browserslist` entry in `package.json` that the user could then edit. For those of us developing for, say, UC Browser for Android, there's currently no recourse to get the full set of required prefixes other than forking this repo and altering the webpack config manually.

lukeed:
I never saw a way to write to `env.browsers` (within babel config) aside from a CLI string, which is very error prone as it passes thru the `split()`.

This will share a `browsers` config across Babel _and_ Autoprefixer. I don't believe there are any other configs...?

Lastly, "upgrades" the default Autoprefixer from `'last 2 versions'` to what Babel was specifying.

Closes #124

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,7 +7,7 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 
-[{*.json,.*rc,*.yml}]
+[{*.json,.*rc,*.yml,*.md}]
 indent_style = space
 indent_size = 2
 
--- a/README.md
+++ b/README.md
@@ -14,9 +14,9 @@
 - Monitor your bundle/chunk sizes with built-in tracking
 - Automatic app mounting, debug helpers & Hot Module Replacement
 - In just **4.5kb** you get a productive environment:
-	- [preact]
-	- [preact-router]
-	- 1.5kb of conditionally-loaded polyfills for [fetch](https://github.com/developit/unfetch) & [Promise](https://npm.im/promise-polyfill)
+  - [preact]
+  - [preact-router]
+  - 1.5kb of conditionally-loaded polyfills for [fetch](https://github.com/developit/unfetch) & [Promise](https://npm.im/promise-polyfill)
 
 
 ### Commands
@@ -79,8 +79,8 @@ $ preact serve
 
   --dir       Directory root to serve static files from.          [default: "build"]
   --cwd       The working directory in which to spawn a server.   [default: .]
-  --server    Which server to run, or "config" to produce a firebase config.      	
-  	      [options: "simplehttp2server", "superstatic", "config"] [default:"simplehttp2server"]
+  --server    Which server to run, or "config" to produce a firebase config.        
+          [options: "simplehttp2server", "superstatic", "config"] [default:"simplehttp2server"]
   --dest      Directory or filename where firebase.json should be written
               (used for --server config)                          [default: -]
   --port, -p  Port to start a server on                           [default: "8080"]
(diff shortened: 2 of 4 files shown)


src/lib/babel-config.js
```diff
@@ -4,12 +4,8 @@ export default (env, options={}) => ({
 		[require.resolve('babel-preset-env'), {
 			loose: true,
 			modules: options.modules || false,
+			browsers: options.browsers,
```
rkostrzewski: ❤️ options.browsers - babel-config.js will be still usable as a babel preset!

However it should be: `target: { browsers: [...] }` like [here](https://github.com/babel/babel-preset-env#babel-preset-env----) (I've wanted to check what this does - I still don't know, but at least I've noticed it is passed differently 😛 )

rkostrzewski:
I think this should be described it in a `readme` or some other form of docs.

src/lib/babel-config.js
lukeed: Oh yeah, nice catch! Looks like that should have always been `target.browsers` 👍 

Sorry, I'm confused by your first sentence. Not sure what you mean.

rkostrzewski: Just that I really dig that you pass browsers via options 😛 . That means I'll still be able to us babel-config.js as a babel preset in #56 and use it like so: ```presets: [path.resolve(..., 'babel-config.js', { browsers: ... } ]```

lukeed:
I'll add a README note as part of this PR.

Question: How do we want to allow devs to specify Babel `target`s? See [full usage](https://github.com/babel/babel-preset-env#export-with-various-targets).

Follow up questions here: https://github.com/developit/preact-cli/issues/124#issuecomment-309910052

src/lib/babel-config.js
lukeed: Oh, haha, great!

I'm thinking now, to answer my own question (below), that I need to respect any `"babel"` config inside the `package.json`.

With #56, how will users customize their babel options? I should account for that too.

rkostrzewski: Via .babelrc or package.json (or even by changing webpack babel-loader query). I don't think you have to worry about that at all because babel is so smart that it extends over this configuration, i.e. a `.babelrc`that looks like this ```{ plugins: ["my-awesome-plugin"] }``` will get everything applied from this file + 'my-awesome-plugin'.

In case of collisions stuff from .babelrc takes precedence over our internal settings (don't remember if options were merged tough, e.g. when `env` preset was declared stuff like `loose: true` was preserved or in this case browsers would be preserved)

README.md
```diff
@@ -99,6 +99,27 @@ npm run serve -- --server config
 # Copy your static files to a server!
 ```
 
+### Custom Configuration
+
+> **TL;DR** Currently in progress. See [#56](https://github.com/developit/preact-cli/pull/56)
+
+#### Browserlist
+
+You may customize your list of supported browser versions by declaring a [`"browerslist"`](https://github.com/ai/browserslist) key within your `package.json`. Changing these values will modify your JavaScript (via [`babel-preset-env`](https://github.com/babel/babel-preset-env#targetsbrowsers)) and your CSS (via [`autoprefixer`]()) output.
```
rkostrzewski: browerslist -> brow**s**erslist

lukeed: //facepalm 😅

rkostrzewski: You're on a roll! bro**s**werslist :rofl:

lukeed:
Confession: I _hate_ typing the word "browser" -- even though I know how to spell it, it rarely comes out right & typos blend in. Good thing I'm a front-end dev, eh?

README.md
```diff
@@ -99,6 +99,27 @@ npm run serve -- --server config
 # Copy your static files to a server!
 ```
 
+### Custom Configuration
+
+> **TL;DR** Currently in progress. See [#56](https://github.com/developit/preact-cli/pull/56)
+
+#### Browserslist
+
+You may customize your list of supported browser versions by declaring a [`"browserslist"`](https://github.com/ai/browserslist) key within your `package.json`. Changing these values will modify your JavaScript (via [`babel-preset-env`](https://github.com/babel/babel-preset-env#targetsbrowsers)) and your CSS (via [`autoprefixer`]()) output.
```
rkostrzewski: Minor: missing link to autoprefixer :smile:

lukeed: I'm tired.

lukeed:
Thanks @rkostrzewski 😆 

Pinging @developit for quick sign-off on README changes

.editorconfig
```diff
@@ -7,7 +7,7 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 
-[{*.json,.*rc,*.yml}]
+[{*.json,.*rc,*.yml,*.md}]
```
developit: weird this wasn't already in here

lukeed:
Is this ready to make its way in?

developit:
IMO yes. and @rkostrzewski reviewed previously but our settings killed it. merging.

PR #127, opened 2017-06-20
fix: Promise pollyfil loaded before webpack bundle

### LINKED ISSUE #101: Promise polyfill is not applied
Apps created via cli do not work on browsers without `Promise` implementation. For Windows users the easiest way to reproduce is to open https://clidemo.preactjs.com in IE11 and look into dev console.

This is problems occurs because of polyfills reside in separate chunks - which would be fine if webpack didn't use `Promise`s in chunk loading code 😆 (which it does and [states so in docs](https://webpack.js.org/guides/code-splitting-async/#promise-polyfill)).

I'm thinking we can go several ways:
- drop non ES2015 browsers support (which means dropping PhantomJS and possibly large part of potential cli users)
- load promise polyfill eagerly
- load promise polyfill via some custom JS code (dynamic script tag?)

**Steps to reproduce**
1. `preact create app`
2. `cd app`
3. `npm run serve`
4. Visit localhost:8080 in browser without ES6 support
5. Click any navigation link (e.g. Profile)

**Expected result**:
- Page is changed

**Actual result**:
- Nothing happens. Error `Promise is not defined` is seen in console

rkostrzewski:
Closes #101.

Tested in Chrome on desktop & mobile and IE 11. Chrome doesn't complain about document.write (ofc I've used `typeof fetch === "function"`). Rendering is not blocked in IE. :smile:


src/lib/webpack-config.js
```diff
@@ -433,6 +435,7 @@ const production = config => addPlugins([
 		minify: true,
 		stripPrefix: config.cwd,
 		staticFileGlobsIgnorePatterns: [
+			/polyfills\.js$/,
```
lukeed: Did you ignore this because all `serviceWorker`-compatible browsers don't need a `fetch` or `Promise` polyfill?

_(^ I'm not certain that that's 100% true)_

src/resources/template.html
```diff
@@ -24,5 +24,6 @@
 		<%= htmlWebpackPlugin.options.ssr({
 			url: '/'
 		}) %>
+		<script>"undefined"==typeof fetch&&document.write('<script src="<%= htmlWebpackPlugin.files.publicPath + "polyfills.js" %>"><\/script>')</script>
```
lukeed: How about `(window.fetch===void 0) && ...`?

rkostrzewski: Sounds great! I'll change it to `window.fetch==void 0&&document.write(...)` (5 less characters & html webpack plugin doesn't minify inline script 😢 )

lukeed: Lol okay. I was just going for perf since `!== void 0` is faster than `typeof === 'undefined'`

src/lib/webpack-config.js
rkostrzewski: From what I understand `serviceWorker` implementation requires `fetch` implementation which in turn requires `Promise` implementation. Based on [FetchEvent](https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent). I could be wrong tough.

lukeed: Ah, right. I believe that's the case too 👍 You upgraded my 90% certainty to 95% 😉

src/resources/template.html
developit: if we're changing, maybe just go with `window.fetch || document.write(..)`.

developit:
Randomly commented in my review - do you guys think we could do something like this to work around using a blocking `document.write(<script>)`?

```html
<body>
  <script src="/bundle.js" defer></script>
  <script>window.fetch || document.body.appendChild(document.createElement('script')).src='/polyfills.js';</script>
</body>
```

ethanroday:
@developit Why not split out `Promise` from the rest of the polyfills and then take the approach you mentioned?
```html
<script>window.Promise || document.body.appendChild(....</script>
```
Then the rest of the polyfills could just be loaded as normal from webpack.

thangngoc89:
@ethanroday currently, preact-cli only includes Promise and fetch. And you need promise supports for fetch anyway. So no splitting needed

lukeed:
Doesn't it _have_ to be blocking in order to ensure it loads before `bundle`?

developit:
@lukeed I wasn't sure if the async script (via createElement) would defer the `defer` script since those wait for document completion.

rkostrzewski:
@developit so I've done my research. 

`document.body.appendChild(document.createElement('script'))` doesn't block deferred scripts (kinda makes sense because `defer` only works for initial html, which `document.write` kinda mutates) so `document.write` has to stay. Good thing is `document.write(<script>)` does not block render (tested in Chrome & IE 11).

I've changed `async` to `defer` to preload `bundle.js` and added parsing blocking `document.write`. Tested in all IE >=9, latest Edge, Chrome, Firefox and android 4 browser. All is well :smile:

lukeed:
@rkostrzewski Makes sense, that's what I had "remembered" too.

However, I'm worried that changing from `async` to `defer` will hurt our load time. It shouldn't matter which attribute the main bundle is using since `document.write` is first & blocks, right?

rkostrzewski:
@lukeed perf should be pretty much the same for `async` & `defer` when using only one script. Additionally, `<script>` tag doesn't block `<script async>` as they are executed as soon as they are downloaded and stack is empty. `bundle.js` is much smaller than `polyfills.js` for initial project which means that it will most likely be downloaded earlier and executed earlier <- Promise doesn't exist yet - `require.ensure` & `import()` fail.

tl;dr `async` is random 😛

lukeed:
Haha okay 😆 Thanks!

src/resources/template.html
```diff
@@ -11,6 +11,7 @@
 			<meta name="theme-color" content="<%= htmlWebpackPlugin.options.manifest.theme_color %>">
 		<% } %>
 		<% for (var chunk of webpack.chunks) { %>
+		 <% if (chunk.names.length === 1 && chunk.names[0] === 'polyfills') continue; %>
 			<% for (var file of chunk.files) { %>
 				<% if (htmlWebpackPlugin.options.preload && file.match(/\.(js|css)$/)) { %>
```
rkostrzewski: BTW `htmlWebpackPlugin.options.preload` comes from `argv.preload` which is a hidden switch that can't be used. Gonna check whole codebase for hidden switches sometime soon.

rkostrzewski:
@lukeed gonna fix regex coz I just noticed it is wrong (`[bundle|polyfills]` should have been `(bundle|polyfills)`. Also `package-json.json` should not been comitted I guess.

reznord:
@rkostrzewski you are using npm5.x? I thought it wasn't that stable and it has many issues.

rkostrzewski:
@reznord works good enough for me. Hadn't had any issues yet.

reznord:
I was had issues with it when I was working with react-native, so thought of upgrading later.

reznord:
any more issues which I can take over?

rkostrzewski:
@reznord if you want feel free take anything with `help wanted` label :smile:

Issue #15, opened 2017-05-20
Extending default configuration


rkostrzewski:
Hi there,
incredible work on the cli & tooling configuration. However, even the greatest config does not meet all needs and thus the question - **Do you have plans on allowing users to extend default configuration?**

Some use cases as a conversation starting point:
- large part of the community likes to follow components & containers folder structure when working with redux and aliasing those in webpack helps with large import paths (i.e.```import Header from '../../../../components/Header```) - **normally this would be done via babel plugin or webpack alias**,
- Service Workers are great because they can cache static files but they are even better because they allow runtime caching - **this would require changing plugin**
- when server side rendering <script defer /> might be better than <script async /> - **again changing plugin config**
- a new cool kid appears around the block (e.g. prepack) and users want to use it - **adding new plugin and in order**

AFAIK [create-react-app](https://github.com/facebookincubator/create-react-app) have chosen not to allow changing configuration. However they provide eject functionality which generates all config files ([docs](https://github.com/facebookincubator/create-react-app#converting-to-a-custom-setup)).
The only problem with this is that once ejected there's no easy way with keeping ejected config in sync with create-react-app.

vitormalencar:
I also think it would be a good feature, this would give much more flexibility to the project.

developit:
I would like to allow configuration extension by looking for a `preact.config.js`, which can export a function that mutates the generated webpack configuration however it pleases.

rkostrzewski:
Do you need help with that? I'll be happy to provide a PR.

Maybe ```preact.config.js``` could be written in ES6 and name property would be enough? As in:
```
config.plugins.map(p =>p.name)
/* ['CommonChunksPlugin', 'ExtractTextPlugin', 'ProvidePlugin'] */
```

Hopefuly it will :smile:

developit:
^ hopefully. That or we provide a helper to get the name from any plugin in the plugins array based on the constructor name.

Writing it in ES6 is fine, I think `preact-cli` already includes `babel-register` in order to do prerendering.  I left a comment here with more thoughts on the idea: https://github.com/developit/preact-cli/issues/37#issuecomment-303383725

kanzelm3:
I would love to have this ability as well. I want to customize postcss plugins so I can do this: `https://github.com/react-toolbox/react-toolbox#settings-configuration-variables-in-javascript`

rkostrzewski:
Custom configuration support has landed in `1.3.0`.

PR #133, opened 2017-06-23
Fix multiple copies of preact in production build (fix #126)

### LINKED ISSUE #126: Production build includes two copies of preact?
### Steps to reproduce:

1. `npm install -g preact-cli`
2. `preact create <app>`
3. `cd <app>`
4. `preact build --json`
5. Inspect `stats.json` using a Webpack stats analyzer (this one is from `webpack-bundle-analyzer`):

image

### Expected behavior:

The preact library should only be included once in `bundle.js`.

### Actual behavior:

It is included twice.

reznord:
Fix for #126

### DIFF
--- a/package.json
+++ b/package.json
@@ -127,6 +127,7 @@
     "promise-polyfill": "^6.0.2",
     "raw-loader": "^0.5.1",
     "recursive-copy": "^2.0.6",
+    "require-relative": "^0.8.7",
     "rimraf": "^2.6.1",
     "script-ext-html-webpack-plugin": "^1.8.0",
     "simplehttp2server": "^2.0.0",
--- a/src/lib/webpack-config.js
+++ b/src/lib/webpack-config.js
@@ -47,6 +47,13 @@ export default env => {
 	let isProd = env && env.production;
 	let cwd = env.cwd = resolve(env.cwd || process.cwd());
 	let src = dir => resolve(env.cwd, env.src || 'src', dir);
+	let preact, type = isProd ? 'preact/dist/preact.min' : 'preact';
+
+	try {
+		preact = require.resolve(type);
+	} catch (err) {
+		throw err;
+	}
 
 	// only use src/ if it exists:
 	if (!exists(src('.'))) {
@@ -77,7 +84,7 @@ export default env => {
 					'preact-cli-entrypoint': src('index.js'),
 					'preact-cli-polyfills': resolve(__dirname, 'polyfills.js'),
 					style: src('style'),
-					preact$: isProd ? resolve(env.cwd, 'node_modules/preact/dist/preact.min.js') : 'preact',
+					preact$: preact,
 					// preact-compat aliases for supporting React dependencies:
 					react: 'preact-compat',
 					'react-dom': 'preact-compat',
--- a/src/lib/webpack-config.js
+++ b/src/lib/webpack-config.js
@@ -23,6 +23,7 @@ import ProgressBarPlugin from 'progress-bar-webpack-plugin';
 import CopyWebpackPlugin from 'copy-webpack-plugin';
 import ReplacePlugin from 'webpack-plugin-replace';
 import SWPrecacheWebpackPlugin from 'sw-precache-webpack-plugin';
+import requireRelative from 'require-relative';
 import createBabelConfig from './babel-config';
 import prerender from './prerender';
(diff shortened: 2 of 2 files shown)


rkostrzewski:
Travis build failed because we Skip installation of dependencies in tests. I'll look into enabling it & caching.

reznord:
@rkostrzewski the tests failed in my local machine as well. Can you please have a look at the log? 

https://paste.fedoraproject.org/paste/~djRyBcRUltRzRRaON3~qg

developit:
I wonder if falling back to `require.resolve()` if the file doesn't exist would work here

reznord:
not sure how that is gonna fit in. Can you tell me how to get this problem fixed?

lukeed:
I think we just need to provide a path to the dev/module bundle? We can do this at the top of the config... something like:

```js
let preact, type = isProd ? 'preact/dist/preact.min' : 'preact';

try {
  preact = require.resolve(type);
} catch (err) {
  throw err;
}

// ...

preact$: preact
```

Also, this is because I'm suspecting that `env.cwd` isn't being set correctly. 

---

Not able to go thru this right now & debug what's happening, else I would! 😄

rkostrzewski:
Nah, tests just skip `npm install` step to be faster. 

@reznord the easiest way is to remove `--no-install` from [here](https://github.com/developit/preact-cli/blob/master/tests/lib/cli.js#L13) tough I'm afraid timeout will occur. I'll tinker around this tomorrow hopefully.

reznord:
Can't we use yarn? It is much faster compared to npm. I know this is just going towards #15 😅

rkostrzewski:
More like #31 😃

reznord:
yeah !!

reznord:
@rkostrzewski the suggestion by @lukeed works perfectly fine. Though that is more like a hack. If it is fine, I'll make another PR to the same with the patch by @lukeed.

I hope that will be fine.

developit:
Suggestion - we can use [require-relative]() for this:

```js
import requireRelative from 'require-relative';

// attempt to resolve a dependency, giving $CWD/node_modules priority:
function resolveDev(dep, cwd) {
  try { return requireRelative.resolve(dep, cwd || process.cwd()); } catch (e) {}
  try { return require.resolve(dep); } catch (e) {}
  return dep;
}

// ...
aliases: {
  preact$: resolveDep(isProd ? 'preact/dist/preact.min' : 'preact', env.cwd),
}
```

lukeed:
Yeah, we can. Is there a benefit here though? Plus I still have a feeling that `env.cwd` isn't set correctly

developit:
Hmm. If `env.cwd` isn't set, a lot of things should be blowing up. It should point to the directory where preact-cli was spawned, or the custom directory passed to the CLI in the case of tests and whatnot ([ref](https://github.com/developit/preact-cli/blob/master/src/lib/webpack-config.js#L48)).

The reason I suggested require-relative is so that if someone installs a specific version of `preact` or `preact-compat`, they'll get that version.  I want to avoid people being stuck with whatever version we happened to ship with `preact-cli`.  We'll want to take that into account when doing prerendering too - currently that uses the one we ship with as well.

lukeed:
Yeah, that's what I thought, but was having doubts looking at the travis logs here.

I agree. But in a Node 6+ environment, it's flat dep tree, so wouldn't a custom install overwrite what we ship? I'm inclined to think it does, since it's the same name (as opposed to a unique dep), but today also seems to be "just one of those days". 😴 😇

developit:
`npm link` creates non-flat trees, I actually think that might have been what @reznord was seeing originally.

reznord:
when I run `npm run test` it is opening up `Google Chrome Canary`. Any idea why that is happening?

reznord:
In the travis log, this kinda caught my mind (https://travis-ci.org/developit/preact-cli/jobs/246455525#L1827) any idea why there are two instances of that?

rkostrzewski:
@reznord it's opening Google Chrome because tests use it to verify if the app works correctly :smile:. As for the warning with AudioManager - it doesn't matter at all (folks at lighthouse have that warning too).

reznord:
@rkostrzewski cool. I have updated the PR. Can you please review it? As @developit suggested, I have added the `require-resolve` module and got it working.

PR #140, opened 2017-06-26
feature: Enable source maps in production

### LINKED ISSUE #103: [feature request] Emit source map for production bundle
It would be great to have source map for production bundle. It should be behind a flag or something (shouldn't be the default behaviour )

I need this for uploading to Sentry.io (an error handling service). 

I can send a PR to add this if you guys are up for it

rkostrzewski:
Closes #103

### DIFF
--- a/src/lib/push-manifest.js
+++ b/src/lib/push-manifest.js
@@ -4,7 +4,7 @@ module.exports = class PushManifestPlugin {
 			let manifest = {};
 
 			for (let filename in compilation.assets) {
-				if (/route-/.test(filename)) {
+				if (/route-/.test(filename) && !/\.map$/.test(filename)) {
 					let path = filename.replace(/route-/, '/').replace(/\.chunk(\.\w+)?\.js$/, '').replace(/\/home/, '/');
 					manifest[path] = {
 						"style.css": {
--- a/tests/build.snapshot.js
+++ b/tests/build.snapshot.js
@@ -1,19 +1,3 @@
-const byteDifferenceThreshold = 50;
-
-export const normalize = obj => {
-	let keys = Object.keys(obj);
-
-	if (keys.length === 1 && keys[0] === 'size' && typeof obj.size === 'number') {
-		return { size: Math.round(obj.size / byteDifferenceThreshold) * byteDifferenceThreshold };
-	}
-
-	return keys.reduce((agg, key) => {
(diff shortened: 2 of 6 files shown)


rkostrzewski:
That was my third try!

developit:
Build seems to be failing due to the tests now seeing `.map` files (yay!!)

tests/build.snapshot.js
```diff
@@ -1,8 +1,10 @@
+const byteDifferenceThreshold = 50;
+
 export const normalize = obj => {
 	let keys = Object.keys(obj);
 
-	if (keys.length === 1 && keys[0] === 'size' && typeof keys[0] === 'number') {
-		return { size: Math.round(obj.size / 10) * 10 };
+	if (keys.length === 1 && keys[0] === 'size' && typeof obj.size === 'number') {
```
rkostrzewski: Turns out `typeof key === 'number'` is never true... wonder why...

Anyway sizes are now compared correctly and without it would never know that push-manifest.json included source maps. 🙈

rkostrzewski:
Seems like snapshots have different sizes on CI because of absolute paths inside maps. Gonna investigate.

lukeed:
I saw that. Would it make sense to just check `file < size`? Dependency upgrades may slightly after the file outputs in the future. Would be a pity to have failing tests because filesize changed ±1 byte.

rkostrzewski:
I was thinking about checking it like this: expected - delta =< actual <= expected + delta. Coz too small could mean no output produced

lukeed:
Yeah, or within 5% of expected. Either way would work imo. 👍

src/lib/push-manifest.js
```diff
@@ -4,7 +4,7 @@ module.exports = class PushManifestPlugin {
 			let manifest = {};
 
 			for (let filename in compilation.assets) {
-				if (/route-/.test(filename)) {
+				if (/route-/.test(filename) && !/\.map$/.test(filename)) {
```
developit: ooh good catch.

developit:
I wonder if there's a way to get webpack to avoid the absolute paths there?  Seems like they could be normalized to be relative to `context`. (totally not in this PR)

rkostrzewski:
@developit seems like module aliases & asnyc-component-loader don't play nicely.

PR #144, opened 2017-06-27
fix(uglify): remove unsafe uglify options

### LINKED ISSUE #136: Uglify settings break library code
I have some crypto library code that when I include it breaks due to the very agressive uglify settings here: https://github.com/developit/preact-cli/blob/master/src/lib/webpack-config.js#L405 

Once I remove the `unsafe_comps` and `unsafe` options everything works fine. I know it's nice to get smaller code but maybe it would be good to disable unsafe options by default as I suspect that there are quite a few libraries out there that will break due to these optimisations. And this is especially hard to catch as tests usually run against dev builds and not against production builds.

#56 would allow me to workaround it, but I still feel that not breaking libraries should be rather the default and unsafe optimisations should be behind a flag.

dignifiedquire:
Closes #136

### DIFF
--- a/src/lib/webpack-config.js
+++ b/src/lib/webpack-config.js
@@ -407,12 +407,10 @@ const production = config => addPlugins([
 		mangle: true,
 		sourceMap: true,
 		compress: {
-			unsafe_comps: true,
 			properties: true,
 			keep_fargs: false,
 			pure_getters: true,
 			collapse_vars: true,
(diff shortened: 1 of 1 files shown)


developit:
having @rkostrzewski's amazing tests makes these almost too easy eh

lukeed:
Yup, 😆 No hesitation in merging that one.

PR #150, opened 2017-06-28
upgrade to webpack3 and add scope hoisting

hassanbazzi:


hassanbazzi:
Not quite sure why this is failing. I have a feeling it has to do with dependency weirdness. Will fix.

developit:
@hassanbazzi `webpack` in this context is `import { webpack } from '@webpack-blocks/webpack2'` - we might need to import the real one instead.

hassanbazzi:
Ahhh yesss

lukeed:
Ah, nice catch @rkostrzewski !

Here is another: https://github.com/developit/preact-cli/blob/cb34094940d7c3a0a646ace6e4b4c031fe285861/src/lib/webpack/webpack-client-config.js#L4

developit:
I'd be up for moving off of webpack-blocks.  We can make it really really easy for ourselves too, all we need is a `compose()` function (I believe there are even webpack-specific ones) that lets us combine configurations the way we use blocks right now.  Then just swap the custom blocks in for partial configs.

lukeed:
@developit Would `webpack-merge` work?

developit:
@lukeed yup, that was the one I was trying to remember the name of.  I believe it just takes a bunch of webpack config partials and produces their combined result. We'd basically just be emulating `customConfig()` and `addPlugins()` since both are fairly useful.  Interestingly, these are already sortof mirrored in @rkostrzewski's helper lib - maybe we merge the two and use our own plugin/helpers API to construct the initial config?  Just spitballing :)

hassanbazzi:
Sorry I'll put more work into this as soon as I'm back from vacation. <3

lukeed:
My bad 😬 Webpack release notes for V3 said that the UglifyJS plugin moved to a separate repo. In that repo, it had separate install instructions. It's still included in the main `webpack` package.

kurtextrem:
Any news? Is there any help needed? :)

rkostrzewski:
@developit @hassanbazzi I've done the work with removing webpack-blocks in favour of webpack-merge. Do we want to have our own blocks (right now it looks like a regular webpack config)?

developit:
I think your webpack-merge stuff looks good, not sure we need more than what's here. Were you just thinking it could be abstracted further?  I'm five with either, this already looks cleaner than blocks (and more understandable) 👍

rkostrzewski:
@developit nah, I just saw in your previous comment some talk about abstracting stuff using things like `addPlugin`. IMHO not using such abstractions actually makes things simpler.

hassanbazzi:
This is awesome. Thanks for picking up @rkostrzewski 

Just wondering, this was supposed to help us shed a bunch of bytes in `polyfill.js`. How come that changed?

hassanbazzi:
Ahhh. The `webpack.optimize.ModuleConcatenationPlugin()` was removed. Any reason why?

lukeed:
@hassanbazzi @rkostrzewski Happened in the merge from `master`. It was there yesterday when I approved.

package.json
```diff
@@ -151,13 +149,15 @@
...
     "source-map": "^0.5.6",
     "stack-trace": "0.0.10",
+    "style-loader": "^0.18.2",
     "sw-precache-webpack-plugin": "^0.11.2",
     "tmp": "0.0.31",
     "unfetch": "^3.0.0",
-    "url-loader": "^0.5.8",
-    "webpack": "^2.3.3",
+    "url-loader": "^0.5.9",
+    "webpack": "^3.3.0",
```
bwhitty: I'd definitely recommend using something higher than 3.3.0 if you're going to use `ModuleConcatenationPlugin`. I had to back out the use of the plugin in an internal app using 3.3.0 due to naming collisions and other strange demons. Check out the release notes of every release since 3.3.0 and you'll see module concat bugfixes. 👍 https://github.com/webpack/webpack/releases

thangngoc89: @bwhitty Don't worry about that, semver will take care of it

PR #153, opened 2017-06-28
Fix Babel options & Respect root's `.babelrc`

### LINKED ISSUE #152: Babel ES-2015 Preset Error
```
./~/preact-classless-component/dist/index.js
Module build failed: Error: Couldn't find preset "es2015" relative to directory "/Users/hassanbazzi/www/vidy-backend/services/website/node_modules/preact-classless-component"
```

Builds on master are failing due to this.

lukeed:
If the user defines a `.babelrc` file in their root directory, we will read it & **overwrite** our settings with what they've specified. This is a hard/forced reset that they may choose to use.

Alternatively, a user may only want to modify the Babel configuration we've provided. Thanks to #56, they can do this now.

```js
// preact.config.js

	let { rule } = helpers.getLoadersByName(config, 'babel-loader');
	rule.options.plugins.push('my-chosen-plugin');
};
```

---

Fixes #152 & others that were bound to open 😆


lukeed:
Oh. This is because we're picking up our own `.babelrc`, which compiled to Node > 4

rkostrzewski:
IMHO `preact build - should use custom .babelrc.` can be ignored for now

rkostrzewski:
@lukeed actually it's because it's picking `.babelrc` which is using custom `plugins` array without jsx transform plugin

lukeed:
Right, the repo's `.babelrc` is being parsed & overwriting our babel config.

It's doing what it's supposed to, we just have to ignore the root `.babelrc` & it'll be fine 😄

developit:
We could move our babel config into the package.json

thangngoc89:
My 2 cents:

Why don't we make a babel presets under `preact-cli/babel` for default babel config, so users can quickly pull them in. next.js does this

lukeed:
@thangngoc89 I was going to suggest this too. Little more difficult since we have some conditional stuff, but should still be doable.

Tests failed because sourcemaps grew as a result of embedded options. @rkostrzewski is on it 🙌

hassanbazzi:
Woooooooooo

lukeed:
@rkostrzewski no, you 🌟 Thanks for editing the how-to above 🙌

ekanna:
```js
// preact.config.js

	let { rule } = helpers.getLoadersByName(config, 'babel-loader');
	rule.options.plugins.push('my-chosen-plugin');
};
```

helpers.getLoadersByName(config, 'babel-loader') returns array.
so change code to below in docs:

```js
// preact.config.js

	let { rule } = helpers.getLoadersByName(config, 'babel-loader')[0];
	rule.options.plugins.push('my-chosen-plugin');
};
```
Also change in examples at 
https://github.com/developit/preact-cli/wiki/Config-Recipes#customising-babel-options-using-loader-helpers

PR #156, opened 2017-06-29
Still generate push manifest when there's no Router

### LINKED ISSUE #154: Push manifest is empty when there is no Router

developit:
When there's no `Router`, still generate a push manifest entry for `/`. Fixes #154.


developit:
lol @lukeed I went back and forth on that twice.

Issue #159, opened 2017-06-29
Webpack fails to resolve scss files imported by other scss files in node_modules


ethanroday:
### Steps to reproduce:
1. Run `preact create <app-name>`
2. `cd <app-name>`
3. `npm install --D material-web-components`
4. In a `.scss` file, include `@import "~@material/drawer/mdc-drawer";` to import the styles for the Material drawer.
5. `npm run build`

### Expected result:
The build will succeed.

### Actual result:
The build fails. It turns out that `@material/drawer/mdc-drawer` is dependent on `@material/rtl/mixins` but, since that import isn't prefixed by a `~`, Webpack doesn't know where to find it.
A quick fix (although probably not the most efficient fix) is to just tell `sass-loader` (and `less-loader`) to look in `node_modules` to resolve imports. I will do a quick PR if that seems like the right solution.

PR #160, opened 2017-06-29
Style import fix

ethanroday:
Addresses #159.
Also changed spaces to tabs, as that seemed to be the convention.


ethanroday:
This would be a breaking change for `.less` files, since providing `paths` to `less-loader` disables Webpack's `~` resolution. For `sass-loader`, though, this is not the case.
@developit I can just remove all of the tildes from `.less` files, but wanted to know if you think this is the right approach.

developit:
Hmm - I've never used `less-loader` with the `path` option.   Does it fully bypass the built-in path handling in `less`?

developit:
@lukeed I think that was my initial reaction too, though this is a lot of moving parts so I wasn't sure.  In the linked #159 issue it seemed like he was hoping to support importing `npm` modules from `.scss` files without using the `~` prefix.

lukeed:
Ah right. I'm inclined to think that moving to `resolve.modules` will get picked up by `sass-loader` (didn't check). But if not, then leaving the `includePaths` will be fine.

The important part is not breaking `less-loader` behavior.

reznord:
@lukeed @developit is this still valid?

developit:
Crap, I botched that merge.

lukeed:
We should also add `paths` to the `stylus-loader`, now that we have it.

PR #162, opened 2017-06-29
fix: tests

rkostrzewski:
This adds:
- `npm install` done by `preact create` in tests for PRs and `master` branch
- `npm install --save-dev ../../path-to-preact-cli-under-test` in above tests so that preact-cli in node_modules is the one under tests (i.e. from branch)
- `.npmignore` which is kinda needed not to install all tests with their output folders when doing above
- operation logging to tests - should be easier to determine what failed (only on PRs and `master`)
- `prebuild` script which removes `./lib` folder with output (might run into some issues with this in the future)
- `--git` switch to `preact create` that is true by default not to create git repository inside another one
- tests now use `jest` and run in parallel (brings down time a bit)

Tests are looking stable.


package.json
```diff
@@ -81,8 +82,6 @@
     "html-looks-like": "^1.0.2",
     "lodash": "^4.17.4",
     "ncp": "^2.0.0",
-    "node-sass": "^4.5.3",
```
rkostrzewski: I've added them for sass tests - no longer needed.

Doesn't matter now coz there is a bug in npm 5 which causes "devDepdencies" to be included when installing locally.

.npmignore
```diff
@@ -0,0 +1,5 @@
+tests/
```
rkostrzewski: Ignored when using local dependency on npm 5 :cry:

.travis.yml
```diff
@@ -8,15 +8,17 @@ addons:
...
 cache:
   npm: true
+  directories:
+    - $HOME/.npm
+    - ./node_modules
 
 before_script:
 - export LIGHTHOUSE_CHROMIUM_PATH=`which google-chrome-stable`
-- google-chrome-stable --headless --disable-gpu --remote-debugging-port=9222 http://localhost &
+- if [ "$TRAVIS_BRANCH" = "master" ]; then export WITH_INSTALL=true; export WITH_LOG=true; fi
```
rkostrzewski: only run tests with `preact create --install` on master (+ PRs)

.npmignore
```diff
@@ -0,0 +1,5 @@
...
+src/
+examples/
```
lukeed: Gotta keep this 💃

.travis.yml
```diff
@@ -9,13 +9,17 @@ language: node_js
...
```
lukeed: This is awesome! Didn't know you could control Travis to this extent.

src/commands/create.js
```diff
@@ -179,7 +186,10 @@ async function initializeVersionControl(target) {
 	let git;
 	try {
 		git = await promisify(which)('git');
-	} catch (e) {}
+	} catch (e) {
+		process.stderr.write('Could not find git in $PATH.\n');
+		process.stdout.write('Continuing without initializing version control...\n');
```
lukeed: Nice 👍

.travis.yml
rkostrzewski: Yeah, I love travis because it's so easy to work with, yet so powerful.

.npmignore
rkostrzewski: TBH I didn't notice the `files` option. I'll be removing this then

Issue #165, opened 2017-06-30
Syntax error issues on preact create


bobc82:
I tried to run the preact-cli installation as follows:
```
sudo npm i -g preact-cli
```
And preact-cli was installed as expected. After, i tried to create a project:
```
sudo preact create example-app
```
On the first attempt terminal returned the following Syntax Error:
```
/usr/local/lib/node_modules/preact-cli/node_modules/webpack-plugin-replace/lib/index.js:9
	let i=0, len=arr.length, map=new Map();
	^^^

SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:374:25)
    at Object.Module._extensions..js (module.js:417:10)
    at Module.load (module.js:344:32)
    at Function.Module._load (module.js:301:12)
    at Module.require (module.js:354:17)
    at require (internal/module.js:12:17)
    at Object.<anonymous> (/usr/local/lib/node_modules/preact-cli/lib/lib/webpack/webpack-base-config.js:25:29)
    at Module._compile (module.js:410:26)
    at Object.Module._extensions..js (module.js:417:10)
    at Module.load (module.js:344:32)
    at Function.Module._load (module.js:301:12)
    at Module.require (module.js:354:17)
    at require (internal/module.js:12:17)
    at Object.<anonymous> (/usr/local/lib/node_modules/preact-cli/lib/lib/webpack/webpack-client-config.js:39:26)
    at Module._compile (module.js:410:26)
    at Object.Module._extensions..js (module.js:417:10)
    at Module.load (module.js:344:32)
    at Function.Module._load (module.js:301:12)
    at Module.require (module.js:354:17)
    at require (internal/module.js:12:17)
    at Object.<anonymous> (/usr/local/lib/node_modules/preact-cli/lib/lib/webpack/run-webpack.js:27:28)
```
I tried to fix myself this issue adding on top of /usr/local/lib/node_modules/preact-cli/node_modules/webpack-plugin-replace/lib/index.js file:
```
"use strict";
```
In the second attempt, the preact create command fired another Syntax error, as follows:
```
/usr/local/lib/node_modules/preact-cli/node_modules/devcert-san/dist/index.js:22
function devcert(appName, options = {}) {
                                  ^

SyntaxError: Unexpected token =
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:374:25)
[...]

PR #168, opened 2017-06-30
Declare Node v6 as minimum support

lukeed:
The `engines` key will only show a warning to NPM installers. It's not perfect, but at least it's something.

Although we were compiling to Node 4 and _our_ code worked in a Node 4 environment (I just tried, dat `npm install` doe...💤 ), we have several dependencies that require 5+ and 6+ _and_ our dependency resolution was fighting against NPM's dependency tree (typical).

So with this, we're now compiling to Node 6 & actively declaring a 6+ minimum.

I removed Node 7 from Travis because we're already testing v8 and v7 literally brings nothing unique to the table. This will make tests completely a bit sooner.

References #165

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -4,7 +4,7 @@
     ["env", {
       "loose": true,
       "targets": {
-        "node": 4
+        "node": 6
       },
       "exclude": [
         "transform-regenerator"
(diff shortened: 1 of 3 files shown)


lukeed:
Oh, and normally this would be a breaking change for a library. But it isn't for us because we've already been _informally_ requiring Node 6. This is just formally writing it down. 📝

PR #171, opened 2017-07-01
check node version on postinstall & while initiating a new project

### LINKED ISSUE #169: Check for dependencies while creating a new project
It would be great to have a dependency check so we can avoid running into issues like #165 

I would love to add the following [code](https://github.com/phenomic/phenomic/blob/fa314107debb6ca1f860b7c7a38dc30b0e2c8b6b/packages/cli/src/check-engine.js) for dependency check.

@rkostrzewski @developit @lukeed any suggestions?

reznord:
Fixes #169


reznord:
Travis checks are failing because `lib/lib/check-engine.js` is not available. 

If we update the build on travis, that would fix the tests.

developit:
Either way `process.version` seems super nice here, since it makes the whole check synchronous.

lukeed:
@developit Are you opposed to moving this `check-engine.js` file to the root & adding it to the NPM files?

This way, we can let the tests run without doing any fancy side stepping. Currently, this will be fine with downloads since a `lib/` will already be built.

developit:
Yup, I'm fine with that. Could almost be a `node -e` in scripts, but the root is good enough.

package.json
```diff
@@ -13,7 +13,7 @@
     "dev": "babel-node src",
     "lint": "eslint src tests",
     "pretest": "npm run -s build",
-    "postinstall": "node lib/lib/check-engine.js",
+    "postinstall": "babel-node check.js",
```
thangngoc89: Please don't do this. `babel-node` is in devDeps so it would break user install

lukeed: Oh right. Forgot not all babel pkgs are normal deps.

Let's rewrite to ES5 and just use `node`.

thangngoc89:
@reznord Your commit message make me laugh :D 

> 	check.js (es5 version)

And you're using template literal

reznord:
> Your commit message make me laugh :D

Bad time for a commit I'd say :P (2AM) anyways, its all sorted :P

lukeed:
Template literals ftw! 🎉

PR #172, opened 2017-07-01
Feature: Immutable build

### LINKED ISSUE #23: Version bundle.js & style.css
I would call this "mild priority" right now, simply because, as is, `bundle.js` is the majority of the application weight//network transfer. It's set for a maximum of 1 hour, which means that users will need to redownload an entire application, regardless of any changes it may or may not have had. Plus, this kind of impedes the function of offline/sw-precache. 

Also, in the _rare_ event that my app goes from v1 to v2 within the hour that User has cached my bundle, their next request would grab all the v2 chunks & `sw.js`, which still points to non-versioned `bundle.js`, which results in User working with a mismatched set of assets... the majority of which (in bundle) is now outdated.

If we're on the same page here, we'd need to:

- [ ] Version `bundle.js`, like everything else
- [ ] Update server headers; would up it to 1YR
- [ ] Replace the HTML `bundle.js` link per-build; handled by [`html-webpack-plugin`](https://github.com/jantimon/html-webpack-plugin)

This all applies to `style.css` as well: same tasks & reasoning.

thangngoc89:
I'm pushing this here to get the ideas first, will fix test later
With this change, only `index.html` and `sw.js` is mutable between builds. 

Fixes #23
Blocked by #162


developit:
We should pull some people in regarding the hashing, I think it has kinda wild implications for ServiceWorkers.

thangngoc89:
I'm not sure but never see anyone hash ServiceWorker runtime, just be sure to set the right HttpCache on your server.

lukeed:
You mean for `style.css` and `bundle.js`? Looks okay to me, what would be the problem? As per #23, they get treated like any other assets.

thangngoc89:
@rkostrzewski Pinging the tests guy. Since filename is changing between build, do you have any plan to test it?

developit:
We could take a page from CRA and strip the hashes from filenames prior to doing the comparison

lukeed:
Yup. That, or look for exact match else regex match.

thangngoc89:
@developit if that is the case, I'll wait for #162 before fix the tests

src/lib/webpack/webpack-base-config.js
```diff
@@ -220,7 +220,7 @@ export default (env) => {
 		// produce HTML & CSS:
 		addPlugins([
 			new ExtractTextPlugin({
-				filename: 'style.css',
+				filename: isProd ? "style.[contenthash:5].css" : "style.css",
```
rkostrzewski: I love the idea on client, not so on server side. I'd much prefer to have static file on server for critical css inlining.

I think we should pass an argument to `webpack-base-config` from `webpack-client-config` and `webpack-server-config`

rkostrzewski:
@thangngoc89 I'll merge the tests fix and add regex matching - that sounds cool 😎

src/lib/webpack/webpack-base-config.js
rkostrzewski: The other option is to expose a `mainfest.json` which would be used to resolving filenames from entries (dunno how this works - only seen folks doing that)

rkostrzewski:
Travis build failed because it got different hashes than on my local env. This could break long term hashing.

Different hashes = some code is undeterministic (i.e. relies on env it was built on). My bet is that [this line](https://github.com/developit/preact-cli/blob/master/src/lib/webpack/async-component-loader.js#L19) produced this output.

thangngoc89:
@rkostrzewski it's not just that line, a lot of things need to to config to support long term caching

rkostrzewski:
@thangngoc89 ok I've found what's the issue here. Entrypoint relative path changes on travis because of different way `preact-cli` is used ('global' vs local). After changes in `async-component-loader` entrypoints chunkhashes are consistend when using locally installed `preact-cli` (can't say the same about route-*.chunks :cry:)

thangngoc89:
@rkostrzewski I'm suggesting that we keep this PR as starting point for immutable build. ANd send another PR for adding Long term caching support.

src/lib/webpack/push-manifest.js
```diff
@@ -1,33 +1,45 @@
...
+			routes.forEach(filename => {
+				let path = filename.replace(/route-/, '/').replace(/\.chunk(\.\w+)?\.js$/, '').replace(/\/home/, '/');
+				manifest[path] = {
+					...defaults,
+					[filename]: {
+						type: "script",
+						weight: 0.9
+					}
+				};
+			});
```
rkostrzewski: Fixed those spaces for you :smile:

thangngoc89: @rkostrzewski Thx. That's usually prettier's job

thangngoc89:
@lukeed  @developit It's good for your review

lukeed:
I'll review this tonight 🙏

PR #173, opened 2017-07-02
Expose development server network interface

### LINKED ISSUE #170: Idea: Expose development server to network interface
Is it possible to expose the development server to the network interface? This would make it very easy for someone to access his preact app on his phone without having to fiddle around with simulators.

This already works in create-react-app:

> $ yarn start
> Compiled successfully!
> 
> You can now view test in the browser.
> 
>   Local:            http://localhost:3000/
>   On Your Network:  http://192.168.2.104:3000/
> 
> Note that the development build is not optimized.
> To create a production build, use yarn build.

If such a feature already exists, how do I enable it?

reznord:
Exposes network address for development build (fix #170)

This is how it looks now:
image 2017-07-02 at 2 13 33 pm

### DIFF
--- a/src/lib/webpack/run-webpack.js
+++ b/src/lib/webpack/run-webpack.js
@@ -27,8 +27,14 @@ const devBuild = async (env, onprogress) => {
 			if (first) {
 				first = false;
 				let devServer = config.devServer;
-				let serverAddr = `${devServer.https?'https':'http'}://${process.env.HOST || devServer.host || 'localhost'}:${process.env.PORT || devServer.port || 8080}`;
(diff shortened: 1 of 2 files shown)


src/lib/webpack/run-webpack.js
```diff
@@ -1,5 +1,6 @@
 import path from 'path';
 import fs from 'fs.promised';
+import ip from 'ip';
```
thangngoc89: You forgot to install this module

reznord:
@thangngoc89 that package is already available in the node_modules 

here you go:
image 2017-07-02 at 2 54 29 pm

thangngoc89:
@reznord Don't rely on that, we don't know if webpack-dev-server gonna remove it in the future. If we use it directly, import it.

reznord:
okay, will add it to the dependencies

Issue #175, opened 2017-07-02
preact watch --https=true, runs http instead of https if cert has been created.


flea89:
I've tried to run the devserver on https, If the webpack certificate is used the server starts an https server as expected.

If a certificate has been created ( the webpack one isn't used)  the server starts over http instead of https.

I've had a quick look to the source code, and I think the issue is https://github.com/developit/preact-cli/blob/2bd02cb8e5c442c0987d90d6cada3a91516fa790/src/lib/webpack/webpack-client-config.js#L126, ssl is an object and not a boolean.

If I'm not wrong that object is set https://github.com/developit/preact-cli/blob/master/src/commands/watch.js#L47.

Thanks

developit:
The flag is `preact watch --https` - it seems like passing `=true` is killing that conditional.  Can you test without the value?

flea89:
Unfortunately that doesn't solved the problem.
Https option is actually picked but doesn't starts an https server in that case ( it does with the same command if webpack cert is used).

Isn't the problem that getSsl returns an object and webpack-client-config tests for a boolean true ?

developit:
hmm yeah - not sure where that was introduced.

flea89:
Nice I was just sending the pull request but I this has been already fixed.
Nice!

reznord:
Closing this issue, fixed in #176

PR #176, opened 2017-07-03
removed boolean check for running dev server in https

reznord:
Removed boolean check for running dev server in https from [here](https://github.com/developit/preact-cli/blob/2bd02cb8e5c442c0987d90d6cada3a91516fa790/src/lib/webpack/webpack-client-config.js#L126)

Now you can run dev server in `HTTPS`

Fix for bug #175

### DIFF
--- a/src/commands/watch.js
+++ b/src/commands/watch.js
@@ -19,8 +19,7 @@ export default asyncCommand({
 		},
 		host: {
 			description: 'Hostname to start a server on',
-			default: '0.0.0.0',
-			alias: 'h'
+			default: '0.0.0.0'
 		},
 		https: {
 			description: 'Use HTTPS?',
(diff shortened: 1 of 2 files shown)


src/commands/watch.js
```diff
@@ -19,8 +19,7 @@ export default asyncCommand({
 		},
 		host: {
 			description: 'Hostname to start a server on',
-			default: '0.0.0.0',
-			alias: 'h'
```
reznord: removing the alias for the `host` command since this is colliding with `yargs` package's `--help, -h`.

reznord: and this is supposed to be a separate PR, but since already a PR is in queue, it just appended !! 

Sorry for the mess 😅

thangngoc89: Please move this to a seperate PR as your PR could be squashed to create a clean git history.

PR #181, opened 2017-07-03
remove h alias of host option

### LINKED ISSUE #180: Cannot watch on specific host
**Problem:**

When I run `preact watch --host "127.0.0.1"` inside a `preact` project created with `preact-cli`, I got this screen:

preactcli

It seems that `yargs` package overrides the `h` alias option and turns it into a help command.

**How to fix:**

Delete [the alias](https://github.com/developit/preact-cli/blob/master/src/commands/watch.js#L23) of `host` option resolves the issue.

preactcli2

minhchu:
fix #180

### DIFF
--- a/tests/lib/cli.js
+++ b/tests/lib/cli.js
@@ -39,8 +39,8 @@ export const serve = (appDir, port) => withLog(
 	'preact serve'
 );
 
-export const watch = (appDir, port) => withLog(
-	() => spawnPreact(['watch', port ? `-p=${port}` : undefined], appDir),
+export const watch = (appDir, host, port) => withLog(
+	() => spawnPreact(['watch', host ? `--host=${host}` : undefined, port ? `-p=${port}` : undefined], appDir),
 	'preact watch'
 );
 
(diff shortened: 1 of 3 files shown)


reznord:
@minhchu I have committed the fix for this [here](https://github.com/developit/preact-cli/pull/176/commits/a4f71a1e5d7242e4e364c85be34cd4e9439aaa89). 

Sorry for not linking the PR in the comment. It was supposed to be a separate PR (my bad). If you are fine with this, I'll close this since I've already made changes there in the above mentioned commit.

minhchu:
It's fine for me. Let close this PR.

reznord:
I've reverted that commit. So, lets continue here !!

tests/lib/cli.js
```diff
@@ -39,8 +39,8 @@ export const serve = (appDir, port) => withLog(
 	'preact serve'
 );
 
-export const watch = (appDir, port) => withLog(
-	() => spawnPreact(['watch', port ? `-p=${port}` : undefined], appDir),
+export const watch = (appDir, host, port) => withLog(
+	() => spawnPreact(['watch', host ? `--host=${host}` : undefined, port ? `-p=${port}` : undefined], appDir),
```
reznord: I don't think it is necessary to provide a host here separately. No harm specifying tho.

cc/@developit what do you think?

PR #184, opened 2017-07-03
Update to stage-1

SaraVieira:
Since stage 0 is very unpredictable and quickly changes I think it would be better to update this to stage-1.

All we need is already in stage-1 and it's more predictable

What do you guys think ?


reznord:
@SaraVieira Yarn support for `preact-cli` is still WIP. You can check the status for that in [here](https://github.com/developit/preact-cli/pull/31).

As far as this yarn.lock is concerned, please don't commit the file, some might use `npm` and some might use `yarn`.

prateekbh:
Will adding it to gitignore not help?

reznord:
That should do !!

package.json
```diff
@@ -84,6 +84,7 @@
     "babel-cli": "^6.24.0",
     "babel-eslint": "^7.2.1",
     "babel-plugin-transform-runtime": "^6.23.0",
+    "babel-preset-stage-1": "^6.24.1",
```
thangngoc89: Could you move this to dependency like stage 0? We need it there, not devDep

SaraVieira:
@lukeed yarn.lock removed and added to the gitignore
@thangngoc89 Also changed

developit:
ooh, the stages changed since I last looked - definitely agreed we don't need Do expressions or the function bind syntax.  good call @SaraVieira!

lukeed:
Anyone know why Node 8 is failing with `cannot find module 'jest-cli'`? I've restarted it once & looks like someone else just did too 😅

thangngoc89:
@lukeed I just restarted it. It failed. When we have a test problem, we shall ping @rkostrzewski

SaraVieira:
Thanks @developit! 

I was gonna ask someone to restart it but apparently it didn't fix it 😐

SaraVieira:
It did pass now 😄 

Is this okay to merge guys ?

rkostrzewski:
Removed cache for build. Commiting `yarn.lock` seems to have changed node_modules structure and cached entities had incorrect structure. Weird.

hzoo:
just as fyi, stage 3 is what is "predictable". Stage 0 is basically automatically ok if someone champions it at TC39, and stage 1 isn't that much different. It's best to opt-in to individual plugins/proposals rather than use stages before Stage 3 but currently we leave it there as an option.

PR #192, opened 2017-07-05
Added multiroute prerendering

framp:
This PR adds support for prerendering a set of routes instead of limiting to `/` (as per https://github.com/developit/preact-cli/issues/185#issuecomment-312759992)

In order to use this functionality, you have to add a `prerender-urls.json` file, which contains the set of routes you want to render.
The format for the file is an array of objects with a `url` key and an optional `title` key which will override the title.
Eg:
```
[{"url": "/", "title": "Homepage"}, { "url": "/route/random"}]
```

I've added a --prerenderUrls option which let you change the path of the `prerender-urls.json` to whatever you like.

Feedbacks appreciated :)


lukeed:
This is great 😃 Thanks!

carlosagsmendes:
Dear @framp with the following URLs configuration for the included sample:

`
[
  {
    "url": "/",
    "title": "Homepage"
  },
  {
    "url": "/profile ",
    "title": "Overriden Profile Title"
  },
  {
    "url": "/profile/john ",
    "title": "John's Profile"
  }
]
`

I'm getting two **profile** folders pre-rendered:

PreRender Results

Does this seem right?

framp:
@carlosagsmendes it does not seem right and it's not possible on a filesystem level.
I suppose it's a visualisation bug in your editor, probably triggered by indexing files in the same directory at a close time.

Do you mind doing a `tree` or `ls -hall` from the terminal in your `build` directory or restarting your editor?

carlosagsmendes:
Here's the output after deleting the folder and doing a new build:

LS-Hall

One thing I noticed is that if I do a "Hard Reload" on Chrome on the https://localhost:8080/profile/ Url I don't get the correct page rendered:

HardReload

Issue #189, opened 2017-07-04
Stylus support?


yavorski:
Hi, In our team we are considering `preact` for a new application.
However we use `stylus` for css preprocessor.
Do you have any plans about including `stylus` in default options?
If so, we would be happy to try to submit a pull request, if some general guidance is available.

Thanks in advance.

thangngoc89:
@yavorski preact-cli@1.3.0 comes with a `preact.config.js` so you can alter your webpack config there. No need to wait for us

yavorski:
@thangngoc89 thank you for fast answer!
We are definitely going to try this in our starter app.
We still think that it would be nice for some users to have the option build in?

yavorski:
s

thangngoc89:
IMHO, it would be better for us to leave this in user land.

cc @developit @lukeed for more opinion on this one

yavorski:
Fair Enough! 
Just wanted to make sure what do you think!

lukeed:
@thangngoc89 I tend to agree; _however_, the counterargument is that we already offer SASS & LESS. Maybe we should just so that we cover all the bases?

yavorski:
Hi, 
After a little exploration of the code base, I think that this is fairly easy to implement.
Of course I could be wrong (I see it for the first time), but I managed to prepare a pull request.
I tested all localy and it works fine for me and our team.
I hope this would be useful to other teams too, and in general we think that this should be a good option for people using stylus out there.
Looking forward for comments and thank you for your great work so far!

lukeed:
Closed via #194 🎉

yavorski:
That's Awesome, Thank you! 
When we should expect new version, so we could install :)

lukeed:
Looks like there's a few changes coming for 1.4, including this. I'd wager end of the week at the latest 👍

yavorski:
Great, thanks! ❤️

PR #194, opened 2017-07-05
Add Stylus Support

yavorski:
#189


yavorski:
That's Awesome, Thank you too!

PR #195, opened 2017-07-05
feat: Friendly prerender message + remove UglifyJS from SSR build

### LINKED ISSUE #178: Prerender error and workarounds
Preact-cli comes with prerender which render your `/` route using server-side rendering technique. So if you run `npm start` successfully but `npm run build` produces some errors, this mean that the prerender is failing because of some code requires DOM to work.

Workarounds:

- Wrap your with this to make sure it won't execute when prerendering. 

```
if (typeof window !== "undefined") {
  // Your code here
}
```

- Turn off preredering by run `npm run build -- --no-prerender`

rkostrzewski:
This closes #178

screen shot 2017-07-05 at 16 52 48

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 - Fully **automatic code splitting** for routes
 - Transparently code-split any component with an [`async!`] prefix
 - Auto-generated [Service Workers] for offline caching powered by [sw-precache]
-- [PRPL](https://developers.google.com/web/fundamentals/performance/prpl-pattern/) pattern support for efficient loading
+- [PRPL] pattern support for efficient loading
 - Zero-configuration pre-rendering / server-side rendering hydration
 - Support for CSS Modules, LESS & autoprefixer
 - Monitor your bundle/chunk sizes with built-in tracking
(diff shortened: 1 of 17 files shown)


rkostrzewski:
Heh, previously I've had `See https://github.com/developit/preact-cli#prerender for further information.` but I was too lazy to write it 😆

lukeed:
Looks great @rkostrzewski ! 😻 

Quick question though: Why was uglify removed on SSR? UglifyJS was mangling the maps, ergo making tracing impossible?

rkostrzewski:
@lukeed that's one of the reasons there other being that chief said so 😆, right @developit ?

lukeed:
@rkostrzewski Lol 😂  Just went thru that Slack. Looks like Chief said that _because_ of the source map issues!

hassanbazzi:
I notice this suggests people should check for window. I always just use this:

https://www.npmjs.com/package/window-or-global

It also makes SSR work for a bunch of features that need global, and it doesn't break even if you run things like eventListeners on global.

rkostrzewski:
@thangngoc89 I've added `Pre-rendering` part to the `README.md` let me know if this is what you had in mind and I'll be merging this

README.md
```diff
@@ -8,7 +8,7 @@
 - Fully **automatic code splitting** for routes
 - Transparently code-split any component with an [`async!`] prefix
 - Auto-generated [Service Workers] for offline caching powered by [sw-precache]
-- [PRPL](https://developers.google.com/web/fundamentals/performance/prpl-pattern/) pattern support for efficient loading
```
kurtextrem: why remove the link? :)

thangngoc89:
@rkostrzewski LGTM

Issue #179, opened 2017-07-03
Suggest preact commands to run dev server on new port if 8080 is already being used


reznord:
As of now, when we run a dev server using `preact serve` or `preact watch` by default opens on `8080` port.

If there is a process which runs on `8080` already, preact doesn't ask the user whether he wants to run on a new port. As of now it terminates the process.

It would be great if we can suggest the CLI to ask the user whether he wants to run on a different port.

thangngoc89:
This would be a great addition. I have some code snippets I can paste here if you need it

reznord:
> I have some code snippets I can paste here if you need it

Cool, I'd love it

thangngoc89:
this time, it's 3 lines


```
getPort(3000).then(port => {
	console.log(port);
	// Will use 3000 if available, otherwise fall back to a random port
});
```

if 3000 !== port, then ask user if they want to spawn in other port.

thangngoc89:
on *nix, you can use this command this see which process is binding to a node, but it may require sudo 

![image](https://user-images.githubusercontent.com/3049054/27781141-d9f3dfd6-5ff6-11e7-9ca2-0d4dbc61b85e.png)

rkostrzewski:
We can always simply spawn http server using `0` port? Even if not all servers accept port `0` we could spawn http server at port `0` get actual port & kill it.

thangngoc89:
@rkostrzewski that's what `get-port` package doing.

```js
'use strict';
const net = require('net');

const getPort = port => new Promise((resolve, reject) => {
	const server = net.createServer();

	server.unref();
	server.on('error', reject);

	server.listen(port, () => {
		const port = server.address().port;
		server.close(() => {
			resolve(port);
		});
	});
});

module.exports = preferredPort => preferredPort ?
	getPort(preferredPort).catch(() => getPort(0)) :
	getPort(0);
```

reznord:
CRA does this in a pretty nice way !! I would just take the handling of PORT from CRA and use it here !! That is fine right?

developit:
We can use the popular [get-port](https://www.npmjs.com/package/get-port) module to handle this:

```js
getPort(3000).then(port => {
    console.log(port);
    // Will use 3000 if available, otherwise fall back to a random port 
});
```

rkostrzewski:
@thangngoc89 I didn't see that require call 😆

thangngoc89:
[...]

PR #196, opened 2017-07-05
Suggest a new port for development server if default port is in use.

reznord:
This PR opens dev server in a new port if the default port is already in use.

Here's how it looks:
image 2017-07-06 at 2 19 44 am


reznord:
Oops !! committed `package-lock.json` file by mistake !! 

My bad :(

src/lib/webpack/run-webpack.js
```diff
@@ -30,13 +33,15 @@ const devBuild = async (env, onprogress) => {
...
 
-				let serverAddr = `${protocol}://${host}:${port}`;
-				let localIpAddr = `${protocol}://${ip.address()}:${port}`;
+				let serverAddr = `${protocol}://${host}:${chalk.bold(port)}`;
+				let localIpAddr = `${protocol}://${ip.address()}:${chalk.bold(port)}`;
 
-				process.stdout.write(chalk.green('Compiled successfully!!\n\n'));
-				process.stdout.write('You can view the application in the browser.\n\n');
+				process.stdout.write(chalk.green('\nCompiled successfully!!\n\n'));
+				if (parseInt(port, 10) !== parseInt(config.devServer.port, 10)) {
```
reznord: Showing this message only when it is not using default port. So, it will be easy for the users to understand what port it has been changed to !!

reznord:
This fixes #179

PR #204, opened 2017-07-07
Add description of install and no-install options.

knight-bubble:
Added two missing options for the create command, changed formatting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,14 +50,15 @@ npm run build
 ```sh
 $ preact create
 
-  --name      directory and package name for the new app.
-  --dest      Directory to create the app within.                 [default: <name>]
-  --type      A project template to start from.
-              [Options: "default", "root", "simple", "empty"]     [default: "default"]
-  --less      Pre-install LESS support.                [boolean]  [default: false]
-  --sass      Pre-install SASS/SCSS support.           [boolean]  [default: false]
-  --stylus    Pre-install STYLUS support               [boolean]  [default: false]
-  --git       Initialize version control using git.    [boolean]  [default: true]
+  --name        Directory and package name for the new app.
+  --dest        Directory to create the app within.                 [default: <name>]
+  --type        A project template to start from.
+                [Options: "default", "root", "simple", "empty"]     [default: "default"]
+  --less        Pre-install LESS support.                [boolean]  [default: false]
+  --sass        Pre-install SASS/SCSS support.           [boolean]  [default: false]
+  --stylus      Pre-install STYLUS support.              [boolean]  [default: false]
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -50,14 +50,16 @@ npm run build
...
-  --git       Initialize version control using git.    [boolean]  [default: true]
+  --name        Directory and package name for the new app.
+  --dest        Directory to create the app within.                 [default: <name>]
+  --type        A project template to start from.
+                [Options: "default", "root", "simple", "empty"]     [default: "default"]
+  --less        Pre-install LESS support.                [boolean]  [default: false]
+  --sass        Pre-install SASS/SCSS support.           [boolean]  [default: false]
+  --stylus      Pre-install STYLUS support.              [boolean]  [default: false]
+  --git         Initialize version control using git.    [boolean]  [default: true]
+  --install     Install dependencies.                    [boolean]  [default: true]
```
reznord: This is not required since by default we will be installing dependencies. Please remove this.

knight-bubble: So basically we shouldn't be able to do
```
preact create appname --install=false
```
Right?
And instead
```
preact create appname--no-install
```

reznord: Yes, it is `preact create my_app --no-install`, with this, it will just simply initialise the project but will not install node_modules.

developit:
Might want to stick with `--install default:true` - already fielding questions about `--no-install` haha

README.md
```diff
@@ -50,14 +50,15 @@ npm run build
...
-  --git       Initialize version control using git.    [boolean]  [default: true]
+  --name        Directory and package name for the new app.
+  --dest        Directory to create the app within.                 [default: <name>]
+  --type        A project template to start from.
+                [Options: "default", "root", "simple", "empty"]     [default: "default"]
+  --less        Pre-install LESS support.                [boolean]  [default: false]
+  --sass        Pre-install SASS/SCSS support.           [boolean]  [default: false]
+  --stylus      Pre-install STYLUS support.              [boolean]  [default: false]
+  --git         Initialize version control using git.    [boolean]  [default: true]
+  --no-install  Disables installing of dependensies.     [boolean]  [default: false]
```
prateekbh: y does this need a `[boolean]  [default: false]`

reznord:
> Might want to stick with --install default:true

what is the point of adding that, since the default functionality is that

README.md
reznord: Ah right !! Will change it (there are a couple of things to change too)

reznord: Ah right !! Will change it (there are a couple of things to change too)

PR #206, opened 2017-07-08
feat: Less verbose error output

rkostrzewski:
+1000 to DX

`preact build` before:
build-old

`preact build` after:
build-new

`preact watch` before:
watch-old

`preact watch` after:
watch-new

Changes:
- `stats.toJSON()` accepts `options` which could be 'errors-only' (like in `webpack.config.js`)
- rewrote `babel-loader` query stripping (placed by `async-component-loader`) to be more precise & maintanable (= renamed it 😆 )

### DIFF
--- a/src/lib/webpack/run-webpack.js
+++ b/src/lib/webpack/run-webpack.js
@@ -38,7 +38,12 @@ const devBuild = async (env, onprogress) => {
 				let serverAddr = `${protocol}://${host}:${chalk.bold(port)}`;
 				let localIpAddr = `${protocol}://${ip.address()}:${chalk.bold(port)}`;
 
-				process.stdout.write(chalk.green('\nCompiled successfully!!\n\n'));
+				if (stats.hasErrors()) {
+					process.stdout.write(chalk.red('\Build failed!\n\n'));
+				} else {
+					process.stdout.write(chalk.green('\nCompiled successfully!\n\n'));
+				}
+
 				if (userPort !== port) {
 					process.stdout.write(`Port ${chalk.bold(userPort)} is in use, using ${chalk.bold(port)} instead\n\n`);
 				}
@@ -71,7 +76,8 @@ const prodBuild = async (env) => {
 	return await new Promise((resolve, reject) => {
 		compiler.run((err, stats) => {
 			if (err || stats.hasErrors()) {
-				reject(err || stats.toJson().errors.join('\n'));
+				showStats(stats);
+				reject(chalk.red('Build failed!'));
 			}
 			else {
 				// Timeout for plugins that work on `after-emit` event of webpack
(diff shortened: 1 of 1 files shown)


src/lib/webpack/run-webpack.js
```diff
@@ -38,7 +38,12 @@ const devBuild = async (env, onprogress) => {
 				let serverAddr = `${protocol}://${host}:${chalk.bold(port)}`;
 				let localIpAddr = `${protocol}://${ip.address()}:${chalk.bold(port)}`;
 
-				process.stdout.write(chalk.green('\nCompiled successfully!!\n\n'));
+				if (stats.hasErrors()) {
+					process.stdout.write(chalk.red('\Build failed!\n\n'));
+				} else {
+					process.stdout.write(chalk.green('\nCompiled successfully!\n\n'));
+				}
```
reznord: Better to expose the IP addresses & port message only when the build is successful !!

```javascript
    if (userPort !== port) {
       process.stdout.write(`Port ${chalk.bold(userPort)} is in use, using ${chalk.bold(port)} instead\n\n`);
    }
    process.stdout.write('You can view the application in browser.\n\n');
    process.stdout.write(`${chalk.bold('Local:')}            ${serverAddr}\n`);
    process.stdout.write(`${chalk.bold('On Your Network:')}  ${localIpAddr}\n`);
```

rkostrzewski: That could be possible but I don't know really. `webpack-dev-server` shows address always on first compilation.

We kinda don't inform users after they fixed the errors if anything occurred. Maybe the output should look like this:
```
You can view the application in browser.

   Local: ...
   On Your Network: ...

Build failed!
--Some error message--
Compiled successfully --date stamp--
Compiled successfully --date stamp--
Build failed!
--Some error message--
Compiled successfully --date stamp--
```

Let me know what you think (also @lukeed @developit @thangngoc89 )

thangngoc89: Only show live URL on first render is more than enough for me. And I really hate any watcher that clear my terminal output

developit: I'm good with that format.  I believe the "compiled successfully" bit can be done as the completion message of ProgressBarPlugin if it isn't already.

src/lib/webpack/run-webpack.js
```diff
@@ -121,20 +127,20 @@ export function writeJsonStats(stats) {
...
 	if (m.reasons) {
-		m.reasons.forEach(normalizeModule);
+		m.reasons.forEach(stripBabelLoaderFromModuleNames);
 	}
 
 	return m;
 };
 
-const normalizeName = p => p.substr(p.lastIndexOf('!') + 1);
+const stripBabelLoaderPrefix = log => log.replace(/@?\s*(\.\/~\/babel-loader\/lib\?{[\s\S]*?}!)/g, '');
```
developit: I have a boatload more of these transforms we can add, I'll get clearance to release them shortly.

src/lib/webpack/run-webpack.js
```diff
@@ -26,23 +26,21 @@ const devBuild = async (env, onprogress) => {
 
 	let compiler = webpack(config);
 	return await new Promise((resolve, reject) => {
-		let first = true;
+
 		compiler.plugin('done', stats => {
-			if (first) {
```
reznord: Removed this since it is better to clear off the console when there is an error and show the network interface when build is successful (every time)

src/lib/webpack/run-webpack.js
rkostrzewski: It is now kinda different with @reznord's changes. Basically we would now show address only for successful builds and clear console after each build (IMHO that's much nicer).

reznord:
@thangngoc89 @lukeed @developit this is good for your review !!

reznord:
Wohoo !! Here we go !!

PR #207, opened 2017-07-08
Fix initial commit when git user email is note defined.

### LINKED ISSUE #202: Git initialization error
There is a strange issue during initialization of git repo. The only output in the command line is:

```
✔ Done!

Error: Exited with status 128
```

I've investigated a bit, and it appears that this line of code is cause of the issue.

```
await spawn('git', ['commit', '--author', gitUser, '-m', 'initial commit from Preact CLI'], { cwd });
```

knight-bubble:
I am sure the initial message shouldn't be suppressed, and if you know how to display it please tell me.
But that commit at least fixes #202

### DIFF
--- a/src/commands/create.js
+++ b/src/commands/create.js
@@ -215,32 +215,39 @@ async function initializeVersionControl(target) {
 		await spawn('git', ['init'], { cwd });
 		await spawn('git', ['add', '-A'], { cwd });
 
-		let currentGitUserEmail;
+		const gitEmail = 'developit@users.noreply.github.com';
+		const gitUser = 'Preact CLI';
+
+		let isUserEmailSet = true;
 		try {
-			currentGitUserEmail = await spawn('git', ['config', '--global', 'user.email']);
+			await spawn('git', ['config', 'user.email']);
 		} catch (e) {
-			/*
-				
-*** Please tell me who you are.
-
-Run
-
-  git config --global user.email "you@example.com"
-  git config --global user.name "Your Name"
-
-to set your account's default identity.
-Omit --global to set the identity only in this repository.
-			 */
+			isUserEmailSet = false;
 		}
 
-		if(!currentGitUserEmail) {
+		let isUserNameSet = true;
+		try {
+			await spawn('git', ['config', 'user.name']);
+		} catch (e) {
+			isUserNameSet = false;
+		}
+
+		if (!isUserEmailSet) {
 			await spawn('git', ['config', '--global', 'user.email', gitEmail]);
 		}
 
-		await spawn('git', ['commit', '--author', gitUser, '-m', 'initial commit from Preact CLI'], { cwd });
+		if (!isUserNameSet) {
+			await spawn('git', ['config', '--global', 'user.name', gitUser]);
+		}
 
-		if(!currentGitUserEmail) {
+		await spawn('git', ['commit', '--author', `${gitUser}<${gitEmail}>`, '-m', 'initial commit from Preact CLI'], { cwd });
+
+		if (!isUserEmailSet) {
 			await spawn('git', ['config', '--global', '--unset', 'user.email']);
+		}
+
+		if (!isUserNameSet) {
+			await spawn('git', ['config', '--global', '--unset', 'user.name']);
 		}
 	}
 }
--- a/src/commands/create.js
+++ b/src/commands/create.js
@@ -215,7 +215,32 @@ async function initializeVersionControl(target) {
 		await spawn('git', ['init'], { cwd });
 		await spawn('git', ['add', '-A'], { cwd });
 
-		const gitUser = 'Preact CLI<developit@users.noreply.github.com>';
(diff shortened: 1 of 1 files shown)


rkostrzewski:
Hi @knight-bubble thanks for raising the issue & making a PR :tada:.

I just found [this](https://stackoverflow.com/questions/22058041/commit-without-setting-user-email-and-user-name) on stack overflow. Do you think we could use this to set commit email and username if it isn't set ? I was wondering if doing sth along these lines would work:
```
let defaultUserEmail = 'developit@users.noreply.github.com'
let defaultUserName = 'Preact CLI'
let isUserSet = true

try {
  // Both throw error when they are not set 
  await spawn('git', ['config', 'user.name'])
  await spawn('git', ['config', 'user.email'])
} catch (err) {
  isUserSet = false
}

if (isUserSet) {
  await spawn('git', ['commit', '-c', `user.name=${defaultUserName}`, '-c', `user.email=${defaultUserEmail}`, '-m', 'Initial commit from Preact CLI'], { cwd });
} else {
  await spawn('git', ['commit', '-m', 'Initial commit from Preact CLI'], { cwd });
}
```

knight-bubble:
@rkostrzewski I've tried using the "-c" option too, but it seems to not work also it gives a different error, but couldn't figured out which one 😢.
But I like your solution so I'll refactor the PR :)

knight-bubble:
@rkostrzewski Running you command gives this kind of output:
fatal: Option -m cannot be combined with -c/-C/-F/--fixup.
So i think we should keep setting of global user and unset it.

rkostrzewski:
@knight-bubble oh that could be because in the original StackOverflow post there was `git commit -c .. -m 'message'`
```
await spawn('git', ['-c', `user.name=${defaultUserName}`, '-c', `user.email=${defaultUserEmail}`, 'commit', '-m', 'Initial commit from Preact CLI'], { cwd });
```

I'm kida afraid that if anything throws an error for some reason the global user won't be unset.

knight-bubble:
@rkostrzewski You are right. What if instead of using git command, we will set GIT_COMMITTER_NAME and GIT_COMMITTER_EMAIL? There should be a way to call spawn with specific env variables.

knight-bubble:
@rkostrzewski This seems to work
```
await spawn('git', ['commit', '-m', 'initial commit from Preact CLI'], {
	cwd,
	env: {
		...(!isUserNameSet ? { GIT_COMMITTER_NAME: gitUser} : {}),
		...(!isUserEmailSet ? { GIT_COMMITTER_EMAIL: gitEmail} : {}),
		GIT_AUTHOR_NAME: gitUser,
		GIT_AUTHOR_EMAIL: gitEmail
	}
});
```

developit:
Maybe can simplify that to:

```js
await spawn('git', ['commit', '-m', 'initial commit from Preact CLI'], {
	cwd,
	env: {
		GIT_AUTHOR_NAME: gitUser || defaultGitUser,
		GIT_AUTHOR_EMAIL: gitEmail || defaultGitEmail
	}
});
```

knight-bubble:
@developit Just tested without GIT_COMMITTER_NAME and GIT_COMMITTER_EMAIL doesn't work.

rkostrzewski:
@knight-bubble sounds great.
@developit I think `GIT_COMMITER` is a must and `GIT_AUTHOR` is not required (and would be taken from `GIT_AUTHOR`)

lukeed:
I'm not sure I follow --- why do we need to specify a Git author name/email? If you just removed that, it should work fine, no?

I have a few yeoman generators, and in them I just spawn `git` with a commit message, and my/user's git config info is picked up automatically.

knight-bubble:
@lukeed You are right. But if the user email is not specified, the git initialization function fails.
Try to unset user.email in git config an try to commit.
You should get the following message in the terminal:
```
*** Please tell me who you are.
Run
  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"
to set your account's default identity.
Omit --global to set the identity only in this repository.
```

lukeed:
Ah, okay. Does it throw? If so, then a `try/catch` sounds right. No point in committing if user can't commit on their own anyways~

developit:
^ I was a dolt and didn't notice the different vars, ignore me 😅

prateekbh:
@developit @rkostrzewski @knight-bubble what exactly is stopping us from merging this right away?

reznord:
This PR might be effected based on the discussions in #216

prateekbh:
oh no, This PR mainly fixes the commit author, whereas #216 mainly will just make the auto --git flag as default false.

if user still passes `--git`, running these commands would make sense

prateekbh:
@rkostrzewski @developit is this good to go? another PR is pending on this

reznord:
lgtm

_Sent from my OnePlus ONEPLUS A3003 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

PR #210, opened 2017-07-10
add lodash plugins

SaraVieira:
Hey guys,

This PR adds:
   - [Babel Plugin Lodash](https://github.com/lodash/babel-plugin-lodash)
  - [lodash-wepack-plugin](https://www.npmjs.com/package/lodash-webpack-plugin)

This allows people to this:

```
import _ from 'lodash';
import { add } from 'lodash/fp';

const addOne = add(1);
_.map([1, 2, 3], addOne);
```

And it will be transformed into:

```
import _add from 'lodash/fp/add';
import _map from 'lodash/map';

const addOne = _add(1);
_map([1, 2, 3], addOne);
```

For way smaller builds and no more `import add from 'lodash/add'


prateekbh:
hi @SaraVieira is there a bug for this, explaining why this has to be a part of core cli and cannot just reside in user land?

SaraVieira:
Hi @prateekbh ,

There is no bug because I don't think this a bug but an enhancement.

I don't think this should reside in user land mainly because lodash is one of the most downloaded npm modules ever and I think everyone had the problem of coming out with huge bundles because of not having this. Also leaving this in user land involves having any user add one babel plugin and one webpack plugin.

I'm aware this is more to add to the core but most people ( including me) sometimes forget to check if a boilerplate has this and then turn out with huge bundle sizes 😞 

This can be more seen as a plus for using the CLI than an issue I think .

Good explanation @prateekbh ?

SaraVieira:
Also this can be something added to the readme as a plus 😄

prateekbh:
>I don't think this should reside in user land mainly because lodash is one of the most downloaded npm modules ever and I think everyone had the problem of coming out with huge bundles because of not having this.

Sounds right, but even if everyone agrees to it I still think this should be added behind a flag `--lodash` so that it just doesn't add it by default to ones who does not want to use it.

reznord:
> Sounds right, but even if everyone agrees to it I still think this should be added behind a flag --lodash so that it just doesn't add it by default to ones who does not want to use it.

IMHO, I wouldn't prefer doing it (adding a flag or something). Eventhough lodash is most commonly used, we don't know if the end user wants to use it or not. 

If he really wants to add and optimize the bundle, he can add it add extra config using `preact.config.js`.

reznord:
and if you really want to use destructuring, I'd prefer adding a module which just exports required the packages from `lodash` and then you can use call the required lodash functions from there

SaraVieira:
@reznord @prateekbh 
The problem with adding a flag is that on the webpack config I don't have access to the flags so I don't know how to make it a flag.

I mean I don't know if the user will need it but it doesn't add anything in terms of size to the bundle and in my opinion is always a nice thing to have.

I'd prefer adding a module which just exports required the packages from `lodash` --> I do this where I use the preact-cli but it has happened for me to forget and import like this:

```
import { add } from 'lodash';

_.map([1, 2, 3], addOne);
```

That was why I created the PR

reznord:
> but it doesn't add anything in terms of size to the bundle 

I'm not really sure about this part. Even if you really want to just import a particular functionality of lodash, I'd always prefer importing that particular package itself by using something like this `import map from lodash/map`

> in my opinion is always a nice thing to have

Agree, but it doesn't makes sense if the user doesn't want to use `lodash`. In this case, it will be just extra load.

SaraVieira:
@reznord I get that but the thing a lot of people don't know they can do that :(

but it doesn't add anything in terms of size to the bundle --> I was saying between having the plugin or not having when not using lodash

reznord:
> a lot of people don't know they can do that :(

True, but this just adds extra complexity to the code. And I guess it is better to leave it in the user land.

Instead we can just add a wiki page for this and leave this, so it will be helpful for any user whoever wants to use lodash.

developit:
I might have missed something in the scrollback, but what about making this a plugin?  We've been throwing around the idea of having plugins that work just like our `preact.config.js`, and this could be a good fit for that model?

I kinda see why @SaraVieira is suggesting this be in core though:  the people importing all of `lodash` are probably not the people who would think to go look for `lodash-webpack-plugin`.

I had a question though:  a lot of the complexity added here is coming from `lodash-webpack-plugin`, but the real gain in size/perf is from `babel-plugin-lodash`.  Maybe we could roll with just the babel plugin?  The Webpack plugin seems to be more geared towards optimizing usage of lodash, rather than optimizing the module/import structures.

reznord:
We can add a wikipage for this and leave it in the userland too!! 

_Sent from my OnePlus ONEPLUS A3003 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

SaraVieira:
@developit True I think we can actually remove the webpack plugin and just use the babel one or create a wiki page but I believe that would alienate a lot of users :/

The plugin idea is great ! But how would that plugin be structured ? Like a preact.config.js that you import in your preact.config.js ?

developit:
Yup, it'd really just be a preact.config.js that is published to npm.

Setting up a plugin would be simple - you'd `npm install preact-cli-lodash`, then list it in your `package.json` or pass `--plugins lodash`, or invoke helpers.addPlugin() from `preact.config.js`.

SaraVieira:
Workflow:

```
npm install preact-cli-lodash
```
When installed:

preact.comfig.js
```
import lodash from 'preact-cli-lodash';

lodash()
```

Would this be okay ? @developit

reznord:
> import lodash from 'preact-cli-lodash';

It would be better support de-structuring. (IMO)

SaraVieira:
Lodash would just be the exporting name in the module, it could
preactLodashPlugin :)

SaraVieira:
@prateekbh @reznord @developit I'm gonna close this PR and create the plugin is that okay ?

developit:
I think it would have to have exports, but yeah before we have full plugin support you could totally do that:

```js
// preact.config.js
import lodash from 'preact-cli-lodash';
export default lodash;
```

rkostrzewski:
@developit @SaraVieira cool idea with plugins. Just putting in my two cents:
```
import lodash from 'preact-cli-lodash'

export default (config) => {
   lodash(config)
}
```
this way users could still customize their config. Another idea is:
```
import lodash from 'preact-cli-lodash'
import otherPlugin from 'preact-cli-other-plugin'

export const plugins = [
  lodash,
  otherPlugin
]

export default (config) => {
   lodash(config)
}
```

lukeed:
I believe this is purely userland & doesn't belong in the core `preact-cli` toolkit since it's 1) not part of Preact and 2) not relevant to PWA code.

However! I think this should be a main highlight of the new (WIP) Recipes section that @developit started.

SaraVieira:
Closing the issue and creating the plugin I will then create a PR for changes on the readme. 

Thanks for the feedback guys!

developit:
@rkostrzewski one huge perk of `--plugins foo,bar,baz` would be that we can auto-pass them the helpers.  Reduces a lot of boilerplate.

reznord:
> one huge perk of --plugins foo,bar,baz

Oh my !! Another flag !! (life is full of flags !! xD)

reznord:
I really should check for a package which can manage all the flags !! (I'm sure there should be one)

PR #212, opened 2017-07-10
Add preact-cli-lodash to Readme.md

SaraVieira:
Freshly created the first plugin , any feedback would be awesome

https://www.npmjs.com/package/preact-cli-lodash

cc @prateekbh @developit @reznord @rkostrzewski

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,15 +116,7 @@ Preact CLI does this by rendering your app inside node - this means that we don'
 
 ### Custom Configuration
 
-#### Plugins
-
-On an effort of making customization easier, we are also creating plugins that allow custom configuration just from installing.
-
-Current Plugins:
-
-- [Preact CLI Lodash] (https://www.npmjs.com/package/preact-cli-lodash): Optimize your builds using `babel-plugin-lodash` and `lodash-webpack-plugin`.
-
-
+#### [Plugins](https://github.com/developit/preact-cli/wiki/Plugins)
 
 #### Browserslist
 
--- a/README.md
+++ b/README.md
@@ -118,9 +118,11 @@ Preact CLI does this by rendering your app inside node - this means that we don'
 
 #### Plugins
 
-On an effort of making customisation as easier as we can we are also creating plugins that allow custom configuration just from installing. Current Plugins are:
+On an effort of making customization easier, we are also creating plugins that allow custom configuration just from installing.
 
-- [Preact CLI Lodash] (https://www.npmjs.com/package/preact-cli-lodash): Optimize your builds using babel-plugin-lodash and the lodash webpack plugin.
+Current Plugins:
+
+- [Preact CLI Lodash] (https://www.npmjs.com/package/preact-cli-lodash): Optimize your builds using `babel-plugin-lodash` and `lodash-webpack-plugin`.
 
 
 
--- a/README.md
+++ b/README.md
@@ -116,7 +116,9 @@ Preact CLI does this by rendering your app inside node - this means that we don'
 
 ### Custom Configuration
 
-#### [Plugins](https://github.com/developit/preact-cli/wiki/Plugins)
+#### Plugins
+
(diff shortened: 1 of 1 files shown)


developit:
amaze

developit:
I think the plugin is missing a build script though, I don't see a lib/dist/whatever folder in the package:
https://unpkg.com/preact-cli-lodash@1.0.1/

I see babel is in the deps, so maybe its just missing `"build": "babel src -d lib"` and `"main": "lib"`?

SaraVieira:
That script existed but I removed it since the preact config supports es6.

Why do you think it would be good to have it @developit?

developit:
Doesn't entirely need it, but the `src` directory is missing too - there's no code in the package right now haha

SaraVieira:
Effing npm ignore that comes by default -.-

developit:
haha yeah. I've stopped using npmignore in favor of `"files"` in the package.json

SaraVieira:
@developit updated , sorry 😢 

https://unpkg.com/preact-cli-lodash@1.1.0/

SaraVieira:
Can I add this one too : https://github.com/SaraVieira/preact-cli-postcss ?

reznord:
Doesn't makes sense to me !!

IMO, no one is going to have a look at the plugins until they need. Instead of putting the documentation for plugins in the main README, it is better to add a separate page `plugins.md` and add all those in that. 

This way the README stays clean too !!

SaraVieira:
Dude I think that's is totally fine, I don't know if you are yelling but it sounds like you are and there is no need for that. 

This was just a proposal, I can create a plugins.md. 

I was just trying to help dude

developit:
I'm fine with either, or maybe we keep the Plugins heading in the readme but point to a Wiki page that lists them?

I created the page and added both plugins 🎉
https://github.com/developit/preact-cli/wiki/Plugins

SaraVieira:
@developit @reznord Updated the Readme to just point to the wiki page 👍

reznord:
Cool, now I guess we can add [preact-cli-postcss](https://github.com/SaraVieira/preact-cli-postcss) plugin too. 😄

SaraVieira:
@developit Already added it to the page 😄

rkostrzewski:
@developit @reznord @SaraVieira should I add to `Plugins` section of `README.md` what's to come (doing `preact build --plugins lodash,postcss,awesome` and `plugins` section in `package.json`) ? Looks kinda empty right now 😛 

Also @developit @reznord also thinking about doing two things exposing `webpackHelpers` from `preact-cli` npm module and exposing `createWebpackConfig` so that plugins could be used as such
```
import awesome from 'preact-cli-plugin-awesome'

export default config => {
   awesome(config)
}
```

and  the plugin would look like this:
```
import { webpackHelpers } from 'preact-cli'

export default config => {
  let loaderWrapper = webackHelpers.getLoadersByName('some-loader')[0]
  ...do awesome stuff...
}
```

and exporting sth like `createWebpackConfig` so users can do sth like this when developing:

```
const createWebpackConfig = require('preact-cli'). createWebpackConfig

let env = { cwd: '.', ... }
let config = createWebpackConfig(env)
debugger;
console.log(createWebpackConfig)
```
when developing plugins & that would be useful for stuff like `eslint-plugin-import` or `storybook` (each requires `webpack.config.js` to work properly with `preact-cli`)

prateekbh:
@rkostrzewski the syntax
```
webackHelpers.getLoadersByName/getPluginByName('some-loader')[0]
``` 
looks great 👍 

just a small thing 
```
export default config => {
   awesome(config)
}
```

⬆️ should have a return statement to keep things clearer

SaraVieira:
@rkostrzewski That sounds like a great idea and yes filling that page would be awesome 👍

SaraVieira:
@rkostrzewski thanks man! 

Is everyone okay with merging this? 🙌

_Sent from my Bq Aquaris M5 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

developit:
All good!

SaraVieira:
🎉🍕

PR #222, opened 2017-07-13
Update readme to reflect actual build command

SaraVieira:
In the Readme we say we can use the --less and --sass flag but it doesn't exist like you can see:

screenshot 2017-07-13 09 37 15


reznord:
Please add `[skip ci]` to the end of the commit message next time to skip travis tests. Since it is just a change in `README.md`, no need to test it. 😄

SaraVieira:
@reznord I will do the magic next time , sorry didn't know that did anything 😱

reznord:
Cancelled the build since it is just a change in README.

PR #225, opened 2017-07-13
feature: yarn install (through a flag)

reznord:
This PR adds support for yarn in the application. 🎉 

User can install & use yarn by adding a flag `--yarn` when he initializes a project using `preact create my_app --yarn`.

Added `install-dependencies.js` since I want to keep `create.js` clean and simple.

For now cancelling the tests since they will fail for sure 😛 

Tests are the only thing that are failing now, else we are good to go with `yarn` 😎 

@rkostrzewski @developit please help with the tests !!


reznord:
Closing #31 in favour of this

reznord:
Here is the latest log from the failed tests - https://gist.github.com/reznord/f69ddcf6f0caad99bb9d83ce61381ca8

Progress: Tests are not failing that worse 😛

src/lib/webpack/dependency-install-loader.js
```diff
@@ -30,19 +31,15 @@ const CACHE = {};
...
 }
 
 function installDeps(deps, save) {
 	process.stdout.write(`\nInstalling ${deps.join(' ')}..`);
-	return spawn('npm', [
-		'install', save && (save==='dev' ? '--save-dev' : '--save'),
-		'--prefix', process.cwd(),
-		...deps
-	].filter(Boolean)).then( () => {
+	return install(false, process.cwd(), deps, save).then( () => {
```
developit: Would it make sense to allow yarn here if it's installed?  We could check for a `yarn.lock` as an indication of whether the current repo is set up with yarn.

Issue #208, opened 2017-07-09
Rebuild the application after installing a new package


reznord:
So, right now when I add a new package to the project, I have to restart the dev build, so that I can use the package that I've added. 

It would be great if we can trigger a rebuild when a new package is added. CRA has this functionality and it is quite useful. 

Any suggestions @developit @lukeed @rkostrzewski

knight-bubble:
@reznord In the CRA it's done by https://github.com/facebookincubator/create-react-app/blob/master/packages/react-dev-utils/WatchMissingNodeModulesPlugin.js
And called like this https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/config/webpack.config.dev.js#L272

reznord:
@knight-bubble I've seen this. Asking if adding this will get or helpful or not.

SaraVieira:
@reznord how about a postintall script in the package.json that runs preact build? 

_Sent from my Bq Aquaris M5 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

reznord:
I don't think that will work, since the user may install a package from a new terminal instance. In that case the postinstall script doesn't work.

developit:
The CRA plugin looks neat.

reznord:
Cool, I'll take this on.

PR #227, opened 2017-07-14
rebuild application when a new package is added

reznord:
With this PR we can simply add a new package to the application and can use the package right away without restarting the local server.

Took the idea from Facebook's CRA.

Fixes - #208

### DIFF
--- a/src/lib/webpack/run-webpack.js
+++ b/src/lib/webpack/run-webpack.js
@@ -27,6 +27,18 @@ const devBuild = async (env, onprogress) => {
 	let compiler = webpack(config);
 	return await new Promise((resolve, reject) => {
 
+		compiler.plugin('emit', (compilation, callback) => {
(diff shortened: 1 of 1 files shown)


reznord:
This works perfectly when you make some changes in JS files (since the webpack catches the changes) but when you straight away install a package, this plugin won't be called.

Any suggestions @developit

thangngoc89:
> This works perfectly when you make some changes in JS files (since the webpack catches the changes) but when you straight away install a package, this plugin won't be called.

Webpack doesn't watch package.json. Watcher has its limitation

reznord:
btw, leaving a note here - we can remove this in the upcoming versions of webpack. 

Reference links:
https://github.com/webpack/webpack/pull/5297
https://github.com/webpack/watchpack/pull/54

developit:
Nice, good to know.

Issue #201, opened 2017-07-07
Better Sass/Less support for css files


reznord:
So, right now when we initiate a project, we are generating `less` by default.

Suggestion:
Generate css with `.css` extension and then when initiating a new project using `--less` or `--sass`, rename the files so that the user will not get confused when he opens the respective stylesheets

Any suggestions @developit @rkostrzewski

prateekbh:
>Move all css to plain .css files and then when initiating a new project using --less or --sass, we should rename the files so that the user will not get confused when he opens the css files.

+1 for better editor support

reznord:
We can use [renamer](https://www.npmjs.com/package/renamer) for moving the files to respective format as specified in the arguments.

thangngoc89:
@lukeed is working on a new release with support for template. We should wait until than.

prateekbh:
the head/body template?
can you refer the PR here?

yavorski:
We should keep in mind that if the files are with the same contents, they may not be compiled by the specified preprocessor less/sass/styus etc, which can lead to some confusions at the beginning.
One solution is to keep them simple, and do not use features from different css preprocessor.

thangngoc89:
@prateekbh There is no PR yet. Since we don't want to introduce thounsands flags to the CLI. 

IIRC, the result will be like `preact new my_project --template=less`

This is install all necessary dependencies and create a project with less styling in it. By copying a folder from the main repo.

reznord:
@thangngoc89 Luke's templating will be a part of 2.0.0 (major release), but until then it is better to handle this issue as I suggested. Since there are quiet a few issues where users got confused because of the default `.less` format.

yavorski:
Hi I managed to do that without adding new deps. Just used `copy` & `rimraf` which are already there, and they do the work. Hope it's okay, until we have better options mention above.

reznord:
@yavorski can you make a PR?

yavorski:
#230 overwrite #228 ... and switching `.less` files to `.css` in `exaples/full` template...

reznord:
This is good to go. Closing.

PR #228, opened 2017-07-14
Better "preact create" preprocessor files support

yavorski:
`preact create` will now rename default `.less` to `.scss` or `.styl` depending on given option.
Discussion is here #201.


reznord:
Temporary fix for #201

yavorski:
Closing in flavor to #230.

PR #230, opened 2017-07-14
Default to CSS

yavorski:
`preact create` will now default to `.css`. 

#201 

To use preprocessor use appropriate flag:

* `--less`
* `--sass`
* `--stylus`

Examples:

* `$ preact create app-css`
* `$ preact create app-less --less`
* `$ preact create app-sass --sass`
* `$ preact create app-stylus --stylus`


examples/full/src/style/helpers.css
```diff
@@ -0,0 +1,18 @@
+.fill {
```
developit: Se can probably ditch these since they aren't imported now.

yavorski: Deleted now.

yavorski:
@developit we can close #228 in flavor of this #230 ?

src/commands/create.js
```diff
@@ -193,6 +194,25 @@ export default asyncCommand({
...
+			if (sass) extension = '.scss';
+			if (stylus) extension = '.styl';
+
+			await copy(srcDir, srcDir, {
+				overwrite: true,
+				filter: ['**/*.css'],
+				rename: filePath => filePath.replace('.css', extension)
+			});
+
+			await promisify(rimraf)(`${srcDir}/**/*.css`);
```
rkostrzewski: I think we should exclude `node_modules` from here for future templates

yavorski:
@rkostrzewski this is just dealing with `src` folder, not with the `target` (root).

rkostrzewski:
@yavorski yeah it doesn't matter - the templates wouldn't be probably hitting that code anyway

src/commands/create.js
```diff
@@ -193,6 +194,25 @@ export default asyncCommand({
...
+			const { less, sass, stylus } = argv;
+			const srcDir = path.resolve(target, 'src');
+
+			let extension;
+
+			if (less) extension = '.less';
+			if (sass) extension = '.scss';
+			if (stylus) extension = '.styl';
+
+			await copy(srcDir, srcDir, {
```
developit: I wonder if there is a way to rename here instead of copy+delete?

```js
import glob from 'glob';

let cssFiles = await promisify(glob)('**/*.css', { cwd: srcDir });
let renameToCss = name => fs.rename(name, name.replace(/.css$/, extension));
await Promise.all(cssFiles.map(renameToCss));
```

reznord: How about [renamer](https://www.npmjs.com/package/renamer) ?

yavorski: @developit i would do that, but there is no `glob` in `package.json`.

reznord: we can use `renamer` though but it is recommended to use it globally. In that case, I'm not really sure if this is a good idea or not.

yavorski: Its no problem for me to reactor with `glob`.
@developit @reznord ?

developit: I'm fine adding `glob` - we'll probably want to use that module elsewhere anyway :)

@reznord I had initially looked at `renamer` too, there is a node counterpart for it - the problem was that it's synchronous.

reznord: ouch !! then it is of no use

yavorski: @developit @reznord Updated to using `glob` now.

src/commands/create.js
```diff
@@ -190,6 +191,19 @@ export default asyncCommand({
...
 		}
 
+		if (argv.less || argv.sass || argv.stylus) {
+			let extension;
+
+			if (argv.less) extension = '.less';
+			if (argv.sass) extension = '.scss';
+			if (argv.stylus) extension = '.styl';
+
+			const cssFiles = await promisify(glob)(`${target}/src/**/*.css`);
```
rkostrzewski: This will only work for the default template. Maybe we could glob the examples folder and replace result's base path with target? Dunno if it is worth the trouble. @developit we're gonna keep all the examples after 2.0?

lukeed: I was about to write this too. Could use target & remove `src` from pattern.

yavorski: @rkostrzewski @lukeed i can change it to this one: 
```js
const cssFiles = await promisify(glob)(`${target}/**/*.css`, { ignore: `${target}/build/**/*` });
```
Just tested with `root` template and it worked.

yavorski: @rkostrzewski @lukeed glob pattern updated.

vitormalencar:
awesome =)

maciej-gurban:
Could you add this to documentation somewhere?

lukeed:
@maciej-gurban This is outdated as of 2.0. The style tool is now determined by the remote-template you use/clone via `preact create`.

PR #233, opened 2017-07-16
Revert to compiling webpack configs in sequence

rkostrzewski:
This turn offs multi-compiler mode in webpack as it seems unreliable.

From [docs](https://webpack.js.org/api/node/#webpack-) :
```
webpack will not run the multiple configurations in parallel. Each configuration is only processed after the previous one has finished processing. To have webpack process them in parallel, you can use a third-party solution like parallel-webpack.
```

Which is untrue (we've had ```webpack([ssrConfig, clientConfig])``` previously and `clientConfig` was processed before `ssrConfig` - now it seems random)

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,7 +126,6 @@
     "get-port": "^3.1.0",
     "html-webpack-exclude-assets-plugin": "0.0.5",
     "html-webpack-plugin": "^2.28.0",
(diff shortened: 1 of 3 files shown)


reznord:
Although we can explore other options like - `paralle-webpack`, not sure if that fits the purpose or not.

PR #235, opened 2017-07-16
fix: Dependency install loader

rkostrzewski:
Webpack checks if all loaders are present so we have to pass source to proxy-loader

### DIFF
--- a/src/lib/webpack/proxy-loader.js
+++ b/src/lib/webpack/proxy-loader.js
@@ -4,14 +4,15 @@ var requireRelative = require('require-relative');
 function proxyLader(source, map) {
 	var options = utils.getOptions(this);
 
-	// First loader run store passed options & remove proxy-loader specific
+	// First run proxy-loader run
 	if (!this.query.__proxy_loader__) {
+		// Store passed options for future calls to proxy-loader with same loaderContext (this)
+		// e.g. calls via 'this.addDependency' from actual loader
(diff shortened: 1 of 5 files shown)


src/lib/webpack/proxy-loader.js
```diff
@@ -0,0 +1,45 @@
...
+
+function proxyLader(source, map) {
+	var options = utils.getOptions(this);
+
+	// First loader run store passed options & remove proxy-loader specific
+	if (!this.query.__proxy_loader__) {
+		this.query.__proxy_loader__ = { loader: options.loader, cwd: options.cwd };
+
+		// remove proxy-loader options and make this.query act as requested loader query
+		swapOptions(this, options);
```
developit: This is clever.

src/lib/webpack/proxy-loader.js
```diff
@@ -0,0 +1,45 @@
...
+
+	var loader;
+	try {
+		loader = requireRelative(proxyOptions.loader, proxyOptions.cwd);
+	} catch (e) {
+		loader = require(proxyOptions.loader);
+	}
+
+	// run loader
+	return loader.bind(this)(source, map);
```
developit: Could use `loader.call(this, source, map)`.

rkostrzewski: Forgot how to ES5

PR #237, opened 2017-07-17
No git

### LINKED ISSUE #216: git stage all files during initialising a repo?
I wouldn't normally prefer having an automated commit since the user might think it as a "magic".

More over, seeing from a user point of view, I wouldn't really like to commit right after creating a repo.

Leaving this with tagged discussion.

prateekbh:
Fixes #216, 
This will put --git in default state of `false` so that when user says --git he knows that there will some git activity, and we wont do aut commits without his knowledge


src/commands/create.js
```diff
@@ -56,7 +56,7 @@ export default asyncCommand({
 		git: {
 			description: 'Initialize version control using git',
 			type: 'boolean',
-			default: true
+			default: false
```
reznord: I'm not really sure if setting `--git` to false by default or not. 

Leaving `--git` to `true` by default and not making an initial commit makes sense to me.

@developit any thoughts?

thangngoc89: I think we are trying to hard here. User know how to initilize a git repo. Just leave it to user.

prateekbh:
ok, 
we have 2 options here
*Option A:* leave `--git` to true and and not make the final commit.

*Option B:* leave `--git` to false and leave everything to user.

@thangngoc89 @reznord which one do you think guys?

reznord:
@prateekbh I'm good with both, anyways for _Option B_ it is just a matter of one command away. So, I'm with @thangngoc89 here.

PR #241, opened 2017-07-18
Fix link :active/:hover color accessibility in the examples

developit:
/cc @reznord


bestwestern:
'preact create my-great-app' still doesn't have different color on the active route?

Issue #245, opened 2017-07-20
Bug: yarn start throws an error


reznord:
Package Versions:
- Preact-cli: v1.4
- node: 8.1.4

When we initiate the project using yarn by `preact create my-app-name --yarn` and then run the development server using `yarn start` it throws the following error: 

```sh
➜  test yarn start
yarn start v0.27.5
$ if-env NODE_ENV=production && yarnpkg -s serve || yarnpkg -s dev
error `install` has been replaced with `add` to add new dependencies. Run "yarn add dev" instead.
error Command failed with exit code 1.
```

PR #246, opened 2017-07-20
yarn start error fix

reznord:
when running custom scripts we have use use `yarn run -s serve`, which I forgot when I was adding yarn support.

Fixes: #245


reznord:
noice !!

developit:
Hate to resurrect this, but this means any project created using yarn can't be used with `npm` :(  It seems like most people just use `npm` in their scripts even when running with yarn.  Maybe we could just roll with that?

developit:
Ooh - so I was about to suggest we make a little module that picks `npm` or `yarn` and spawns it as a proxy, but someone already did! We could just have it use this: https://github.com/simonkberg/nyr

PR #259, opened 2017-07-25
Only remove prop types in a production build.

rbscott:
Disable 'babel-plugin-transform-react-remove-prop-types'
plugin during a development build. Some libraries
expect prop types to be available during a development
build.

Semantic UI React uses prop types during development
for validation. [Issue 1869](https://github.com/Semantic-Org/Semantic-UI-React/issues/1869) has some additional context.

### DIFF
--- a/src/lib/babel-config.js
+++ b/src/lib/babel-config.js
@@ -1,30 +1,34 @@
-export default (env, options={}) => ({
-	babelrc: false,
-	presets: [
-		[require.resolve('babel-preset-env'), {
-			loose: true,
-			uglify: true,
-			modules: options.modules || false,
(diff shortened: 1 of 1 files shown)


rbscott:
@reznord  I am not sure I fully understand how they are used, but it is a part of the AutoControlledComponent validation logic [here](https://github.com/Semantic-Org/Semantic-UI-React/blob/93deaa985/src/lib/AutoControlledComponent.js#L88). I don't fully understand it, and perhaps it is only useful when are you developing a component for Semantic-UI-React. I recently switched from react to preact, which is when the error started, so I decided to submit a PR.

PR #266, opened 2017-07-27
Revert "Resolve 0.0.0.0 to windows in localhost"

reznord:
Reverts developit/preact-cli#265

### DIFF
--- a/src/lib/webpack/run-webpack.js
+++ b/src/lib/webpack/run-webpack.js
@@ -44,9 +44,7 @@ const devBuild = async (env, onprogress) => {
 			let protocol = devServer.https ? 'https' : 'http';
 			let host = process.env.HOST || devServer.host || 'localhost';
 
-			if (host === '0.0.0.0') host = 'localhost';
-
-			let serverAddr = `${protocol}://${host}:${chalk.bold(port)}`;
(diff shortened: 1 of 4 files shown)


thangngoc89:
I'm not sure about your last change. Gonna test it tmr

src/lib/webpack/run-webpack.js
```diff
@@ -41,9 +41,8 @@ const devBuild = async (env, onprogress) => {
 
 		compiler.plugin('done', stats => {
 			let devServer = config.devServer;
-
 			let protocol = devServer.https ? 'https' : 'http';
-			let host = process.env.HOST || devServer.host || 'localhost';
+			let host = process.env.HOST || 'localhost';
```
thangngoc89: I'm not so sure about this. How about seeking another opinions? You're omitting the ability to specify host via cli argument. 

In my head, this PR is as simple as replacing 0.0.0.0 with localhost if you're on Windows (or just blindly replace it because it isn't effect *nix either) here https://github.com/developit/preact-cli/blob/beb468f3bec907b7cb1f121633e0d7e05b4b8903/src/lib/webpack/run-webpack.js#L61

PR #271, opened 2017-07-28
feature: preact init

### LINKED ISSUE #248: Interactive `preact create` command
Tagging this issue as discussion so I can get opinion on having an interactive `preact create` for initializing the project using `preact-cli`.

If we want to use interactive `preact create` i.e. `preact init` there will not be any flags. Which in my opinion is awesome. 😄 

All the flags in the current `preact create` command will be prompted as questions to the user.

reznord:
This commit adds one more commit to the preact-cli `preact init` where we will prompt questions to the user. All questions are the same as `preact create` flags.

Fixes #248

### DIFF
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -22,17 +22,26 @@ export default asyncCommand({
 
 	desc: 'Create a new application interactively',
 
-	async handler() {
+	builder: {
+		default: {
+			description: 'Use default values',
+			type: 'boolean',
+			alias: 'y',
+			default: false
+		}
+	},
+
+	async handler(argv) {
 		const questions = [
 			{
 				type: 'input',
-				name: 'appName',
+				name: 'name',
 				message: 'Directory and package name for the app',
 				default: 'my_app'
 			},
 			{
 				type: 'input',
-				name: 'destination',
+				name: 'dest',
 				message: 'Directory to create the app within',
 				default: '<appName>'
 			},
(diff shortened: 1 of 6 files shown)


reznord:
You can now start the application with default values using `preact init -y` or `preact init --default`

The following are the default values which I've set for now: 

```javascript
{
	name: 'my_app',
	dest: 'my_app',
	type: 'full',
	style: 'css',
	yarn: false,
	git: false,
	install: true,
	enableForce: false
}
```

lukeed:
As per our chat, shouldn't `name` be "My App"?

src/commands/init.js
```diff
@@ -0,0 +1,258 @@
...
+		}
+	},
+
+	async handler(argv) {
+		const questions = [
+			{
+				type: 'input',
+				name: 'name',
+				message: 'Directory and package name for the app',
+				default: 'my_app'
```
lukeed: If the `name` is both the directory _and_ package name, then what's the point of offering `dest`?

IMO, `name: 'My App'` + `dest: 'my_app'`.

In other words, `dest` defaults to `snakeCase(name)`

reznord: when we use `preact create` command for creating a project, it has `name` and `dest` key. Since those were flags, `dest` key was set to `<name>` itself.

In other words, whatever you set for `name` key, that will be set for `dest` too.

reznord:
Do I need to add tests? If so, I will just add one for `preact init`.

src/commands/init.js
developit: I think the wording here can just be changed to make things clearer:

```js
  message: 'Package name for the app'
```

... since the next question says `default: '<name>'`

PR #272, opened 2017-07-30
Look for process.env.HTTPS for devServer

### LINKED ISSUE #242: Run dev server as https
When I run the dev server via `npm start` I can access my preact app at `http://0.0.0.0:8080`.

When working with `navigator.geolocation` I get the follow console message in Chrome:
```
[Deprecation] getCurrentPosition() and watchPosition() no longer work on insecure origins. 
To use this feature, you should consider switching your application to a secure origin, such as HTTPS. 
See https://goo.gl/rStTGz for more details.
```

My only workaround at the moment is to run `npm run serve` or use Firefox on my Phone to test & develop my app.

Is it possible to run the dev server as https or add that feature in future releases?

reznord:
Now you can do `HTTPS=true npm start` for starting the development server using `https`.

We can avoid that serious regression of flags in
```
npm start -- -- --https
```

Fixes #242

### DIFF
--- a/src/commands/watch.js
+++ b/src/commands/watch.js
@@ -43,9 +43,7 @@ export default asyncCommand({
 	async handler(argv) {
 		argv.production = false;
 
-		if (process.env.HTTPS) argv.https = true;
-
-		if (argv.https) {
+		if (argv.https || process.env.HTTPS) {
 			let ssl = await getSslCert();
 			if (!ssl) {
 				ssl = true;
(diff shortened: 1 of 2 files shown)


reznord:
True! Forgot that part in the first time I committed. (Completing it now)

reznord:
> You'll need to update the `webpack-client-config` and probably `watch` command

We just need to change it in the `watch.js` file. Since from there we will pass down the config to the `webpack-client-config`

PR #305, opened 2017-08-11
full example : fixed async path 

kuldeepkeshwar:


thangngoc89:
@prateekbh Please squash and merge next time, it would reduce noises in git history

PR #312, opened 2017-08-13
watch on whole 'src' folder instead of 'components' & 'routes'

### LINKED ISSUE #309: Classnames not generated on other filestructure
A bug (or missing documentation if this is expected)

If I create a component that is not in the "default" folders (`component`/`routes`) Webpack won't generate a classname on import of a sass file.

1. Create a new Preact app with `preact create testProject --sass`
2. Move the `header` component from `/src/components/header` to `/src/header` and change the reference in the app component*
3. The styling is broken as there isn't any classname generated

*You could also create a custom components directory and move it there. Something like `/src/sections/header` instead of the regular `/src/components/header`

It should import the style and create the classname no matter where the component is located as long as the reference to the file is correct.

- node version: 8.2.1
- npm version: 5.3.0
- Operating system: MacOS Sierra 10.12.6

reznord:
Fixes #309 

Instead of watching over `components` & `routes` folder in `src`, it will look up for all the folders.

### DIFF
--- a/src/lib/webpack/webpack-base-config.js
+++ b/src/lib/webpack/webpack-base-config.js
@@ -176,8 +176,7 @@ export default (env) => {
 					{
 						test: /\.(css|less|s[ac]ss|styl)$/,
 						include: [
-							src('components'),
-							src('routes')
+							src('./')
 						],
 						loader: ExtractTextPlugin.extract({
 							fallback: 'style-loader',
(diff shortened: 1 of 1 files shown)


lukeed:
Did you run this yet? I imagine it makes for (1) slower build times and (2) more memory usage while watching.

reznord:
Yes, I did run this locally. 

I agree, this increases the build time and a lot of memory during watching, but since many users wanted this, we don't have any other option.

lukeed:
TBH, I don't like this approach. I would much rather users modify/add values as needed... but that's probably just me.

rijkvanzanten:
I get where you're coming from @lukeed. Maybe it could be enough to add a little extra snippet to the documentation explaining the cli is watching the two folders only and how that can be extended? I wasn't able to get the custom config with one extra folder working..

reznord:
Closing this PR since we can extend this using `preact.config.js`

Issue #262, opened 2017-07-25
Can't generate certificates on fresh install


ethanroday:
### Steps to reproduce:
1. `npm install -g preact-cli`
2. `preact create <app> --yarn`
3. `cd <app>`
4. `yarn serve`

### Result:
```
Setting up SSL certificate (may require sudo)...
INFO: Could not find files for the given pattern(s).
Attempting to spawn simplehttp2server to generate cert.
Failed to generate dev SSL certificate: Error: spawn C:\<app_dir>\node_modules\simplehttp2server\vendor\simplehttp2server_windows_amd64.exe ENOENT
Falling back to shared directory + simplehttp2server.
(dir: C:\Users\<username>\AppData\Roaming\preact-cli)

  server error> Error: spawn C:\<app_dir>\node_modules\simplehttp2server\vendor\simplehttp2server_windows_amd64.exe ENOENT
Error: spawn C:\<app_dir>\node_modules\simplehttp2server\vendor\simplehttp2server_windows_amd64.exe ENOENT
```

### Environment:
OS: Windows 10 x64
Terminal: Powershell (Administrator mode) (PSVersion: 5.1.14393.1480)

thangngoc89:
Hi. `ENOENT` means there is no such file. Could you run `yarn install` inside your project folder again to make sure everything get installed?

ethanroday:
I'm fairly certain the file exists...I can navigate to it in a terminal (and an Explorer window).
Also, I think the `simplehttp2server` is not the only issue here. The `devcert` library is also failing to generate the SSL certificates.

zweehn:
I have the same problem and copy pasting the ENOENT file in powershell and executing it works
`PS F:\sources\Node\my-preact> npm run serve

> my-preact@1.0.0 serve F:\sources\Node\my-preact
> preact build && preact serve

Build [==================  ] 91% (1.1s) additional asset processing
  Total precache size is about 144 kB for 14 resources.
Setting up SSL certificate (may require sudo)...
INFORMATION: Es konnten keine Dateien mit dem angegebenen
Muster gefunden werden.
Attempting to spawn simplehttp2server to generate cert.
Failed to generate dev SSL certificate: Error: spawn F:\sources\Node\my-preact\node_modules\simplehttp2server\vendor\simplehttp2server_windows_amd64.exe ENOENT
Falling back to shared directory + simplehttp2server.
(dir: C:\Users\User\AppData\Roaming\preact-cli)

[...]

PR #315, opened 2017-08-15
Allow users to specify homepage variable in package.json

ethanroday:
- Allow users to specify a `homepage` variable in `package.json`, referring to the public path under which the app is served in production. This may be a relative path or absolute URL, though only the pathname is relevant. The value of `homepage` is ignored in non-production environments. 
- Use `homepage` as Webpack's `publicPath`.
- Expose `process.env.HOMEPAGE` so users can use the variable in their own code.
- Prefix all asset paths (including the service worker path and mainfest path) with the public path.


ethanroday:
@reznord Fixed the indentation issues (sorry, tabs vs. spaces). 

Regarding the failed checks, I can't think of why the service worker check would succeed on Node 6 and fail on Node 8. Unfortunately, I can't actually run the tests locally because I'm on a Windows machine and #262 prevents the `preact serve` tests from running. But I did just `create`, `build`, and `http-server` (with a proxy redirect through Fiddler to a domain with valid certificates) using Node 8 and the service worker registered just fine in Chrome.

I just looked at the pending checks for my latest commit, and the Node 6 check (which succeeded on the previous commit) seems to be failing all over the place. This is very confusing, since the only difference between those two commits was spacing.

ethanroday:
Meanwhile, the most recent Node 8 check is passing this time.
Any ideas why this might be happening? I just double-checked the diffs and it's just spacing.

reznord:
I'm gonna test this locally in my PC and will let you know.

reznord:
@ethanroday the tests pass perfectly in my local machine, tested it on node v6, v7, v8 too.

ethanroday:
Huh. @lukeed? @developit? Any ideas on why these tests would be succeeding locally but failing in Travis?

lukeed:
Admittedly, I know nothing about our tests 😞 Pinging the resident mastermind @rkostrzewski

lukeed:
Makes me wonder if we should start a `"preact-cli"` config key within `package.json`, instead of adding keys here & there.

lwakefield:
Perhaps this could also be solved setting [`publicPath` directly](https://webpack.js.org/guides/public-path/). This would help keep custom configuration close together. Then this could be achieved like follows:

```javascript
// preact.config.js
    const BASE_URL = env.production ? '/my-gh-pages-url' : ''
    config.output.publicPath = BASE_URL
}
```

lukeed:
@lwakefield Correct, I prefer that approach to be honest, but I'm not sure what others feel about it.

lwakefield:
Whatever the outcome, it may not be a bad idea to consolidate as much of the configuration as possible. Currently, it looks like there are 5 different ways to configure the preact-cli build:
 browserlist, babel, webpack, pre-render and template.

ethanroday:
@lwakefield That's actually what I've been doing as a workaround for now, but it leaves two problems unsolved:
1. The path to the service worker is hardcoded as `/sw.js` and doesn't care what the public path is, and 
2. The template doesn't use the public path in the path to them manifest (this could be solved by a custom template, but I figured I would also fix it in this PR).

I see your point about there being too many entries for configuration - and this PR just adds another one (in the form of a `homepage` entry in `package.json`). I agree that only having to set the public path in `preact.config.js` would be vastly preferable, but somehow `entry.js` still needs to get the value of that variable so it can set the service worker path correctly (that's currently being achieved with `process.env.HOMEPAGE`). In your scenario, how does `entry.js` get access to the public path (short of requiring the user to define a specifically-named Webpack constant in their `preact.config.js`)?

lwakefield:
@ethanroday, I agree - I ran into similar issues with `/sw.js` and I think this PR currently addresses this. Are we able to do the following to do away with `process.env.HOMEPAGE`:

```javascript
// lib/entry.js
if (process.env.NODE_ENV === 'development') {
	require('preact/devtools');
} else if ('serviceWorker' in navigator && location.protocol === 'https:') {
	navigator.serviceWorker.register(__webpack_public_path__ + 'sw.js');
}
```

ethanroday:
@lwakefield A significantly more elegant solution! One note: in theory, the Webpack public path can be an absolute URL. In those cases, you would need to check to make sure it was the same domain as the current location before attempting to register the service worker.

With that change in place, though, I would be happy to abandon this in favor of a re-opened #323.

lukeed:
I agree. To make things simpler (re: `/sw.js`), let's just add the `publicPath` to the [prerender options](https://github.com/developit/preact-cli/blob/c2af0f6b707a2a3f5196684eb06ec2cf5e671646/src/lib/webpack/webpack-client-config.js#L224-L226) so that it's available to the template.

A quick-n-dirty fix is to add:

```js
var baseUrl = '<%= htmlWebpackPlugin.options.publicPath %>';

/// entry.js
navigator.serviceWorker.register(baseUrl + 'sw.js');
```

Exposing/mounting `baseUrl` is safe since it's essentially equivalent to `location.href` on the first load.

ethanroday:
@lukeed, are you suggesting just adding `baseUrl` as a global variable in a script tag in `template.html`? Why is that preferable to @lwakefield's `__webpack_public_path` solution? Or am I misunderstanding you?

ethanroday:
@rkostrzewski, any thoughts on the failing tests?

jojo05:
Hi,
I did a build using "homepage": "pwa" and the preact.config.js from @lwakefield above and all URLs are prefixed properly except manifest.json.  
We also need to figure out how to reconfigure the Routes (prefixing with pwa doesn't work either)

link rel="manifest" href="/manifest.json"
link rel="shortcut icon" href="/pwa/favicon.ico"
link href="/pwa/style.a1486.css" rel="stylesheet"

body
script defer="defer" src="/pwa/bundle.1405f.js"
script window.fetch||document.write('<script src="/pwa/polyfills.a668a.js"

developit:
We could use DefinePlugin to make `PUBLIC_PATH` available to script, then use that for the serviceworker URL.

lwakefield:
I would need to check that it works, but are you thinking of something like:

```javascript
webpack.definePlugin('PUBLIC_PATH', JSON.stringify(__webpack_public_path__))
```

jojo05:
Any idea why the manifest.json is not properly prefixed?

Also, shouldn't both the dev an production show the real path ? 
(I don't see a problem with dev showing 0.0.0.0:8080/pwa/)

The homepage option is really useful because any website can set up a PWA version of the site in a separate path. So, the Router integration is also key.

rodush:
Guys, any plans to solve this issue in the nearest couple days? I wouldn't want to create a fork for myself because of this... Thanks!

reznord:
@rodush hey, for now you can just use `preact.config.js` which is suggested here: https://github.com/developit/preact-cli/pull/315#issuecomment-323397020

rodush:
@reznord It doesn't help, as the SW path is hardcoded in entry.js - `/sw.js`, this is the main problem.

reznord:
Ah true, my bad!

lwakefield:
Hey, I reopened #323 as an alternative approach.

With respect to https://github.com/developit/preact-cli/pull/315#issuecomment-325124297, after some experimenting, it is difficult to define `PUBLIC_PATH` without introducing another config/env variable. However, it is more or less a one-liner to expose`PUBLIC_PATH` in `preact.config.js`.

lukeed:
We decided to go with the path in #323 since it was simpler & doesn't involve any additional `package.json` reliance. 

Thank you for your PR @ethanroday ❤️  It sparked some good ideas & conversations. We'll be revisiting those points in the near future, but for now, we're just going with the path of least resistance.

🙇

Issue #310, opened 2017-08-12
Build can't find node_module ending in .js or starting with @


rijkvanzanten:
**Do you want to request a *feature* or report a *bug*?**
Report a bug

**What is the current behavior?**
Trying to production build an app which uses a Node module which ends in `.js` (like [`headroom.js`](https://npmjs.com/headroom.js)) results in a `ENOENT` error

**If the current behavior is a bug, please provide the steps to reproduce.**
Import a Node module ending in `.js` into your Preact app

**What is the expected behavior?**
It should require the module from `node_modules/headroom.js`

**Please mention other relevant information.**
- node version: 8.2.1
- npm version: 5.3.0
- Operating system: MacOS Sierra 10.12.6

thangngoc89:
How do you import your module? Could you generate a simply reproduce repo so I can clone and play with it?

rijkvanzanten:
Hi @thangngoc89 

I import it the "regular" way:
`import Headroom from 'headroom.js';`

I just now noticed it also fails when I import an organization package like
`import Player from '@vimeo/player';`

I'll create a repo for you. I'll post a comment here with a link when it's done 🙂

thangngoc89:
that's weird. I'll take a look. btw, I usually use https://github.com/KyleAMathews/react-headroom . it's super  small with declarative API

rijkvanzanten:
Yeah I just changed to that one (it misses some of the options however) 🙂 

I've uploaded a small demo on https://github.com/rijkvanzanten/preact-cli-import-demo

It is the default `preact create` app, with the import added to the home component https://github.com/rijkvanzanten/preact-cli-import-demo/blob/master/src/routes/home/index.js

thangngoc89:
This looks like a bug in ours' implementation. Investigating it

reznord:
@rijkvanzanten does it work fine when you pass a flag `--no-prerender` ?

thangngoc89:
@reznord it doesn't work in dev mode too.

rijkvanzanten:
@reznord the no prerender flag doesn’t change anything

hassanbazzi:
no pre-render does fix it for me btw. Having this issue with `import player from 'player.js'`

developit:
I'm seeing it work without prerendering, so this makes me thinks it's related to our use of `target:'node'`.

reznord:
It failed for me when I tried with `preact build --no-prerender`

developit:
[...]

PR #317, opened 2017-08-16
Edge case fix for prerender errors

developit:
This adds error and node_modules lookup handling to prerender exception hook.  As it turns out, #310 was actually a legit error: `@vimeo/player` and `headroom.js` both try to access `window.xx` on import, so they can't be imported in an SSR context at all.  That was all going as planned, but the nice error handler we have for SSR was bailing because it generated an incorrect sourcemap filename.  I've added some (possibly non-windows-friendly?) edge case handling to the filename generation, basically implementing rudimentary support for webpack's `~/foo` lookup style to point those at `node_modules/foo`.


lukeed:
Sorry, I looked thru this but forgot to mark _Approve_  /facepalm

PR #320, opened 2017-08-17
Fix linting errors in templates.

reznord:
This fixes all the lint issues in templates (demos).

### DIFF
--- a/examples/full/src/components/header/index.js
+++ b/examples/full/src/components/header/index.js
@@ -2,6 +2,7 @@ import { h, Component } from 'preact';
 import { Link } from 'preact-router/match';
 import style from './style';
 
+// eslint-disable-next-line
 export default class Header extends Component {
 	render() {
 		return (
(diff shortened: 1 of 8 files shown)


examples/full/src/components/header/index.js
```diff
@@ -2,6 +2,7 @@ import { h, Component } from 'preact';
 import { Link } from 'preact-router/match';
 import style from './style';
 
+// eslint-disable-next-line
```
reznord: Added this line because, currently we don't have a `eslint-preact-preset`. 

As of now we are using `eslint-config-synacor`.

SaraVieira: @reznord is that in the plans? I could help with that! Wanted to learn how to make eslint plugins and configs

_Sent from my Bq Aquaris M5 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

reznord: I was talking to @developit regarding this, he said he will soon add one. But I took care of the comment in another PR (up-coming). Added an extra set of rules to the `eslintConfig`

reznord:
Closing this PR since now we are taking a whole different approach.

PR #322, opened 2017-08-18
make interop conditional more clear

hassanbazzi:


src/lib/entry.js
```diff
@@ -9,7 +9,7 @@ else if ('serviceWorker' in navigator && location.protocol === 'https:') {
 }
 
 
-const interopDefault = m => m && m.default || m;
+const interopDefault = m => m && m.default ? m.default : m;
```
reznord: This is easy to understand.

👍 LGTM!

thangngoc89:
The way I see it is extractly the same. This is actually common in JS codebases. Try jQuery :D

src/lib/entry.js
developit: but it takes more space :(

developit:
heh yeah they look the same to me

src/lib/entry.js
reznord: true, but easy to understand ;)

lukeed:
I'm a fan of the `&&||` 😉 so i must be one too

developit:
we are mutants

lukeed:
or simulations 💻

reznord:
Count me in !! ;)

PR #323, opened 2017-08-18
Make template.html and sw.js use publicPath

lwakefield:
This PR aims to respect the use of `webpack.output.publicPath` and is intended as an alternative to #315.

HTMLWebpackPlugin handles injecting the rest of the resources, so that you can set `output.publicPath` and the generated `index.html` will reflect the change.

The intent here is to prefer customizability through the already exposed `preact.config.js` instead of adding new config/env flags.

I also propose that the following snippet be added to [Config Recipes Wiki Page](https://github.com/developit/preact-cli/wiki/Config-Recipes) under the title "Specify the root url to serve from"

```javascript
// preact.config.js
import { defineConstants } from '@webpack-blocks/webpack2';

  const PUBLIC_PATH = env.production ? '/my-gh-pages-url' : ''
  config.output.publicPath = PUBLIC_PATH
  config.plugins.push(defineConstants({ PUBLIC_PATH }))
}
```


reznord:
There's already a PR for this. Just that it doesn't touch `preact.config.js`.
 
You can follow it on #315

lwakefield:
Ah, my mistake @reznord! I'll close this in favour of #315

src/resources/template.html
```diff
@@ -6,7 +6,7 @@
 		<meta name="viewport" content="width=device-width,initial-scale=1">
 		<meta name="mobile-web-app-capable" content="yes">
 		<meta name="apple-mobile-web-app-capable" content="yes">
-		<link rel="manifest" href="/manifest.json">
+		<link rel="manifest" href="<%= htmlWebpackPlugin.files.publicPath %>manifest.json">
```
rodush: For me it works right now - 
`<link rel="manifest" href="<%= __webpack_public_path__ %>manifest.json">`,
why can't we use this?

src/lib/entry.js
```diff
@@ -5,7 +5,8 @@ if (process.env.NODE_ENV==='development') {
 	require('preact/devtools');
 }
 else if ('serviceWorker' in navigator && location.protocol === 'https:') {
-	navigator.serviceWorker.register('/sw.js');
+	// eslint-disable-next-line no-undef
+	navigator.serviceWorker.register(__webpack_public_path__ + 'sw.js');
```
rodush: @developit Please take a look at this and decide finally about the solution. I really need this feature in my project :-) 
In my opinion it is a nice solution, what are your concerns?

rodush:
@lwakefield I don't understand, how the description of this PR correlates to the patch code?

lwakefield:
@rodush I'm not fully sure I understand why you think the PR description doesn't relate to the patch code? Please feel free to suggest edits!

In the meantime, I have been using [@lwakefield/preact-cli](https://www.npmjs.com/package/@lwakefield/preact-cli) which has these changes. Maybe it can be of help, or you can create your own namespaced package?

src/resources/template.html
lwakefield: I wanted to stay away from `__webpack_public_path__` where possible. I think it is generally bad practice to be reading from that as it is mutable and not so reliavle. For `lib/entry.js` I used `__webpack_public_path__` because it is the only available reference in the context.

I understand the inconsistency, but let me know what you think!

rodush: Yes, it is mutable, and this how webpack itself proposes to use this variable if the `publicPath`'s value is not known at compilation time, so you can even set it in the entry point.

rodush:
@lwakefield  I prefer not to get locked to custom fork unless it is 100% unavoidable 😉 
Since this is quite young project, would be very nice to have this important option out of the box, that is why I'm hoping that the owner (@developit) will make a final decision.

lwakefield:
I understand. Perhaps, you can branch yourself and change the package when this (or another PR) is merged?

rodush:
@lwakefield here `const PUBLIC_PATH = env.production ? '/my-gh-pages-url' : ''` I would suggest '/' as default, otherwise you risk to have a relative urls ;-) that is what I was fight with with an Angular build to be served from subfolder

rodush:
@lwakefield any update on this? Any plans about merging this PR? @developit ?

lwakefield:
@rodush, still waiting on review of either this or #315.

src/lib/entry.js
developit: I'm fine with this, webpack will never change this var name and it keeps the resolution at runtime.

tschnoelzer:
Are there any updates on this? We need to use subdirectories for a couple of WebApps. I would love to use preact as we did lots of prototyping already. What is the latest idea to use subdirectories? I followed this tread and also the other related ones... not clear what is current state ....

imaitland:
Did the recipe get added to the wiki?
> I also propose that the following snippet be added to [Config Recipes Wiki Page](https://github.com/developit/preact-cli/wiki/Config-Recipes) under the title "Specify the root url to serve from"

I see a test for this so assume it's the accepted way:
https://github.com/preactjs/preact-cli/blob/7d33cd1380b5dc7d4b4b970eceabfc8f126da1af/packages/cli/tests/subjects/public-path/preact.config.js

rschristian:
A recipe was never added, no. but indeed, that's the way to alter the public path if you need to.

PR #324, opened 2017-08-18
Adds examples to the Readme [skip ci]

### LINKED ISSUE #67: (nice to have) Document what each type of project mean
It might be nice to have a quick document that explains the difference between each project type.
For example - the default is using CSS but the full is using less.

SaraVieira:
@rkostrzewski @developit @reznord 

I can't figure out the difference between root and full so I only uploaded the other 3.

This is just to give people the idea of each example is.

closes #67

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,25 @@ $ preact serve
 
 ```
 
+### Templates
+
+### Full: 
+
+- Preact Router Included
+- Dynamic routes
+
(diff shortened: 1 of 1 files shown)


reznord:
Can we move this to the FAQs?

reznord:
Naaah, putting this in FAQ won't be of help. Let it be in readme 😁

PR #349, opened 2017-09-03
Configure css-loader using an object so changing the config is easier.

### LINKED ISSUE #303: WebpackConfigHelpers: getLoadersByName and css-loader
_WebpackConfigHelpers / Webpack base config feature request_

Background: I’d like to modify the behavior of css-loader. I’d like to change the `localIdentName` in development mode. I’m having a hard time getting and changing to the right rule in the gigantic Webpack loaders structure.

Ideally, I’d like to use …

```js
helpers.getLoadersByName(config, 'css-loader')
```

But this doesn’t work since css-loader is configured using Webpack-1-style query parameters …

```
`css-loader?modules&localIdentName=[local]__[hash:base64:5]&importLoaders=1&sourceMap=${isProd}`,
```

https://github.com/developit/preact-cli/blob/master/src/lib/webpack/webpack-base-config.js#L185

… and the helper only does a `=== name` check:

```
.filter(({ loader }) => loader === name || (loader && loader.loader === name));
```

https://github.com/developit/preact-cli/blob/master/src/lib/webpack/transform-config.js#L126

We could use `.indexOf(name) === 0` or similar to check whether the string starts with the name.

But I rather propose to consistently use the Webpack-2-style notation in webpack-base-config.js, e.g.

```js
{
  loader: 'css-loader',
  options: {
    modules: true,
    localIdentName: '[local]__[hash:base64:5]',
    importLoaders: 1,
    sourceMap: isProd
  }
},
```

Then `helpers.getLoadersByName(config, 'css-loader')` works as expected since the `loader === name` check matches.

Also it’s simpler to change the loader options, it’s just `loader.options.localIdentName = '…'`.

What do you think? I can prepare a PR if you agree. Let me know if I missed something.

molily:
Fixes #303.

I’ve tested it manually by adding …

```js
	const { loader: cssLoader } =
		helpers.getLoadersByName(config, 'css-loader')[0];
	cssLoader.options.localIdentName = '[path][local]-[hash:base64:5]';
```

… to `tests/subjects/custom-webpack/preact.config.js`.

Does it make sense to add this to the test setup permanently? I would add a CSS file then to demonstrate it and update the snapshot. It would be a nice example / test for `helpers.getLoadersByName` as well. What do you think?


lukeed:
Just letting it run~ Nothing changed, but better safe than sorry 😉

Issue #358, opened 2017-09-05
Ensure valid `name` key within `package.json`


lukeed:
Currently, can end up with this:

```json
{
  "name": "Hello World",
  "version": "1.0.0",
  "main": "index.js"
}
```

... which will throw an error when you try to `yarn/npm install`

Issue #352, opened 2017-09-04
Prompt user if insufficient args


lukeed:
If not enough data has been passed to the `preact create` command, ***should we*** prompt the user for the data?

Currently, Yargs will exit and print help text along with ~`Not enough non-option arguments: got 0, need at least 2`.

If we _were_ to prompt users for missing data, we'd also print an info message along these lines:

```js
warn('Insufficient command arguments! Prompting...');
info('Alternatively, run `preact create --help` for usage info.');
```

What should we do?

[![](https://m131jyck4m.execute-api.us-west-2.amazonaws.com/prod/poll/01BS6MAVSEG63WE5EYHMS62X3T/Prompt%20for%20REQUIRED%20items%20only)](https://m131jyck4m.execute-api.us-west-2.amazonaws.com/prod/poll/01BS6MAVSEG63WE5EYHMS62X3T/Prompt%20for%20REQUIRED%20items%20only/vote)
[![](https://m131jyck4m.execute-api.us-west-2.amazonaws.com/prod/poll/01BS6MAVSEG63WE5EYHMS62X3T/Prompt%20for%20ALL%20items)](https://m131jyck4m.execute-api.us-west-2.amazonaws.com/prod/poll/01BS6MAVSEG63WE5EYHMS62X3T/Prompt%20for%20ALL%20items/vote)
[![](https://m131jyck4m.execute-api.us-west-2.amazonaws.com/prod/poll/01BS6MAVSEG63WE5EYHMS62X3T/No%20prompt%3B%20Exit)](https://m131jyck4m.execute-api.us-west-2.amazonaws.com/prod/poll/01BS6MAVSEG63WE5EYHMS62X3T/No%20prompt%3B%20Exit/vote)

> PS: If we do want to prompt, I have this completed in a local branch.
 
---

_Extracted from #329 for clarity._

PR #359, opened 2017-09-05
Interactive Prompts

lukeed:
During `preact create`, if user does not provide the required arguments (`template` and `dest`) we will run them thru the list of ***all*** fields (6 total). This ensures we have all the info we need & exposes the user to all available options that the may have missed.

This is what it looks like:

screen shot 2017-09-05 at 12 14 30 pm

Prompting all fields as a result of the Poll in #352

---

_Closes #358, #352_

### DIFF
--- a/src/lib/setup.js
+++ b/src/lib/setup.js
@@ -61,3 +61,25 @@ export async function initGit(target) {
 		warn('Could not locate `git` binary in `$PATH`. Skipping!');
 	}
 }
+
+// Formulate Questions if `create` args are missing
+export function isMissing(argv) {
+	let out = [];
+
+	const ask = (name, message, val) => {
+		let type = val === void 0 ? 'input' : 'confirm';
+		out.push({ name, message, type, default:val });
+	};
+
+	// Required data
+	!argv.template && ask('template', 'Remote template to clone (user/repo#tag)');
+	!argv.dest && ask('dest', 'Directory to create the app');
+	// Extra data / flags
+	!argv.name && ask('name', 'The application\'s name');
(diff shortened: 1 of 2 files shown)


src/lib/setup.js
```diff
@@ -61,3 +61,25 @@ export async function initGit(target) {
...
+export function isMissing(argv) {
+	let out = [];
+
+	const ask = (name, message, val) => {
+		let type = val === void 0 ? 'input' : 'confirm';
+		out.push({ name, message, type, default:val });
+	};
+
+	// Required data
+	!argv.template && ask('template', 'Remote template to clone (user/repo#tag)');
```
reznord: Better to show the default value here. Like what we did for `preact init`?

reznord: basically its like having a fallback value for all the prompts

lukeed: There is no default value -- it's required.

lukeed:
Oh, @reznord, here's another screenshot of the prompt _before_ answering. It shows `N` because `false` is the default here.

screen shot 2017-09-05 at 10 59 22 pm

reznord:
True! That is what we can do for the input fields too right?

reznord:
@lukeed slack'in you (have a few suggestions for this)

lukeed:
> True! That is what we can do for the input fields too right?

For `dest` and `template`, no. We don't want to set _any_ default for those since those are the only two distinct, required fields that the user needs to declare.

For `name`, it's possible, but I don't see a point in doing it. Perhaps `My App` but even then, it's not that useful imo.

reznord:
Makes sense! :)

PR #361, opened 2017-09-07
Adds update-notifier

reznord:
This adds update notifier to the CLI and reminds the user on every CLI command her runs. 

/cc @developit

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
 
+import updateNotifier from 'update-notifier';
 import yargs from 'yargs';
 import create from './commands/create';
 import build from './commands/build';
@@ -9,13 +10,14 @@ import installHooks from './lib/output-hooks';
 import pkg from '../package.json';
 import logo from './lib/logo';
 import checkVersion from './../check';
-import updateNotifier from 'update-notifier';
 
 global.Promise = require('bluebird');
 
 checkVersion();
 
 installHooks();
+
+updateNotifier({pkg}).notify({ defer: true });
 
 yargs
 	.command(create)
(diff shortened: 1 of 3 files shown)


reznord:
Don't know how to test it locally lol 😛 

Hoping that it will work 😅

src/index.js
```diff
@@ -9,6 +9,7 @@ import installHooks from './lib/output-hooks';
 import pkg from '../package.json';
 import logo from './lib/logo';
 import checkVersion from './../check';
+import updateNotifier from 'update-notifier';
```
lukeed: Let's move this to line 2, so that it precedes all the local imports.

src/index.js
```diff
@@ -31,3 +32,5 @@ yargs
 	.demandCommand()
 	.strict()
 	.argv;
+
+updateNotifier({pkg}).notify({ defer: true });
```
lukeed: This should be before all the `yargs(...)` stuff takes place -- aka, right after the `checkVersion()`.

Also, we don't need `defer:true` --- it's `true` by default.

src/index.js
reznord: Done. Didn't think about this earlier.

src/index.js
reznord: Done

src/index.js
```diff
@@ -16,6 +17,8 @@ checkVersion();
 
 installHooks();
 
+updateNotifier({pkg}).notify({ defer: true });
```
lukeed: Can remove `defer:true` because [it is already the default](https://github.com/yeoman/update-notifier#defer). 👍

reznord: Ah! shoot

reznord: Done 🙏

Issue #281, opened 2017-08-03
All routes are chunked, causing hydration of prerendered pages to flicker


ajoslin:
It appears that *all* routes are being chunked and async loaded, even without the presence of the `async!` prefix.

I've reproduced this using preact-cli 1.4.1 here: https://github.com/ajoslin/preact-cli-issue-all-routes-chunked. Just build and serve and check the network tab, it loads both bundle and route-home.chunk. And it does a render before the home chunk is loaded.

**Details**

I have a prerendered homepage. Here's the process when the user visits the site:

1. Index.html loads with prerendered content
2. Main JS bundle loads 
    - Initial render is triggered *without* the home component downloaded
    - Bundle begins downloading route-home.chunk.js
3. Wait for route-home.chunk.js is loaded
4. Home route chunk loads and the home content renders

Between (2) and (4), a version of the app is rendered that *does not* include the home route's content, because the home route's JS is still loading.

This results in a noticeable "flicker" of the home content being removed then re-appearing as soon as the initial bundle loads.

I'm not using the `async!` prefix when importing my home route, but it looks like the home route is being chunked anyway.

prateekbh:
also related #274

ajoslin:
I'm "fix"ing this for now by moving all of my routes into `components/routes`.

thangngoc89:
@ajoslin The async loader was implemented in a way that during prerendering , it act as a normal require so the above behavior is correct (maybe?)

But anyway, we should fix this !

ajoslin:
@thangngoc89 The problem is that it's happening whether or not I use the `async!` prefix.

I've fixed this problem on my end for now by moving my routes into `src/components/pages/`.

thangngoc89:
@ajoslin ah. So it's because of route spliting behavior of preact

lukeed:
This is [supposed to happen](https://github.com/developit/preact-cli/blob/master/src/lib/webpack/webpack-client-config.js#L45-L49).

We also state this in the README:

> Fully automatic code splitting for routes

The `async!` prefix is primarily designed for lazy-loading large components. For example, a REPL component or a WYSIWYG editor.

As for your issue, you're seeing a content flash, correct? If so then we need to fix that part 😄

ajoslin:
Got it. Yes, the problem is just the content flash and that it's rendering before the current page is ready.

lukeed:
[...]

PR #364, opened 2017-09-09
Prevent Route Flickers

lukeed:
This approach reworks the `preact-cli-entrypoint` & `AsyncComponent` so that the initial client-side mount won't wipe/reset the existing static/prerendered markup while waiting for chunks to finish loading.

Currently, the first bundle load kicks off the webpack instance (which initiates the Async loads). That sounds great (and it is!) unless you're on a bad connection and/or are in an emerging market. In those cases, what looked like a functional, ready page, will disappear (to your surprise) and you'll have to wait through another 5+ seconds until the app is back "online"

With this, the first bundle load kicks off the webpack instance (which _still_ initiates the Async loads). The difference here is that it's now "rendering" to a cloned copy of `document.body`, but nothing here will affect the page. Its main purpose is to begin the async chunk-loading.

Once the chunk(s) have been **fully** loaded, a second render will take place, targeting the _real_ `document.body`. In basic cases this will re-create the same markup, so when Preact enters the diff-phase, no changes are made. However, Preact now has full control of the client-side application, allowing libs like `preact-router` to take hold.

As @ajoslin pointed out, this approach _does_ fire two sets of side effects (eg; a network request during `componentWillMount` x2), but anything attached to the fake/clone `document.body` will not yield any _true_ visual effect. Also, this only applies to the very first load, since all other routes (eg, `/` ==> `/about`) are handled on the client-side _only_ without any type of static-vs-jsx negotiation. 

Sorry for the wall of text ~ 🙇 

TL;DR:

* [Original Behavior](http://ajoslin.co/GIH2Ts/5kzN1wTO)
* [PR'd Behavior](https://www.dropbox.com/s/uvk6nvz21emd5fk/async-fix.mp4?dl=0)

---

_Closes #281. Big thanks to @ajoslin for getting started on this._


ajoslin:
I still don't know how I feel about mounting the app twice.. I can imagine some people thinking they can get away with some side effect that should *really* only ever run once, and then it runs twice and they are completely mystified.

Nevertheless, that case is less common and less important than this case. If there's a FAQ somewhere, maybe we should add a "My app's componentDidMount() is running twice! Why?".

And BTW, the flicker effect that this fixes shows up even on fast connections. On fast Wi-Fi, you still see 100-200ms of "no content" flicker.

lukeed:
Yeah, I realize I was lucky enough to rarely see the flicker before your post. When I _did_ see it, I thought it was a font-load issue or something.

The alternative to the "double" mounting is to create a built-in `App` HOC that `dangerouslySetInnerHTML`s the existing `root.innerHtml` on first render only. The trick would be to somehow infer the dynamic portion (eg, `<Router>` w/ async chunks) from the static parts. We would then `dangerouslySetInnerHTML` + `<Router />`. This still initializes the chunk-fetch for the given route, without actually rendering anything _additional_ to the page. The second render would pass a `ready` prop, which mounts the app itself & discards the previous `innerHTML` bit.

Ideally, there should be a Webpack-runtime function we can call pre-mount to know what needs to be loaded.... Maybe it exists? That would make this whole thing a breeze!

```js
function init() {
  // basic init
}

function boot() {
  loadChunks().then(init);
}
```

ajoslin:
Or alternatively, what if we could just tell Preact "when rendering, do not touch the inside of this element."

```js
AsyncComponent.prototype.render = (state) => !state.child ? h(Preact.DoNotModifyInnerHtml) : h(state.child)
````

**Comment by ajoslin** [comment]
Good news, looks like that approach might work:

https://codesandbox.io/s/0o8jjvz7vv

You can `dangerouslySetInnerHTML` **and** set children. The children's componentDidMount event fires, but the end HTML is from dangerouslySet.

This means the app could be rendered, and the HTML set to the prerendered value. 😄

**Comment by lukeed** [comment]
Hey @ajoslin, I had similar at one point.

Your one-off demo works, but it's not actually casting an `AsyncComponent` for which there is a loading period. The reason I mention that is because _our_ `<Home />` has to be rendered/called in order for the `async-loading` event to even fire.

Maybe I'm missing something, but here's how the flow would work with `dangerouslySetInnerHTML` -- notes taken from a local attempt I had going before this PR:

1. Pre-rendered HTML markup

2. With `ready=false` (default) client render uses `dangerouslySetInnerHTML`:
    ```js
    // html ==> root.innerHTML, passed down
    h(props.root, props.ready ? {} : {
      dangerouslySetInnerHTML: { __html:props.html }
    })
    ```

    This is essentially:

    ```js
    <App dangerouslySetInnerHTML={{ __html:props.html }} />
    ```

3. `<App/>` mounts, discarding innards for hardset HTML

4. `<Home/>` mounts, despite being discarded 
4. `AsyncComponent` initiates via `<Home />`

5. Home

**Comment by ajoslin** [comment]
Alright, so I tried out the `dangerouslySetInnerHTML` method as you described above. Unfortunately, it still results in a double mount. 😐 

I think the best solution is not to mess with the entry -- we should go back to trying to patch `AsyncComponent` only. I'm going to try messing with that some more.

**Comment by lukeed** [comment]
Maybe. I still believe the true answer lies within Webpack since it's the master-orchestrator of what loads, when it loads, and when it's done. Preact does it's job perfectly as it is, shouldn't be fighting it.

We need some type of [Manifest Plugin](https://github.com/danethurber/webpack-manifest-plugin) that maps pathnames to `route-*` chunk names.

```js
// ignore *.js.map files here
var mapping = {
  '/': ['route-home'],
  '/about': ['route-about']
}
```

> This would be inferred(?) and generated at compile-time and would be for static pages only -- no dynamic URLs since we don't pre-render those anyway.

Then the `entry.js` picks up a slight modification:

```js
const { h, render } = require('preact');

// ...

function getApp() {
  let x = require('preact-cli-entrypoint');
  return h(x && x.default || x);
}

function init() {
  let app = getApp();
  root = render(app, body, root);
}

const mapping = { ... };
let chunks = mapping[location.pathname] || false;

if (chunks.length > 0) {
  // will be handled by webpack @ runtime
  require.ensure(chunks, init);
} else {
  init();
}

module.hot && module.hot.accept('preact-cli-entrypoint', init);
```

ajoslin:
Yunno, that might be possible with prerender...

- Use webpack to intercept `src/route` calls and wrap in an HOC
- This HOC looks for `path` prop
- During prerender, add all path props to a list and [convert them to regexp](https://github.com/pillarjs/path-to-regexp)
- Add these regexes to an array with a corresponding chunk name: `pathChunks = [{regex, chunk}]`
- On startup, iterate through the array and filter by `pathChunks.filter(regex.test(window.location.pathname))`
- Load all chunks
- Init app

The downside is this would only work if the routing library people use follows `path-to-regexp` conventions and uses `path` props.

lukeed:
> On startup, iterate through the array, filter by matched paths, load chunks, and init:

Webpack already splits each `chunk` into an array of all chunk pieces, so we can skip that step entirely.

The above code snippet I wrote will work. I've been with family all day today, but will start back up with this now. It's just a matter of printing the `route:[chunks]` mapping to the entry // making it accessible to the entry.

Also, it would need to be an exact match for the routes. Partly to enforce static-routes only && partly because `/` will match all routes and `/foo` may be a separate route from `/foo/bar`. Don't need to reinvent a route-matching library just to get a first-boot to work properly.

developit:
Random thoughts:

- The `async!` loader is re-aliased to an identity function during prerendering, but it could instead be used to build up a list of routes.
- Perhaps we could use the upcoming [Router.getMatchingRoutes()](https://github.com/developit/preact-router/pull/222) method on the client here?  Have a wrapper around the Router invoke that method during client bootup, find the current route, replace the component with one that does innerHTML preservation (as you guys figured out already), waits for the route's load method to resolve (specific to async-component, but that seems reasonable here), then swaps to the original defined route?

CedarDCX:
Any progress update on this? I'm also experiencing the same issue where the page flickers on a slow connection.

developit:
I have a possible solution I can PR.

lukeed:
@developit Did you have that sumthin' sumthin'?

lukeed:
I got it 🍾 I think I'll open a new PR instead of updating this one~

ajoslin:
@lukeed my $100 bounty is still open on this too!

PR #370, opened 2017-09-17
Fix issue 369. Eslint warnings and errors on full yarn sample.

lambrospetrou:


reznord:
we are planning to move to a new preset instead of using the [current one](https://github.com/synacor/eslint-config-synacor).

Synacor's eslint is a bit too strict and some modifications needs to that config.

Thanks for working on the PR but since we are going to deprecate the examples, I'm closing this.

Issue #373, opened 2017-09-17
Custom `https` certs during `watch`


lukeed:
The `webpack-dev-server` is able to [accept an object](https://webpack.js.org/configuration/dev-server/#devserver-https) for custom HTTPS configuration. 

Right now, we only [display an `https://` origin](https://github.com/developit/preact-cli/blob/master/src/lib/webpack/webpack-client-config.js#L114) if we've received `true`. 

However, we still forward the `config.https` object to `webpack-dev-server`... which creates an inconsistency.

PR #375, opened 2017-09-18
Migrate to Webpack 3.x

### LINKED ISSUE #367: Upgrade to Webpack 3
And remove/replace `webpack-blocks` in the process? 
  _We _currently_ have to since it only supports `1.x` and `2.x`_

lukeed:
- Drops the dependencies on `@webpack-blocks/*` which was explicitly requiring `webpack@2.x` and didn't look like an update was on the horizon.

- Includes `webpack.ModuleConcatenationPlugin` on client builds.

- Installs style-type loaders that `@webpack-blocks` installed for us previously

- Merges `webpack` configuration values via `webpack-merge`

Thanks to @hassanbazzi who did most of the config rewrite in #150. This PR is largely based off his work 👏 This exists as a separate PR because some of Hassan's original changes were lost over time during merge-commits & this was easier (imo) than trying to go back and fix them.

---

Closes #367, #373, #150 _(outdated)_


lukeed:
@reznord Lol, it's not possible 😜

src/lib/webpack/prerender.js
```diff
@@ -1,17 +1,17 @@
...
 import { SourceMapConsumer } from 'source-map';
-import chalk from 'chalk';
 
 export default function prerender(env, params) {
 	params = params || {};
 
-	let entry = resolve(env.dest, './ssr-build/ssr-bundle.js'),
-		url = params.url || '/';
+	let entry = resolve(env.dest, './ssr-build/ssr-bundle.js');
+	let url = params.url || '/';
```
hassanbazzi: is `let` neccesary here? Can this be a `const`?

lukeed: @developit prefers `let` for most things. I broke the rule & used `const` a few times to be emphatic.

src/lib/webpack/render-html-plugin.js
```diff
@@ -0,0 +1,44 @@
...
+		preload: config.preload===true,
+		title: title || config.title || config.manifest.name || config.manifest.short_name || (config.pkg.name || '').replace(/^@[a-z]\//, '') || 'Preact App',
+		excludeAssets: [/(bundle|polyfills)(\..*)?\.js$/],
+		config,
+		ssr(params) {
+			return config.prerender ? prerender({ cwd, dest, src }, { ...params, url }) : '';
+		}
+	});
+
+	const pages = readJson(resolve(cwd, config.prerenderUrls || '')) || [{ url: '/' }];
```
hassanbazzi: Should't this be using `params.url` for the fallback?

src/lib/webpack/webpack-client-config.js
```diff
@@ -1,132 +1,167 @@
...
-		setOutput({
+	return {
+		entry: {
+			bundle: resolve(__dirname, './../entry'),
+			polyfills: resolve(__dirname, './polyfills')
+		},
+
+		output: {
 			path: env.dest,
 			publicPath: '/',
```
hassanbazzi: Can we use params.url here?

src/lib/webpack/render-html-plugin.js
lukeed: For `pages`? The `params` is only within the `ssr` function. I think the `/` route is a safe default. Plus, it's just copied over from what was there already.

hassanbazzi: Yea maybe we should do it in a separate PR. Basically, the whole "subdirectory" stuff is something we should fix. But yea, let's keep it separate.

hassanbazzi:
I won't be able to test until later tonight if that's ok :)

lukeed:
All good~!

reznord:
@lukeed every thing works pixel perfect 😄

reznord:
Merging this.

PR #377, opened 2017-09-19
adding flag to disable service worker

### LINKED ISSUE #371: How to unregister / remove any registered service workers?
Is it possible to remove any registered service workers by this CLI for any visitors on a production app? The application runs currently with the precache sw.js and I would like to have it removed. I understand that I can do it myself with the console tools, but I can't let my users do the same.

prateekbh:
fixes #371


src/lib/webpack/webpack-client-config.js
```diff
@@ -199,7 +199,12 @@ const production = config => addPlugins([
 			/push-manifest\.json$/,
 			/.DS_Store/
 		]
-	})
+	}),
+	new webpack.DefinePlugin({
```
developit: I had posted in chat about this but I'll echo here for visibility - does the object form here replace as optimally as this?

```js
new webpack.DefinePlugin({
  'process.env.ADD_SW': config.serviceWorker
});
```

lukeed: I believe `yargs` will have camelCased the flag name at this point.

@prateekbh Have you checked/ran this? This was my thought as well.

prateekbh:
@developit  @lukeed it worked for me bothways with and without camel casing

prateekbh:
btw made suggested changes, is it ready to merge yet?

PR #379, opened 2017-09-21
Clean-up Dependencies

lukeed:
No reason to include any `less` dependency, since those packages will be specified by a template.

Then `mkdirp` and `offline-plugin` were removed because they weren't used anywhere ion the codebase.

---

_Closes #372_


hassanbazzi:
👍

PR #383, opened 2017-09-25
Update Readme with custom-templates listing

reznord:
Updates the `README` reflecting changes to `custom templates` which is the highlight for 2.0.

Any inputs are more than welcome.

Please *do not* merge this PR until we are planning for a release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,15 +2,7 @@
 
 > Start building a [Preact] Progressive Web App in seconds 🔥
 
----
-
-This is the documentation for our **development** version. For the documentation of the latest **stable** release, please visit the latest tagged. For example:
-
-[**1.4.1**: https://github.com/developit/preact-cli/tree/1.4.1](https://github.com/developit/preact-cli/tree/1.4.1)
-
----
-
-### Features:
+### Features
 
 - **100/100 Lighthouse score**, right out of the box ([proof])
 - Fully **automatic code splitting** for routes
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -24,131 +16,119 @@ This is the documentation for our **development** version. For the documentation
...
-npm start
+Current available templates include:
 
-# go to production:
-npm run build
-```
+- [default] - Default template with all features.
 
-### Using Yarn
+- [material] - material-ui template using preact-material-components
```
prateekbh: "material template...." `material-ui` might be confusing with react's material-ui

README.md
```diff
@@ -24,132 +16,126 @@ This is the documentation for our **development** version. For the documentation
...
+  --port, -p    Port to start a server on                        [string]   [default: "8080"]
+  --host,       Hostname to start a server on                    [string]   [default: "0.0.0.0"]
+  --https       Use HTTPS?                                       [boolean]  [default: false]
+  --prerender   Pre-render static app content on initial build   [boolean]  [default: false]
 ```
 
-### Templates
+Note:
+
+1. You can run dev server using`HTTPS` then you can use the following `HTTPS=true preact watch`
```
lukeed: Missing space before `HTTPS`

README.md
```diff
@@ -24,132 +16,126 @@ This is the documentation for our **development** version. For the documentation
...
-              (used for --server config)                          [default: -]
-  --port, -p  Port to start a server on.                          [default: "8080"]
+```sh
+$ preact watch
 
+  --cwd         A directory to use instead of $PWD.              [string]   [default: .]
+  --src         Entry file (index.js)                            [string]   [default: "src"]
+  --port, -p    Port to start a server on                        [string]   [default: "8080"]
+  --host,       Hostname to start a server on                    [string]   [default: "0.0.0.0"]
+  --https       Use HTTPS?                                       [boolean]  [default: false]
```
lukeed: Don't need `?`

README.md
```diff
@@ -24,132 +16,126 @@ This is the documentation for our **development** version. For the documentation
 - In just **4.5kb** you get a productive environment:
   - [preact]
   - [preact-router]
-  - 1.5kb of conditionally-loaded polyfills for [fetch](https://github.com/developit/unfetch) & [Promise](https://npm.im/promise-polyfill)
+  - 1.5kb of conditionally-loaded polyfills for [fetch] & [Promise]
 
+### Installation
 
-### Commands
+Prerequisites: [Node.js](https://nodejs.org/en/) required version (> 6.x)
```
lukeed: How about `> **Important:** Node v6.x is the minimum requirement.`

silverlight513:
Made a few tweaks and put it into gist:
https://gist.github.com/silverlight513/35a3c85df18eef73963bc61be146a9b7

I've bullet pointed my suggested updates into a review

lukeed:
@silverlight513 Thanks! 🙏  

I think this is a good example to point out that we need to clarify that the "template" in `Official Templates` is a different thing than the `--template` flag.

Nesting them under each other will surely, imo, add to the confusion.

silverlight513:
Ah, from the looks of it they appeared to be the same thing. My bad.

Would you mind explaining the difference between the `--template` flag and the `Official Templates` so that I can update my gist to try and make it clearer that they're separate.

reznord:
@lukeed @silverlight513 thanks for the reviews and your time 🙏

**Official Templates**

Till now (the current version) we were just using the code in the [examples folder] for boot-strapping new projects using preact cli.

But in the next version (v2.0) we moved to custom-templates which will pull the the project from the `github/bitbucket`. There is a specific structure which these custom-templates need to follow, i.e. they need to place the entire code in a `template/` folder.

The contents in the `template/` folder from the repo will be used as a template for the project.

---

**Template flag**

The template here refers to a custom `HTML` file which the user can specify for either `preact build`, `preact watch` commands. 

By default in preact-cli we don't expose the HTML file as [CRA] does. So, we support a `--template` flag through which we can pass a custom `HTML file` where you can add external stylesheets/scripts.

You can use them in the following way:

- Build command: `preact build --template path/to/template.html`
- Serve command: `preact serve --template path/to/template.html`


[examples folder]: https://github.com/developit/preact-cli/tree/master/examples
[CRA]: https://github.com/facebookincubator/create-react-app

lukeed:
@silverlight513 No problem, I'm sure you're not/won't be the only one.

The `--template` flag is for passing a path to a custom `index.html` replacement. This is useful for adding `<meta>` tags to your head, etc. TBH, this may be removed soon in favor of _always_ writing the [built-in `template.html`](https://github.com/developit/preact-cli/blob/master/src/resources/template.html) into your working directory.

The "Official Templates" (or any template) is a new 2.0 feature which allows you to scaffold a new project from an external repository. Here, "template" is synonymous with "boilerplate." The motivation here is to allow community-driven, reusable templates so that users don't feel restricted to what's included in Preact CLI directly.

lukeed:
We'll definitely update this as we go along. Better to have something reflective of the current repo sooner rather than later.

PR #384, opened 2017-09-25
feature: preact list

reznord:
New command to the CLI: `preact list`

Now that we have moved to custom templates, it is better to add a CLI command which will list out the official repos.

**Edit**: It fetches the list of repos from the URL `https://api.github.com/users/preactjs-templates/repos`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,7 +95,6 @@
   },
   "dependencies": {
     "autoprefixer": "^7.1.0",
-    "axios": "^0.16.2",
     "babel-loader": "^7.0.0",
     "babel-plugin-jsx-pragmatic": "^1.0.2",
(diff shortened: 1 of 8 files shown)


reznord:
Added to README in #383 (commit ref: https://github.com/developit/preact-cli/pull/383/commits/0c6df1b452a418d4e033363e75e2e1c74813697f)

reznord:
> Please use unfetch or isomorphic-unfetch instead of axios

I didn't use `unfetch` since we are not running this in browser. We will be running it on node.

Re `isomorphic-unfetch`, I tried that but it gives a weird output instead of the body.

/cc @lukeed

developit:
heh - streams!

PR #394, opened 2017-10-11
Update dependencies

reznord:
Update `yarn.lock` file


lukeed:
Only need to update `yarn.lock` file when updating a minor or major dependency version.

PR #399, opened 2017-10-12
Update material template link, version bump

reznord:


lukeed:
Maybe it's just me, but I find it odd to PR the version changes. That's what the `npm version (major|minor|patch) && npm publish` is for IMO

Issue #403, opened 2017-10-16
options.sourceMap isn't set - CLI 2.0.0


timarney:
Trying to track down why this error might be happening in dev.  CLI version 1.4.1 was fine.

Seeing the following when I try to start the dev server.

 @ ./node_modules/preact-cli/lib/lib/entry.js
⚠ WARN ./node_modules/preact-cli/lib/lib/webpack/proxy-loader.js?{"__proxy_loader__":{"loader":"sass-loader","cwd":"my-dir-here"},"sourceMap":true}!./src/components/menu/style.scss
(Emitted value instead of an instance of Error)

Previous source map found, but options.sourceMap isn't set.
In this case the loader will discard the source map entirely for performance reasons.
See https://github.com/postcss/postcss-loader#sourcemap for more information.

**Details**
npm 5.5.1
node 8.7.0

euangoddard:
I have exactly the same issue following the same upgrade path.

npm 5.5.1
node 8.1.3

lukeed:
This was a change within `postcss-loader` that I overlooked. PR open, just waiting for another team member to approve & then we'll publish a patch release.

Thanks guys.

lukeed:
Now available in `2.0.1`

timarney:
Hi @lukeed 

Based on your PR I think my issue is with the proxy-loader.  If I update as you did to use isProd vs true that seems to work.  Not sure is there is a reason for that to be forced to true.

```
use: [{
					loader: 'proxy-loader',
					options: {
						cwd,
						loader: 'sass-loader',
						options: { sourceMap: isProd }
					}
				}]
```

lukeed:
Hey, `proxy-loader` just passes the `options` along to the loader its proxying. They're all set to `true` to always enable it, initially. 

In 1.4, the `isProd` value was still the [final determinant](https://github.com/developit/preact-cli/blob/1.4.1/src/lib/webpack/webpack-base-config.js#L185-L199). 

The `proxy-loader` settings haven't changed from 1.4 to 2.0 --- only `postcss-loader` has & the last PR patches that difference.

Thanks!

euangoddard:
@lukeed I've just updated to 2.0.1 and am still getting errors:
[...]

PR #407, opened 2017-10-17
Allow PostCSS to `sourceMap` in Dev mode

lukeed:
Previous PR (#404) enabled `sourceMap`s for production builds _only_.  Derp. This enables mapping during development.

_Closes #403 ... again_

### DIFF
--- a/src/lib/webpack/webpack-base-config.js
+++ b/src/lib/webpack/webpack-base-config.js
@@ -142,7 +142,7 @@ export default function (env) {
 							{
 								loader: 'postcss-loader',
 								options: {
-									sourceMap: isProd,
+									sourceMap: true,
 									plugins: [autoprefixer({ browsers })]
 								}
 							}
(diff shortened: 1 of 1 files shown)


src/lib/webpack/webpack-base-config.js
```diff
@@ -142,7 +142,7 @@ export default function (env) {
 							{
 								loader: 'postcss-loader',
 								options: {
-									sourceMap: isProd,
+									sourceMap: true,
```
reznord: can't we leave this in the user land whether to use sourceMaps in production or not?

lukeed: This isn't production. This just always _allows_ PostCSS to accept source maps. Whether or not source maps are created is determined by the `(css|sass|stylus|less)-loader`s. 

This can still be overridden in Userland, if wanted, but there'd be no point.

Development sourcemaps don't work with out this.

PR #414, opened 2017-10-20
Fix Tests

lukeed:
... hopefully 😒

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "preact-cli",
-  "version": "2.0.0",
+  "version": "2.0.1",
   "description": "Start building a Preact Progressive Web App in seconds.",
   "repository": "developit/preact-cli",
   "main": "lib/index.js",
--- a/package.json
+++ b/package.json
@@ -90,6 +90,7 @@
     "eslint-plugin-react": "^7.0.1",
     "html-looks-like": "^1.0.2",
     "jest": "^20.0.4",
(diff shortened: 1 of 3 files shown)


lukeed:
(ノಠ益ಠ)ノ彡┻━┻

lukeed:
:tada: **BIG THANKS** :tada:  to @fokusferit and @reznord for tracking down the underlying issue.

As @fokusferit described it:

> The tests are currently failing because of [this line](https://github.com/developit/preact-cli/blob/master/src/commands/create.js#L162).
> The .replace Regular Exp is only checking for any whitespace character but within the tests when we are using it, the .name of argv is set to the path , e.g. ‘/Users/ferittopcu/Desktop/projekte/preact-cli/tests/output/test3’ -> in the package.json of our temp /test/output folder we had an invalid name and therefore the npm install command was throwing an error but the spawn has no error handling therefore it got lost.

Basically, the auto-generated test directory names were invalid `package.json` names, and that if-block was ensuring that the `.name` key _always_ had a value, even if it was wrong.

Luckily this is taken care of elsewhere in the CLI with relative pathing, but the tests are passing around absolute paths.

To remedy, we just had to ensure a _valid_ `.name` fallback for tests, instead of trusting the incoming (absolute) destination path.

reznord:
That moment when we see a green tick from Travis 😍♥️

fokusferit:
Awesome 👏🏼 !

PR #417, opened 2017-10-21
hmr fix

zouhir:
I have a little preact-cli project I work with on weekends.

Today HMR stopped working and I am able to get it to work back via this fix.

dev-server was not emitting socket-js with the dev bundle and seems like the combo of Webpack HMR plugin + `hot:true` in dev-server options is not working.

This fix will ensure in Dev mode, the required socket stuff are committed to `hmr` bundle and the reload happens.

is there better ideas? happy to discuss.

### DIFF
--- a/src/lib/webpack/webpack-client-config.js
+++ b/src/lib/webpack/webpack-client-config.js
@@ -10,14 +10,19 @@ import PushManifestPlugin from './push-manifest';
 import baseConfig from './webpack-base-config';
 
 function clientConfig(env) {
-	const { isProd, source, src } = env;
+	const { isProd, source, src /*, port? */ } = env;
+
+	let entry = {
+		bundle: resolve(__dirname, './../entry'),
+		polyfills: resolve(__dirname, './polyfills')
+	};
+
+	if (!isProd) {
+		entry['hmr'] = `webpack-dev-server/client?http://localhost:${process.env.PORT || env.port || 8080}`;
+	}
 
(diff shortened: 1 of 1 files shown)


src/lib/webpack/webpack-client-config.js
```diff
@@ -10,14 +10,19 @@ import PushManifestPlugin from './push-manifest';
...
-			bundle: resolve(__dirname, './../entry'),
-			polyfills: resolve(__dirname, './polyfills')
-		},
+	let entry = {
+		bundle: resolve(__dirname, './../entry'),
+		polyfills: resolve(__dirname, './polyfills')
+	};
 
+	if (!isProd) {
+		entry['hmr'] = `webpack-dev-server/client?http://localhost:${process.env.PORT || env.port || 8080}`;
```
reznord: This is a temporary fix with with we can go for now. But I guess the HMR failure is related to webpack-dev-server or html-webpack-plugin.

Once it is fixed in the upstream, we can clean this off.

zouhir: It's webpack-dev-server.

zouhir:
@developit the port in here has to exactly be your devServer port option otherwise you'll get console messages that socket can't connect, therefore no hmr.

marvinhagemeister:
Thank you so much for fixing this! We definitely tort this upstream in case someone hasn't already.

PR #429, opened 2017-11-12
Fixes #374

### LINKED ISSUE #374: Firebase config overwrites unrelated keys
*bug*

Output production on `hosting` key options

With an existing `firebase.json` file:
`npm run serve -- --server config --dest .`

The file is merged only overriding the `hosting` key. A deep merge of the `hosting` key might be even better, but I don't know that it's really needed.

Merged `firebase.json`:
```
{
  "database": {
    "rules": "database.rules.json"
  },
  "hosting": {
    "public": "build",
    "rewrites": [
      {
        "source": "**",
        "destination": "/index.html"
      }
    ],
    "headers": [
      {
        "source": "**",
        "headers": [
          {
            "key": "Cache-Control",
            "value": "public, max-age=3600, no-cache"
          },
          {
            "key": "Access-Control-Allow-Origin",
            "value": "*"
          },
          {
            "key": "Access-Control-Max-Age",
            "value": "600"
          }
        ]
      },
      {
        "source": "/sw.js",
        "headers": [
          {
            "key": "Cache-Control",
            "value": "private, no-cache"
          }
        ]
      },
      {
        "source": "**/*.chunk.*.js",
        "headers": [
          {
            "key": "Cache-Control",
            "value": "public, max-age=31536000"
          }
        ]
      },
      {
        "source": "/",
        "headers": [
          {
            "key": "Link",
            "value":
              "</bundle.6acb8.js>; rel=preload; as=script, </style.ba975.css>; rel=preload; as=style, </route-home.chunk.8111b.js>; rel=preload;as=script"
          }
        ]
      }
    ]
  }
}

```

**If this is a feature request, what is motivation or use case for changing the behavior?**
When using CI/CD pipelines like Gitlab pipelines this config is going to change when the pipeline runs and should therefore be an automatable. In my case I also have a top level `database` key that I don't want overridden. 

- node version
`8.4.0`
- npm version
`5.3.0`
- Operating system
mac sierra 10.12.6

ralphsmith80:
merged preact-cli created config into firebase `hosting` key
to preserve other config options

### DIFF
--- a/src/commands/serve.js
+++ b/src/commands/serve.js
@@ -222,10 +222,14 @@ const SERVERS = {
 			}
 		}
 
-		let config = JSON.stringify({
-			...configObj,
-			public: path.relative(dir, configObj.public)
-		}, null, 2);
+		let config = await readJson(path.resolve(dir, outfile))
+		config = Object.assign({}, config, {
+			hosting: {
+				...configObj,
+				public: path.relative(dir, configObj.public),
+			}
+		})
+		config = JSON.stringify(config, null, 2);
 
 		if (outfile) {
 			await fs.writeFile(path.resolve(dir, outfile), config);
--- a/src/commands/serve.js
+++ b/src/commands/serve.js
@@ -222,13 +222,13 @@ const SERVERS = {
 			}
 		}
 
-		let config = await readJson(path.resolve(dir, outfile))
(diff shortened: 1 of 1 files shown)


developit:
It looks like this should still work with `simplehttp2server`.

Issue #428, opened 2017-11-12
Write `template.html` to Project


lukeed:
**Do you want to request a *feature* or report a *bug*?**

Feature

**If this is a feature request, what is motivation or use case for changing the behavior?**

In order to make any customizations to the `lib/template.html`, one has to manually copy & paste the contents into the current directory.

I have two suggestions, an either/or scenario:

1) Write the `lib/template.html` to the project by default during `create`
2) Add some kind of command or flag to print the current template to project.
    _This could go nicely with #427 + its duplicate requests... write all internal configs to repo?_

reznord:
This is totally a much needed thing for the next release. I spoke to @developit on this and he said he will look into it when he gets some time.

prateekbh:
>1. Write the lib/template.html to the project by default during create

I think this option is pretty straight forward and super easy to understand and we should do this by default. Thoughts?

FYI I have free cycles now. feel free to assign me stuff

developit:
I have some time too that I can commit to this. Reason I've held off pushing for this is because I really want the `template.html` to be simpler than it currently is, and not specific to `html-webpack-plugin` (since that would be a very leaky abstraction).  Ideally, we should look for an `index.html` in the user's project since that's the built name anyway (nice to keep things 1:1).  That file would just be a simple HTML document, and preact-cli would manage parsing it into a DOM and injecting the scripts/styles/prerendered content/etc into it.

prateekbh:
Can we split the work here and first just copy the existing one to user land, and then keep simplifying it as we go?

prateekbh:
Closing this as #705 fixes it. Lets open a new issue for improving the template itself

PR #430, opened 2017-11-12
Allow any/all Prerender Values

lukeed:
Right now, we only allow for `url` and `title` values to be defined inside a prerender config file:

```js
// prerender-urls.json
[{
  "url": "/",
  "title": "Homepage"
}, {
  "url": "/route/random"
}]
```

This change allows the developer to pass **anything they want** inside the file. For example, descriptions, keywords, Google Analytics tracking IDs, etc.

It's meant to work alongside a [custom template file](https://github.com/developit/preact-cli#template) so that these values can be looked up & printed at `build` time.

```js
// prerender-urls.json
[{
  "url": "/",
  "title": "Home",
  "description": "Welcome to my website!"
}, {
  "url": "/about",
  "title": "About",
  "keywords": "mission statement, about us, foobar",
  "description": "Read about our mission!!"
}]
```

Merging this brings us 1 step closer to serving as a static-site generator, of sorts. 🎉

### DIFF
--- a/src/lib/webpack/render-html-plugin.js
+++ b/src/lib/webpack/render-html-plugin.js
@@ -9,28 +9,31 @@ import prerender from './prerender';
 export default function (config) {
 	const { cwd, dest, isProd, src } = config;
 
-	const htmlWebpackConfig = ({ url, title }) => ({
-		filename: resolve(dest, url.substring(1), 'index.html'),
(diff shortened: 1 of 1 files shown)


lukeed:
Oh, this was also the motivation for opening #428. Having the `template.html` readily-available, one way or another, makes this process seamless.

PR #440, opened 2017-12-11
fix: #431 duplication warning

### LINKED ISSUE #431: Remove `DeprecationWarning` during build

Bug

While running `preact build`, you will always get this deprecation warning:

```
DeprecationWarning: Chunk.modules is deprecated. Use Chunk.getNumberOfModules/mapModules/forEachModule/containsModule instead.
```

```sh
$ preact create default my-foobar --no-install
$ cd my-foobar && yarn
$ yarn build
```

There shouldn't be any deprecation message. Eventually, this will move from a soft-warning to a hard error.

---

Pretty certain that [this line](https://github.com/developit/preact-cli/blob/master/src/lib/webpack/run-webpack.js#L139) is the culprit. 👓

matthewlynch:
Look like you were right @lukeed!

All tests passed locally.

---

Closes #431

### DIFF
--- a/src/lib/webpack/run-webpack.js
+++ b/src/lib/webpack/run-webpack.js
@@ -136,7 +136,7 @@ export function writeJsonStats(stats) {
 	function strip(stats) {
 		stats.modules.forEach(stripLoaderFromModuleNames);
 		stats.chunks.forEach(c => {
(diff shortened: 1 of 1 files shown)


lukeed:
Will let this run out on Travis now that tests are working again. We need to see some green for morale 😆 

Thanks!!

PR #445, opened 2017-12-21
[WIP] Rework CLI; Simplify tests

lukeed:
We've known for a while that our tests are/can be problematic. For a while (it was my fault 😞), they were completely broken & that stalled development pretty significantly.

This PR aims to: 
  1. Make tests faster & simpler by 
    a. removing as many spawned processes as possible
    b. removing (all) `npm install`s 
    c. having fewer moving parts
  2. Make the CLI itself faster
    _I was waiting upwards of 2s per `preact *` command_

## Results

1. The `preact build` tests complete in less than `30s` on Node v6, down from `6m`!
2. All commands export their handlers directly, which means tests can use them directly.
  _AKA: easier debugging & no more silent test errors!_ :tada:
3. Average `preact *` command interaction is `0.5s` on my machine, down from `1.9s`

## Changes

- Only display flags that the developer can use to _change_ the output.

   IMO, it's pointless to show `--prerender  (default true)` & then ***not*** show them that `--no-prerender` is how to get rid of it. We've had questions about this a few times.

- Currently, there's no more "interactive mode" if dev does not provide all required arguments.

    > Sade will exit with an "Insufficient arguments" error, but I may add a different approach for optional (`[repo]` vs `<repo>`) arguments.

- Currently no Preact logo in the CLI header. TBH, this never rendered correctly for me, but I can get it back in.

## TODO

- [ ] Add `--no-*` flags to the CLI output
- [ ] Re-include the Preact logo/header
- [x] Finish "Insufficient args" error
- [ ] Final tests pruning / pass

---

## Update

* Required Node 8.x as minimum (native async)
* _ALL_ tests complete in less than 30 seconds 🎉
* Entire TravisCI suite [takes 2 minutes](https://travis-ci.org/developit/preact-cli/jobs/348976802)
* Removed serve command 👋 (will be its own package)
* Pass custom HTTPS certs into watch via webpack-style CLI flags
    ```
    $ preact watch --https --key <foo> --cert <bar>
    ```


src/lib/webpack/transform-config.js
```diff
@@ -1,28 +1,29 @@
...
 	} catch (e) {
-		if (env.config) {
-			throw new Error(`preact-cli config could not be loaded!\nFile ${env.config} not found.`);
-		}
-		return;
+		if (env.config === FILE) return;
+		throw new Error(`preact-cli config could not be loaded!\nFile ${env.config} not found.`);
 	}
 
 	require('babel-register')({
```
developit: I think we could probably really cut down further on init time if we found a way to avoid pulling in babel-register here.  Maybe `buble-register` as a faster stand-in?

developit:
now that's an awesome perf boost

src/lib/webpack/transform-config.js
lukeed: Yeah, Buble would certainly be faster, but we're only pulling in Babel in the first place because of 1) async and 2) imports.

I'd advocate for Nodent instead for `async` (I found out how to avoid all the overhead).  And we can use Reify for the imports.

All of this drops any Babel runtime code and I think the new stats would be even better.

package.json
```diff
@@ -93,6 +93,8 @@
     "jest": "^20.0.4",
     "mkdirp": "^0.5.1",
     "ncp": "^2.0.0",
+    "node-sass": "^4.7.2",
+    "sass-loader": "^6.0.6",
```
ForsakenHarmony: Is this needed as a top level dev dep?

lukeed: DevDep, yes. That way we can skip installation _during_ the tests. Moving it here allows it to be cached, doesn't reinstall all other devdeps, and was a big player in speeding up the `build` test from minutes to seconds.

ForsakenHarmony: Alright, makes sense ig

lukeed:
Still a WIP 😄 

At this point, all files are in CommonJS format & object spreads (`{ ... }`) have been replaced with `Object.assign` instead. All `async` statements are untouched.

These are (probably?) just temporary, but it's really helping me debug the tests, since there's no longer a load of compiled crap to weed through.

> But, worth nothing, ***ALL*** tests (except `serve` and `watch`, still broken) complete within ***9 seconds***, down from 30s for _just_ `build`!

Pushed these changes only so that @reznord could take a look. 👀

reznord:
@lukeed sorry, had to clean up some old issues (before this). Hence we ended up in some conflicts 🙈 (will fix them)

ForsakenHarmony:
We had conflicts before afaik

developit:
@lukeed - thoughts on using [esm](https://github.com/standard-things/esm) to make `import` work? Would simplify the diff a fair bit.

lukeed:
Sure thing. Didn't know of that!

The CJS exports just a temporary adjustment so that I could see more clearly into the test errors. I was planning on writing them back to ESM format since I didn't run CJS by anybody...not this PR at least

reznord:
Linter was there for a while in CLI (remember your commit?) **Linters are dumb** :P

lukeed:
I conveniently forgot

lukeed:
*walks away*

lukeed:
### Update

* _ALL_ tests complete in less than 30 seconds 🎉 
* Entire TravisCI suite [takes 2 minutes](https://travis-ci.org/developit/preact-cli/jobs/348976802)
* Removed `serve` command 👋 (will be its own package)
* Pass custom HTTPS certs into `watch` via webpack-style CLI flags
    ```
    $ preact watch --https --key <foo> --cert <bar>
    ```

lukeed:
Closing this in favor of the `next` branch, which is now available on npm under the `next` tag 🎉 

```
npm install --save-dev preact-cli@next
```

There's a preliminary summary of this PR on the [pre-release notes](https://github.com/developit/preact-cli/releases/tag/v3.0.0-next.1).

PR #454, opened 2018-01-06
Fix webpack config not matching win32 paths

### LINKED ISSUE #453: Some paths in webpack configs aren't matched on Windows
Bug

I discovered this bug because routes are named `*.chunk.*.js` instead of `route-*.js` in the build output. This happens because Windows' backslash-delimited file paths (e.g. `C:\Users\me\dev\project\src\routes\home\index.js` aren't matched by these strings and regexes:

https://github.com/developit/preact-cli/blob/9eb46228922b05ecf88c0803a64f8fa5f04b2ab7/src/lib/webpack/webpack-client-config.js#L55-L58

There are probably more occurences of this throughout the webpack configs.

1. `preact create default test` on Windows
2. `cd test`, `npm install`
3. `npm run build`
4. In the `build` folder, route chunks aren't prefixed with `route-`

In this specific case, routes should be prefixed with `route-`, like on UNIX systems.
Generally, both `/` and `\`-delimited paths should be matched in webpack configs.

- node version 8.9.1
- npm version 5.6.0
- Operating system Windows 10 1709

jgierer12:
Fixes #453 

AFAIK, this issue only occurred in this place, other webpack configs don't use unescaped paths. Nevertheless, I put the `normalizePath` function into `util.js` in case it has to be reused elsewhere in the future.

### DIFF
--- a/src/lib/webpack/webpack-client-config.js
+++ b/src/lib/webpack/webpack-client-config.js
@@ -56,8 +56,8 @@ function clientConfig(env) {
 					options: {
 						name(filename) {
 							filename = normalizePath(filename);
-							if (!filename.includes('/routes/')) return false;
 							let relative = filename.replace(normalizePath(src), '');
+							if (!relative.includes('/routes/')) return false;
 							return 'route-' + cleanFilename(relative);
 						},
 						formatName(filename) {
--- a/src/lib/webpack/webpack-client-config.js
+++ b/src/lib/webpack/webpack-client-config.js
@@ -8,6 +8,9 @@ import SWPrecacheWebpackPlugin from 'sw-precache-webpack-plugin';
 import RenderHTMLPlugin from './render-html-plugin';
 import PushManifestPlugin from './push-manifest';
 import baseConfig from './webpack-base-config';
+import { normalizePath } from '../../util';
+
+const cleanFilename = name => name.replace(/(^\/(routes|components\/(routes|async))\/|(\/index)?\.js$)/g, '');
 
 function clientConfig(env) {
 	const { isProd, source, src /*, port? */ } = env;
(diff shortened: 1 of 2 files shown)


src/lib/webpack/webpack-client-config.js
```diff
@@ -52,15 +55,15 @@ function clientConfig(env) {
...
+							if (!filename.includes('/routes/')) return false;
+							let relative = filename.replace(normalizePath(src), '');
+							return 'route-' + cleanFilename(relative);
 						},
 						formatName(filename) {
-							let relative = filename.replace(source('.'), '');
-							// strip out context dir & any file/ext suffix
-							return relative.replace(/(^\/(routes|components\/(routes|async))\/|(\/index)?\.js$)/g, '');
+							filename = normalizePath(filename);
+							let relative = filename.replace(normalizePath(source('.')), '');
```
jgierer12: Any specific reason to use `source('.')` here instead of `src` like above?

thangngoc89:
Hey @developit 
Could you take a look at this because you're the only person I know to code on Windows

ForsakenHarmony:
What exactly is this a fix for?

src/lib/webpack/webpack-client-config.js
```diff
@@ -52,15 +55,15 @@ function clientConfig(env) {
 					loader: resolve(__dirname, './async-component-loader'),
 					options: {
 						name(filename) {
-							let relative = filename.replace(src, '');
-							let isRoute = filename.indexOf('/routes/') >= 0;
-
-							return isRoute ? 'route-' + relative.replace(/(^\/(routes|components\/(routes|async))\/|(\/index)?\.js$)/g, '') : false;
+							filename = normalizePath(filename);
+							if (!filename.includes('/routes/')) return false;
+							let relative = filename.replace(normalizePath(src), '');
```
developit: We might want to run this check on `relative` instead of `filename` - currently this will trigger a false positive if the repo is within a directory called "routes" (no clue why it would be, but still)

jgierer12: ✅ https://github.com/developit/preact-cli/pull/454/commits/a6608c18d4351b6a9124ed45b4d060467eb4865d

src/lib/webpack/webpack-client-config.js
ForsakenHarmony: maybe noone ever changed it, not sure

PR #460, opened 2018-01-12
[FIX] Non-Unique Filename Hashing

### LINKED ISSUE #459: Same bundle hash for modified content across two builds
Bug

I get stable hashes for my bundle.js across multiple builds with modified files. 

Create a new project via preact-cli (default template) build it, add something in app.js and rebuild it. 
Both artifact versions (bundle.js) get the same hash which breaks distributing via CDN. 

The second run should generate a different hash. I don’t tried the same thing with the style.css bundle. Maybe it’s the same thing.

- Node version 9.4
- npm version bundled version
- Operating system OSX
- CLI version latest 
- Browser all

P.S.: besides this I’m in love with this cli. Everything works OOTB or can be added via plugin or a small configuration. Great work!

lukeed:

Fixes #459 

N/A

**Summary**

Apparently, since October 2017, an underlying dependency has had an incompatibility with Webpack's `ModuleConcatenationPlugin`. Because of this, **all files and chunks** were producing the same `[hash]` and `[chunkhash]` values.

> Before removing, I tried adjusting the `output` patterns, but that didn't change anything.

Luckily, it seems as though we weren't actually using this plugin in the way it was intended, so simply removing it worked. 

We're now relying on Webpack's own hashing algorithm, which is probably a safe(r) bet since that's one of its first-class priorities.

---

@developit This will have to go out ASAP since everyone is producing repetitive builds. I suspect this is also playing a role in the SW issues.

### DIFF
--- a/package.json
+++ b/package.json
@@ -157,7 +157,6 @@
     "update-notifier": "^2.3.0",
     "url-loader": "^0.5.8",
     "webpack": "^3.7.0",
-    "webpack-chunk-hash": "^0.4.0",
     "webpack-dev-server": "^2.9.0",
     "webpack-merge": "^4.1.0",
     "webpack-plugin-replace": "^1.1.1",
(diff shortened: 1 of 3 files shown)


prateekbh:
doesn't webpack rely on `-p/NODE.ENV===production` flag to add the hashes?

lukeed:
Yes, but we only care about `preact build` since that's the only time hashing is added to files.

(The DevServer is in-memory, no files added to disk.)

PR #468, opened 2018-01-20
Validate Project Names

milesthedisch:

Add test and validation to `dest` and/or `name` options.

Yes

**Summary**

https://github.com/developit/preact-cli/issues/466

No

**Other information**

- node v6.10.3
- npm 3.10.10
- yarn 1.3.2
- macOS Sierra 10.12.6

### DIFF
--- a/src/commands/create.js
+++ b/src/commands/create.js
@@ -5,6 +5,7 @@ import fs from 'fs.promised';
 import { green } from 'chalk';
 import { prompt } from 'inquirer';
 import { resolve, dirname } from 'path';
+import validateNpmPackageName from 'validate-npm-package-name';
 import asyncCommand from '../lib/async-command';
 import { info, isDir, hasCommand, error, trim, warn } from '../util';
 import { install, initGit, addScripts, isMissing } from './../lib/setup';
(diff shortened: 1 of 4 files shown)


src/commands/create.js
```diff
@@ -89,6 +90,16 @@ export default asyncCommand({
...
 		}
 
+		// Use `--name` value or `dest` dir's name
+		argv.name = argv.name || argv.dest;
+
+		const npmPackageNameErrors = validateNpmPackageName(argv.name).errors;
+
+		if (npmPackageNameErrors) {
+			const npmPackageNameMsg = npmPackageNameErrors.join(', ');
+			return error(`${argv.name} is not a valid package name. ${npmPackageNameMsg}`, 1);
```
prateekbh: can you add a `chalk` message before this?

milesthedisch: Not sure what the chalk message would have in it. Doesn't the `error` method from the utils already display the message given?

reznord: You are right. Internally `error` method in `utils` uses chalk itself. 

@prateekbh that is totally valid. 😄

lukeed:
Thank you @milesthedisch ~!

src/commands/create.js
prateekbh: gosh I need to do my revision for this codebase

PR #472, opened 2018-01-27
Prefer c.modules (fixes #467)

### LINKED ISSUE #467: preact build --json fails with `TypeError: c.getModules is not a function`
bug

`preact build --json` fails with `TypeError: c.getModules is not a function`

I made sure I have latest preact-cli and webpack, etc.  I created a new preact-cli app with `preact create default my-project` ran `preact build --json` got error.

to actually output build stats json file

- Node version
v9.4.0
- npm version
5.6.0
- Operating system
MacOS 10.13.2 (17C205)
- webpack version
3.10.0
- CLI version
preact-cli 2.1.1
- Browser

developit:
This should fix #467.

### DIFF
--- a/src/lib/webpack/run-webpack.js
+++ b/src/lib/webpack/run-webpack.js
@@ -136,7 +136,7 @@ export function writeJsonStats(stats) {
 	function strip(stats) {
 		stats.modules.forEach(stripLoaderFromModuleNames);
 		stats.chunks.forEach(c => {
-			(c.mapModules!=null ? c.mapModules(Object) : c.getModules()).forEach(stripLoaderFromModuleNames);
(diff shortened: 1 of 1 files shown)


src/lib/webpack/run-webpack.js
```diff
@@ -136,7 +136,7 @@ export function writeJsonStats(stats) {
 	function strip(stats) {
 		stats.modules.forEach(stripLoaderFromModuleNames);
 		stats.chunks.forEach(c => {
-			(c.mapModules!=null ? c.mapModules(Object) : c.getModules()).forEach(stripLoaderFromModuleNames);
+			(c.modules || (c.mapModules!=null ? c.mapModules(Object) : c.getModules())).forEach(stripLoaderFromModuleNames);
```
reznord: `getModules()` is deprecated right? 

That's reason why we see that weird message when we run `preact build`

PR #489, opened 2018-02-08
Prevent Route Flickers (V2) 🎉

lukeed:
Changes our `Async` component so that it `dangerouslySetsInnerHTML` with the existing (SSR) markup. The second cycle is initiated via `require.ensure`'s payload, which `Async` uses to update `state.child`. At this point, we've loaded the `route-*` chunk and have a client-side Preact app.

For sequential routing, the old/stale content _is_ removed while the new chunk downloads. IMO, this is a better approach than holding onto the content until the update is ready, as it would feel frozen/stuck to the visitor.

Contrary to #364, this approach ***does not*** mount the entire `<App/>` twice. 👍 

Demo: [_Video_](https://www.dropbox.com/s/9t1fsudsm2e1czf/route-async.mp4?dl=0)

---

_Closes #364, #281_

### DIFF
--- a/src/components/async.js
+++ b/src/components/async.js
@@ -10,17 +10,17 @@ export default function(req) {
 			req(m => {
 				this.setState({ child: m.default || m });
 			});
-		}
+		};
 
 		this.shouldComponentUpdate = (_, nxt) => {
 			nxt = nxt.child === void 0;
 			if (nxt && old === void 0) {
-				old = (0, _preact.h)(b.nodeName, { dangerouslySetInnerHTML: { __html: b.innerHTML } });
+				old = h(b.nodeName, { dangerouslySetInnerHTML: { __html: b.innerHTML } });
 			} else {
 				old = ''; // dump it
 			}
 			return !nxt;
-		}
+		};
 
 		this.render = (p, s) => s.child ? h(s.child, p) : old;
 	}
--- a/src/components/async.js
+++ b/src/components/async.js
@@ -1,16 +1,29 @@
 import { h, Component } from 'preact';
 
-export default function(load) {
+export default function(req) {
 	function Async() {
 		Component.call(this);
-		let done = child => {
-			this.setState({ child: child && child.default || child });
-		};
-		let r = load(done);
-		if (r && r.then) r.then(done);
+
+		let b, old;
+		this.componentWillMount = () => {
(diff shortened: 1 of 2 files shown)


prateekbh:
I have a question:
If we go from SSR to a client render, this stops rendering since next.child will not be there.. right?
If yes, what happens when I go from Route A -> Route B??? I assume same condition will occur again and nothing will change till Route B's script is downloaded. Am i right?

lukeed:
If you try navigating _before_ the route-chunk has finished loading (aka, before Preact has taken control of the page), you will make a server request to the next page (eg, `/profile`), effectively a hard-reload on the new url. 

It's a "web 1.0" site until Preact/JavaScript takes control of the routing.

prateekbh:
nups, my question is after that.

If all the client js is loaded and all is good. Now if i navigate to another route, say "Route B". Will I keep looking at current route till Route B's script is downloaded?

lukeed:
Ah, my bad, misunderstood.

So yeah, once Preact is initialized, my 2nd paragraph applies:

> For sequential routing, the old/stale content is removed while the new chunk downloads. IMO, this is a better approach than holding onto the content until the update is ready, as it would feel frozen/stuck to the visitor.

You'll see _no_ content while in between routes.

I think this is better approach than staring at "A" while "B" is being prepared.

If others disagree, it's simply a matter of updating the `shouldComponentUpdate` booleans.

lukeed:
You can see this in my video preview around the 7 second mark.

prateekbh:
may be just a shameless plug, but should we just add a loader prop as well? which one can show while sequential routing happens.

developit:
@prateekbh That'd go in preact-router, but I'd be down. Maybe it could be a reason to make `<Route>` useful, right now it's just a passthrough.

prateekbh:
@developit that's one way of doing it, but i just feel a contextual loader defination with every route just expresses the whole thing in a cleaner way.

but yeah nothing blocking this PR

ajoslin:
Testing this out now.

developit:
I like this, can't find any reason not to go for it. The cWM setup is key since a preloaded chunk (or a previously used one) will render fully on first render due to the use of callbacks 🔥

lukeed:
Tested locally in all my projects, works well & can't think of a simpler/better approach~!

PR #490, opened 2018-02-09
prevent TypeError when stats is undefined

DemianD:
Bugfix

No. But I will if this is necessary for this.

**Summary**
As I'm working on my personal website, I'm trying to implement the [preact-cli-plugin-critical-css](https://github.com/matthewlynch/preact-cli-plugin-critical-css) plugin.

I was facing an error: 
```shell
Fatal TypeError: Cannot read property 'toJson' of undefined
```

This was caused by this lines: https://github.com/developit/preact-cli/blob/master/src/lib/webpack/run-webpack.js#L109-L119

There was an error, but stats is undefined. As there is an error, we will also show the stats.
This PR will only show stats when there are stats.

**Output before:**
image

**Output after:**
image

No

**Other information**
- Node version: v8.0.0, v6.7.0
- npm version: 5.0.0
- CLI version: latest
- Operating System: macOS High Sierra: 10.13.2
- Browser: Irrelevant

### DIFF
--- a/src/lib/webpack/run-webpack.js
+++ b/src/lib/webpack/run-webpack.js
@@ -106,8 +106,11 @@ async function prodBuild(env) {
 
 const runCompiler = compiler => new Promise((res, rej) => {
 	compiler.run((err, stats) => {
-		if (err || stats.hasErrors()) {
+		if (stats && stats.hasErrors()) {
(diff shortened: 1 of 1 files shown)


developit:
LGTM!

PR #497, opened 2018-02-12
Enable webpack devServer overlay for compiler errors

andreruffert:
Shows a full-screen overlay in the browser when there are compiler errors.
https://webpack.js.org/configuration/dev-server/#devserver-overlay

screen shot 2018-02-12 at 22 16 38

### DIFF
--- a/src/lib/webpack/webpack-client-config.js
+++ b/src/lib/webpack/webpack-client-config.js
@@ -185,7 +185,7 @@ function isDev(config) {
 			historyApiFallback: true,
 			quiet: true,
 			clientLogLevel: 'none',
-			overlay: false,
+			overlay: true,
 			stats: 'minimal',
 			watchOptions: {
 				ignored: [


lukeed:
I personally like having the overlay, but I think it was turned off by default because some people find it annoying?

Just in case you didn't already know, you can do this for your own local config:

```js
// preact.config.js

module.exports = function (config) {
  // turn on overlay!
  config.devServer.overlay = true;
}
```

andreruffert:
Whoops I did not know about the local config thanks for pointing it out. 
To have it configurable makes sense 👌 🙃

lukeed:
All good 😆 Thanks for the intent tho!

developit:
I never knew about that Webpack feature! I always have my console open though

lukeed:
haha yeah, it's pretty nifty. I like hiding my console since I tend to already have 40 tabs open. That _one_ thing makes a difference... 🙄

ForsakenHarmony:
So should we possibly merge this or no, did we ever have it?

lukeed:
We never had it. I wouldn't mind having it, but again, I know quite a few people who can't stand it 😆

andreruffert:
I'm used to it from custom webpack setups and CRA. My terminal process running preact/webpack is not focused all the time thats why I prefer seeing compiler errors directly in the browser. It makes my developer experience definitely smoother and saves me a little bit of time. 

To be honest I never thought about that some people could find it annoying. I personally don't see any disadvantages but of course there are always different opinions. Anyways my guess is that the majority would like to have this enabled by default.

PR #543, opened 2018-04-09
closes #474 - add index.html to default ignore list for SWPrecacheWeb…

### LINKED ISSUE #474: index.html should be excluded from service worker cache

bug

bundle.{hash}.js tries to load the old version of route-{page}.chunk.{hash}.js when JS code is modified, which results in an error.

1. Create new project from template (I used [material](https://github.com/preactjs-templates/material))
1. Run the project on localhost (`$ npm run serve`)
    - The following JS files are created
        - bundle.2eb65.js
        - route-home.chunk.f678b.js
        - route-profile.chunk.c26a1.js
1. Open https://localhost:8080 in browser
    - index.html, the JS files listed above, and other resources are cached by service worker
1. Edit src/routes/profile/index.js (I added `console.log('something has changed');` at `render()` method)
1. Rerun `$ npm run serve`
    - The JS files are updated as follows
        - bundle.2eb65.js -> bundle.a5db3.js
        - route-home.chunk.f678b.js (unchanged)
        - route-profile.chunk.c26a1.js -> route-profile.chunk.27ce5.js
1. Reload https://localhost:8080 in browser
    - index.html cached at the step 3 is used as the response for https://localhost:8080
    - Cached index.html requests bundle.2eb65.js (the old version), which is also restored from the service worker cache
    - In the background, service worker is updated and activated immediately due to `skipWaiting()`, which triggers deletion of the old caches (such as route-profile.chunk.c26a1.js)
1. Click link to the Profile page
    - bundle.2eb65.js (the old version) requests route-profile.chunk.c26a1.js (the old version), which exists neither on the server nor in the service worker cache
        - At this point an error occurs.

All the client side code should be updated to the latest version when JS code is modified.

Excluding index.html from the service worker cache (i.e., adding `/index\.html$/` [here](https://github.com/developit/preact-cli/blob/741922b82697c7e33597e104453bbd1536c21f3a/src/lib/webpack/webpack-client-config.js#L148-L153)), which prevents the old index.html from requesting the old bundle at the step 6-ii, should fix the problem.

osdevisnot:
closes #474 - add index.html to default ignore list for SWPrecacheWebpackPlugin

bugfix

yes (appears to be working ok in my local tests)

**Summary**
as documented in #474 caching index.html causes issues with service worker caching.

**Other information**

### DIFF
--- a/src/lib/webpack/webpack-client-config.js
+++ b/src/lib/webpack/webpack-client-config.js
@@ -154,6 +154,7 @@ function isProd(config) {
 				minify: true,
 				stripPrefix: config.cwd,
 				staticFileGlobsIgnorePatterns: [
+					/index\.html$/,
 					/polyfills(\..*)?\.js$/,
 					/\.map$/,
 					/push-manifest\.json$/,


prateekbh:
But this still allows it in `navigationFallback`, so the offline behavior is not broken right?

lukeed:
I believe that offline ability is broken without `index.html` being saved. Otherwise it won't have anything to serve.

I have a new project open, lemme confirm quickly.

osdevisnot:
my bad, did not test offline mode. let me find out better solution.

osdevisnot:
This works correctly as designed.

PR #550, opened 2018-04-11
closes #432 Add host option to serve command

### LINKED ISSUE #432: Replace localhost with a local domain name?
a question, but may be a feature if it's not yet supported

I don't found how to serve on a specific URL rather than localhost. There is also your auto-SSL generation that doesn't works for me ( /usr/local/share/ca-certificates/devcert.cer doesn't exists on my Fedora 27 ) 

yarn serve

An option, like « yarn serve --host myproject.local » should be really useful if this doesn't yet exists.

**If this is a feature request, what is motivation or use case for changing the behavior?**
 It's better to isolate each dev project on different URL, to separate correctly each service workers. I already have self-signed SSL certs on my local hostname.

- node version 8.7.0
- npm version 5.4.2 but I uses yarn 1.3.2
- Operating system Fedora 27

osdevisnot:
feature: Adds a --host options to `serve` command

No

**Summary**
This fixes #432 

No

**Other information**
When testing with `simplehttp2server` I am running into [SSL warning on localhost](https://github.com/GoogleChromeLabs/simplehttp2server#that-browser-warning) and it seems normal

### DIFF
--- a/src/commands/serve.js
+++ b/src/commands/serve.js
@@ -35,6 +35,11 @@ export default asyncCommand({
 			description: 'Directory or filename where firebase.json should be written\n  (used for --server config)',
 			defaultDescription: '-'
 		},
+		host: {
+			description: 'Hostname to start a server on.',
+			defaultDescription: 'HOST || localhost',
+			alias: 'H'
+		},
 		port: {
 			description: 'Port to start a server on.',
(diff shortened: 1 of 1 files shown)


osdevisnot:
Not relevant for next release (serve is being deprecated).

PR #564, opened 2018-04-29
webpack4

ForsakenHarmony:


packages/cli/lib/lib/babel-config.js
```diff
@@ -27,11 +27,11 @@ module.exports = function (env, options={}) {
 			require.resolve('babel-plugin-transform-react-constant-elements'),
 			isProd && require.resolve('babel-plugin-transform-react-remove-prop-types'),
 			[require.resolve('babel-plugin-transform-react-jsx'), { pragma: 'h' }],
-			[require.resolve('babel-plugin-jsx-pragmatic'), {
-				module: 'preact',
-				export: 'h',
-				import: 'h'
-			}]
+			// [require.resolve('babel-plugin-jsx-pragmatic'), {
```
reznord: Is this commented out by mistake?

ForsakenHarmony: no, intentional, webpack was throwing errors about `h` already being defined, so I used the webpack provide plugin instead

lukeed:
I've checked `sw.js` and prerendered outputs as well, and they're correct & working perfectly 👌

packages/cli/lib/lib/webpack/push-manifest.js
```diff
@@ -29,14 +31,14 @@ module.exports = class PushManifestPlugin {
...
-						type: 'script',
-						weight: 0.9
-					}
-				});
+			let path, css, obj;
+			scripts.forEach((filename, idx) => {
+				css = styles[idx];
+				obj = Object.assign({}, defaults);
+				obj[filename] = { type:'script', weight:0.9 };
+				if (css) obj[css] = { type:'style', weight:0.9 };
```
lukeed: This part _may_ be risky. In the event a route emits multiple `*.chunk.*.css` files, then the indexing won't align.

I'm not sure if this will happen, but I'm assuming it _might_ at some point~

lukeed:
I went ahead & made the fixes for the above issue.

Also left a comment where issues could arise in the future... not sure.

packages/cli/lib/lib/webpack/push-manifest.js
ForsakenHarmony: always expect the unexpected

ForsakenHarmony: also can't we just leave the let inside the loop, this just looks ugly and I don't think this needs optimization

lukeed: Haha, it doesn't matter. No side effects in this loop, and multiple `let`s inside the loop looks ugly to me 😉

ForsakenHarmony: so we should have let inside the loop for consistent code style? :p

PR #569, opened 2018-05-13
Add `build --analyzer` flag 

lukeed:
Installs & uses [`webpack-bundle-analyzer`](https://github.com/webpack-contrib/webpack-bundle-analyzer) to start an interactive treemap that allows the Dev to inspect any/all bundle contents. This is probably the most popular package of this sort, and IMO it's a lot friendlier/nicer to use than our current `--json` flag & a message to import the output to an external tool.

> FWIW, our stats output has been buggy in the past and (I think) is pretty fragile since it relies on internals so closely. Maybe remove `--json` flag moving forward? The new plugin can spit out a similar report.

Here's a screenshot of the `default` template. I believe the deep filenames are the result of my quick testing with custom `cwd`:

```
node packages/cli build --cwd foobar --analyze 
```

screen shot 2018-05-13 at 9 28 54 am

### DIFF
--- a/packages/cli/lib/index.js
+++ b/packages/cli/lib/index.js
@@ -28,6 +28,7 @@ prog
 	.option('--sw', 'Generate and attach a Service Worker', true)
 	.option('--json', 'Generate build stats for bundle analysis')
 	.option('--template', 'Path to custom HTML template')
+	.option('--analyze', 'Launch interactive Analyzer to inspect production bundle(s)')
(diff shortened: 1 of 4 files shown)


marvinhagemeister:
@lukeed you rock! I was just thinking about this the other day and here you, already with PR 🎉 I have nothing to add, besides that I really appreciate the work you and the others put into `preact-cli` 💯

lukeed:
Thanks @marvinhagemeister~! You do far more, so thank _you_!

PR #571, opened 2018-05-15
adds shrink to fit to viewport

PeteSchuster:
forces iOS Safari to work like other browsers

nope

**Summary**
shrink-to-fit=no fixes a "feature" in iOS Safari.

more info/discussion:
https://github.com/h5bp/html5-boilerplate/pull/1984
and
https://bitsofco.de/ios-safari-and-shrink-to-fit/

nope

### DIFF
--- a/src/resources/template.html
+++ b/src/resources/template.html
@@ -3,7 +3,7 @@
 	<head>
 		<meta charset="utf-8">
 		<title><%= htmlWebpackPlugin.options.title %></title>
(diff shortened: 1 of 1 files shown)


hassanbazzi:
@PeteSchuster This is something that users can simply over-ride using a template.html of their own choosing.

Although I do agree it is a somewhat sane default, usually the fix to these kinds of overflow issue is to fix the code itself. 

Latest Safari has this fixed too. Feel free to re-open if you feel this is a necessary change.

PR #573, opened 2018-05-16
next: Set reduce_vars to false for uglify compress options

lukekarrys:
Based on https://github.com/developit/preact/issues/961

bugfix

no

**Summary**

I was trying out the `next` tag and found out that I was running into the same issues as https://github.com/developit/preact/issues/961 when minifying and running my app. Setting `reduce_vars` to `false` fixed it for me.

Possibly. But considering it would be released as the next major version, I think it should be ok.

**Other information**

I wish I understood more about *why* this was happening, and why this fix works. https://github.com/developit/preact/issues/961#issuecomment-358615579 has a good explanation, but this is my first time using preact and webpack 4 so the explanation is a little over my head.

### DIFF
--- a/packages/cli/lib/lib/webpack/webpack-client-config.js
+++ b/packages/cli/lib/lib/webpack/webpack-client-config.js
@@ -139,6 +139,7 @@ function isProd(config) {
 							hoist_funs: true,
 							if_return: true,
 							join_vars: true,
+							reduce_vars: false,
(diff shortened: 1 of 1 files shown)


developit:
This bug is no longer triggered as of [Preact 8.3.0](https://github.com/developit/preact/releases/tag/8.3.0). Think we can close it?

PR #583, opened 2018-06-01
Esm builds

prateekbh:
Introduces babel7 and build mirrored es6 bundles for eeverything. Ships es6 with `<script type=module`

No, The snapshot thing is flaky with the `babel-esm-plugin`.

**Summary**

This is part 1 of 2 PRs to come which will put a new flag(`--esm`) in `preact build`. This will ship es6 and es5 code both using `module/nomodule`.

There will be followup PRs to compliment the functionality in SW

No.

**Other information**
Spiderman dies in Avengers infinity war. #Sorry


packages/cli/lib/lib/webpack/webpack-client-config.js
```diff
@@ -75,7 +76,11 @@ function clientConfig(env) {
 
 		plugins: [
 			...RenderHTMLPlugin(env),
-			new PushManifestPlugin(),
+      new PushManifestPlugin(),
+      new BabelEsmPlugin({
+        filename: '[name].esm.js',
+        chunkFilename: '[name].esm.js'
+      }),
```
ForsakenHarmony: should be tabs here I think

ForsakenHarmony:
very nice

prateekbh:
About the flag... This experiment is so new that there definitely should be a way to turn it off... So may be a flag with default on

reznord:
+1 for the `--esm` flag on `preact build` command. 

We can leave it off by default and we can enable it by the `--esm` flag (just my thoughts)

packages/cli/lib/index.js
```diff
@@ -23,14 +23,15 @@ prog
...
 	.option('--cwd', 'A directory to use instead of $PWD', '.')
-	.option('--sw', 'Generate and attach a Service Worker', true)
+  .option('--sw', 'Generate and attach a Service Worker', true)
 	.option('--json', 'Generate build stats for bundle analysis')
 	.option('--template', 'Path to custom HTML template')
 	.option('--analyze', 'Launch interactive Analyzer to inspect production bundle(s)')
 	.option('--prerenderUrls', 'Path to pre-rendered routes config', 'prerender-urls.json')
-	.option('-c, --config', 'Path to custom CLI config', 'preact.config.js')
+  .option('-c, --config', 'Path to custom CLI config', 'preact.config.js')
+  .option('--esm', '[HIGHLY EXPERIMENTAL] Builds ES-2015 bundles for your code.', 'esm')
```
ForsakenHarmony: spaces instead of tabs again 👀

prateekbh: Yeah I'll fix my editor... 😑.

packages/cli/package.json
```diff
@@ -47,18 +47,18 @@
...
-    "babel-plugin-syntax-dynamic-import": "^6.18.0",
-    "babel-plugin-transform-class-properties": "^6.24.1",
-    "babel-plugin-transform-decorators-legacy": "^1.3.4",
     "babel-plugin-transform-export-extensions": "^6.22.0",
-    "babel-plugin-transform-object-assign": "^6.22.0",
-    "babel-plugin-transform-object-rest-spread": "^6.26.0",
-    "babel-plugin-transform-react-constant-elements": "^6.23.0",
-    "babel-plugin-transform-react-jsx": "^6.24.1",
     "babel-plugin-transform-react-remove-prop-types": "^0.4.5",
-    "babel-preset-env": "^1.3.3",
```
reznord: Isn't this required? Preact build command fails at the moment

ForsakenHarmony: look up a couple of lines, has to be changed in the babel config though

ForsakenHarmony: https://github.com/developit/preact-cli/pull/583/files#diff-7234cda78ba01b8899055aae82c84a7fL7

prateekbh: oh yeah...

prateekbh:
Requesting a round of review... Tested, Works fine for me.

PR #592, opened 2018-06-28
Added missing CSS loaders

alvinsight:

This PR adds the loaders for the CSS preprocessors that are currently in the base Webpack config.

**Summary**

At the moment, despite the fact that webpack is setup to include alternative ways to write CSS (sass, stylus, less) the loaders in questions aren't in the `package.json`, so importing a `.sass` file will throw an error and halt the dev server/build process.

No.

**Other information**

Latest CLI version as of today (2.2.1)


lukeed:
Thank you, but this is by design. You are to manually install the loader and preprocessor tool if you want to use it, then it'll just work. Otherwise everyone is forced to download all loaders despite their preference or choice to use one at all.

alvinsight:
Right, it makes sense, can I add something in the readme, then ? 
My problem was that it threw an error and broke, which made me think it was an oversight.

developit:
Originally we had this set up to install the missing loaders and then rebuild, but I don't think that works in recent Webpack iterations.  Perhaps the simplest solution here would be to print out installation instructions when encountering the missing loaders?

ForsakenHarmony:
Yes either readme or when the error happens, but I don't know how easy it is to detect those errors specifically

alvinsight:
@developit Totally, that does sound like the simplest/best solution 👍

ppsirius:
This line for SASS in README should be enough 🙈 
`npm install --save-dev node-sass sass-loader`

hassanbazzi:
Closing this in favor of a Readme PR.

PR #606, opened 2018-08-04
Build service worker for --esm

prateekbh:

This completes the `service-worker` story for the `--esm` builds.

No 🤓 

**Summary**
Now there will be two service workers `sw` and `sw-esm` which will precache es5 and esm builds respectively. This will also register the right service worker based on the bundle.

Also this introduces an environment variable which tells the code which build is being executed. ES5 or ESM.

Nah!

### DIFF
--- a/packages/cli/lib/lib/webpack/webpack-client-config.js
+++ b/packages/cli/lib/lib/webpack/webpack-client-config.js
@@ -169,6 +169,7 @@ function isProd(config) {
 				minify: true,
 				stripPrefix: config.cwd,
 				staticFileGlobsIgnorePatterns: [
+          /\.esm\.js$/,
 					/polyfills(\..*)?\.js$/,
 					/\.map$/,
 					/push-manifest\.json$/,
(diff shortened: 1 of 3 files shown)


packages/cli/lib/lib/webpack/webpack-client-config.js
```diff
@@ -183,9 +185,41 @@ function isProd(config) {
...
 			new BabelEsmPlugin({
 				filename: '[name].[chunkhash:5].esm.js',
-				chunkFilename: '[name].chunk.[chunkhash:5].esm.js'
+				chunkFilename: '[name].chunk.[chunkhash:5].esm.js',
+				beforeStartExecution: plugins => {
+					plugins.forEach(plugin => {
+						if (plugin.constructor.name === 'DefinePlugin' && plugin.definitions) {
+							for (const definition in plugin.definitions) {
+								if (definition === 'process.env.ES_BUILD') {
+									plugin.definitions[definition] = true;
```
lukeed: I believe this is the cause of #622

lukeed: AKA, forcing `ES_BUILD` to always be true, even when not flagged. And in `lib/entry.js`, we're looking for `sw-esm.js` when `ES_BUILD` is true.

packages/cli/lib/lib/webpack/webpack-client-config.js
```diff
@@ -183,9 +185,41 @@ function isProd(config) {
 		prodConfig.plugins.push(
 			new BabelEsmPlugin({
 				filename: '[name].[chunkhash:5].esm.js',
-				chunkFilename: '[name].chunk.[chunkhash:5].esm.js'
+				chunkFilename: '[name].chunk.[chunkhash:5].esm.js',
```
developit: `.mjs`?

prateekbh: Webpack has issues with mjs IIRC

packages/cli/lib/lib/webpack/webpack-client-config.js
prateekbh: But the Babel esm plugin itself is protected by `config.esm` condition above. So for a normal build it should not execute

packages/cli/lib/lib/webpack/webpack-client-config.js
SleeplessByte: https://github.com/webpack/webpack/issues/5686
https://github.com/webpack/webpack/pull/4691

Works fine in Webpack 4 :)

PR #612, opened 2018-08-09
Add host option to superstatic server

MaxMilton:

Feature.

No. The change is simple enough and it could be challenging to test connections from a different host.

**Summary**

Let me start this by saying I'm a fan of `preact-cli` and the simplicity it provides developers to get started quickly. As it stands now, it's set up for local development which is fine for the majority of use cases.

In my specific case, my team is developing an app for smart TV devices and need to serve to them over a LAN. We _could_ set up a proxy to do this but why complicate things when it's simple to do directly from `superstatic`. It would be a similar issue for any developers wishing to develop on mobile.

The caveat here is only `superstatic` supports the `host` option (much like only `simplehttp2server` supports the `cors` option). I updated the async-command description to reflect that.

No.

**Other information**

Not relevant.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,9 @@ $ preact serve
               [options: "simplehttp2server", "superstatic", "config"]         [string]  [default: "simplehttp2server"]
   --dest      Directory or filename where firebase.json should be written
               (used for --server config)                                      [string]  [default: -]
+  --host, -h  Host to allow connections from (superstatic only).              [string]  [default: HOST || "localhost"]
   --port, -p  Port to start a server on.                                      [string]  [default: PORT || 8080]
+  --cors      Set allowed origins (simplehttp2server only).                   [string]  [default: "https://localhost:${PORT}"]
 ```
 
 #### preact list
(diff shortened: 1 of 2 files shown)


ForsakenHarmony:
sorry but we removed the serve command (temporarily) in the next branch, so not sure if it would ever get released

MaxMilton:
@ForsakenHarmony thanks for the heads up. I'll close this for now and review things again once the [next branch](https://github.com/developit/preact-cli/tree/next) is more ready.

Looks like I'll need to put together a separate package for our team.

MaxMilton:
FYI, for anyone who looks at this later, I ended going with <https://github.com/lukeed/sirv/tree/master/packages/sirv-cli>. So long as what you need to serve is completely static, this solution has better performance while still providing most features you may require.

Add the dependency:
```sh
yarn add sirv-cli
```

Add this to your `package.json` `"scripts"`:
```sh
"serve": "sirv start build --single --port 1337 --cors --etag --maxage",
```

PR #619, opened 2018-08-21
Esm push fixes

prateekbh:
**Bugfix**

Yes.

**Summary**
- With the addition of `--esm` we need to push module bundle instead of regular ones.
I also cleaned `template.html` a bit.
- This also adds the needed `crossorigin` attribute to preload the module script.

No.

### DIFF
--- a/packages/cli/lib/lib/webpack/create-load-manifest.js
+++ b/packages/cli/lib/lib/webpack/create-load-manifest.js
@@ -32,7 +32,7 @@ module.exports = (assets, isESMBuild = false) => {
     },
   },
   manifest = {
-  '/': defaults
+    '/': defaults
   };
 
   let path, css, obj;
@@ -42,8 +42,7 @@ module.exports = (assets, isESMBuild = false) => {
     obj[filename] = { type:'script', weight:0.9 };
     if (css) obj[css] = { type:'style', weight:0.9 };
     path = filename.replace(/route-/, '/')
-      .replace(/\.chunk(\.\w+)?\.js$/, '')
-      .replace(/\.chunk(\.\w+)?\.esm\.js$/, '')
+      .replace(/\.chunk(\.\w+)?(\.esm)?\.js$/, '')
       .replace(/\/home/, '/');
     manifest[path] = obj;
   });
--- a/packages/cli/lib/lib/webpack/create-load-manifest.js
+++ b/packages/cli/lib/lib/webpack/create-load-manifest.js
@@ -1,15 +1,20 @@
-module.exports = assets => {
+module.exports = (assets, isESMBuild) => {
   let mainJs, mainCss, scripts=[], styles=[];
   for (let filename in assets) {
     if (!/\.map$/.test(filename)) {
-      if (/route-/.test(filename)) {
+      if (/route-/.test(filename) && !isESMBuild) {
+        scripts.push(filename);
+      } else if (/route-(.+)\.esm\.js/.test(filename) && isESMBuild) {
         scripts.push(filename);
       } else if (/chunk\.(.+)\.css$/.test(filename)) {
         styles.push(filename);
       } else if (/^bundle(.+)\.css$/.test(filename)) {
         mainCss = filename;
-      } else if (/^bundle(.+)\.js$/.test(filename)) {
+      } else if (!isESMBuild && /^bundle(.+)\.js$/.test(filename)) {
         mainJs = filename;
+      } else if (isESMBuild && /bundle\.\w{5}\.esm\.js/.test(filename)) {
+        mainJs = filename;
+        console.log('setting filename');
       }
     }
   }
--- a/packages/cli/lib/lib/webpack/create-load-manifest.js
+++ b/packages/cli/lib/lib/webpack/create-load-manifest.js
@@ -14,7 +14,6 @@ module.exports = (assets, isESMBuild) => {
         mainJs = filename;
       } else if (isESMBuild && /bundle\.\w{5}\.esm\.js/.test(filename)) {
         mainJs = filename;
-        console.log('setting filename');
       }
     }
   }
--- a/packages/cli/lib/lib/webpack/create-load-manifest.js
+++ b/packages/cli/lib/lib/webpack/create-load-manifest.js
@@ -24,9 +24,7 @@ module.exports = assets => {
       weight: 1
     },
   },
-  manifest = {
-    '/': defaults
-  };
+  manifest = {};
 
   let path, css, obj;
   scripts.forEach((filename, idx) => {
@@ -34,8 +32,15 @@ module.exports = assets => {
     obj = Object.assign({}, defaults);
     obj[filename] = { type:'script', weight:0.9 };
     if (css) obj[css] = { type:'style', weight:0.9 };
-    path = filename.replace(/route-/, '/').replace(/\.chunk(\.\w+)?\.js$/, '').replace(/\/home/, '/');
-    manifest[path] = obj;
+    path = filename.replace(/route-/, '/')
+      .replace(/\.chunk(\.\w+)?\.js$/, '')
+      .replace(/\.chunk(\.\w+)?\.esm\.js$/, '')
+      .replace(/\/home/, '/');
+    if (!manifest[path]) {
+      manifest[path] = obj;
+    } else {
+      manifest[path][filename] = { type:'script', weight:0.9 };
+    }
   });
 
   return manifest;
--- a/packages/cli/lib/lib/webpack/create-load-manifest.js
+++ b/packages/cli/lib/lib/webpack/create-load-manifest.js
@@ -1,19 +1,22 @@
-module.exports = (assets, isESMBuild) => {
+const isESM = filename => /\.esm\.js$/.test(filename);
+const isMatch = (filename, condition) => (isESM(filename) === condition);
+
+module.exports = (assets, isESMBuild = false) => {
   let mainJs, mainCss, scripts=[], styles=[];
   for (let filename in assets) {
     if (!/\.map$/.test(filename)) {
-      if (/route-/.test(filename) && !isESMBuild) {
-        scripts.push(filename);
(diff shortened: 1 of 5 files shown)


packages/cli/lib/lib/webpack/create-load-manifest.js
```diff
@@ -1,14 +1,18 @@
...
+      } else if (/route-(.+)\.esm\.js/.test(filename) && isESMBuild) {
         scripts.push(filename);
       } else if (/chunk\.(.+)\.css$/.test(filename)) {
         styles.push(filename);
       } else if (/^bundle(.+)\.css$/.test(filename)) {
         mainCss = filename;
-      } else if (/^bundle(.+)\.js$/.test(filename)) {
+      } else if (!isESMBuild && /^bundle(.+)\.js$/.test(filename)) {
+        mainJs = filename;
+      } else if (isESMBuild && /bundle\.\w{5}\.esm\.js/.test(filename)) {
```
hassanbazzi: Such a crazy if statement we got here lol

prateekbh: heh.. i am not the best regexp maker out thr 😂

packages/cli/lib/resources/template.html
```diff
@@ -27,8 +28,8 @@
 		}) %>
 		<% if (webpack.assets.filter(entry => entry.name.match(/bundle.\w{5}.esm.js$/)).length > 0) { %>
 			<!-- Fix for safari < 11 nomodule bug. TODO: Do the following only for safari.-->
-			<script>!function(){var e=document,t=e.createElement("script");if(!("noModule"in t)&&"onbeforeload"in t){var n=!1;e.addEventListener("beforeload",function(e){if(e.target===t)n=!0;else if(!e.target.hasAttribute("nomodule")||!n)return;e.preventDefault()},!0),t.type="module",t.src=".",e.head.appendChild(t),t.remove()}}();</script>
-			<script src="<%= htmlWebpackPlugin.files.publicPath %><%= webpack.assets.filter(entry => entry.name.match(/bundle.\w{5}.esm.js$/))[0].name %>" type="module"></script>
+			<script defer>!function(){var e=document,t=e.createElement("script");if(!("noModule"in t)&&"onbeforeload"in t){var n=!1;e.addEventListener("beforeload",function(e){if(e.target===t)n=!0;else if(!e.target.hasAttribute("nomodule")||!n)return;e.preventDefault()},!0),t.type="module",t.src=".",e.head.appendChild(t),t.remove()}}();</script>
```
rodush: `defer` does not make any sense for inline script, only to those with the `src` attribute. Recently found out for myself.

packages/cli/lib/lib/webpack/create-load-manifest.js
```diff
@@ -24,18 +28,23 @@ module.exports = assets => {
...
     css = styles[idx];
     obj = Object.assign({}, defaults);
     obj[filename] = { type:'script', weight:0.9 };
     if (css) obj[css] = { type:'style', weight:0.9 };
-    path = filename.replace(/route-/, '/').replace(/\.chunk(\.\w+)?\.js$/, '').replace(/\/home/, '/');
-    manifest[path] = obj;
+    path = filename.replace(/route-/, '/')
+      .replace(/\.chunk(\.\w+)?\.js$/, '')
+      .replace(/\.chunk(\.\w+)?\.esm\.js$/, '')
+      .replace(/\/home/, '/');
```
lukeed: We already relied on `home.js` before (which we should probably fix), but by removing:

```js
manifest: {
  '/': defaults
}
```

...we're now _requiring_ a file named `routes/home.js` in order to populate a manifest entry for the `/` path.

packages/cli/lib/lib/webpack/create-load-manifest.js
lukeed: For this block & the `route-` one above it, it probably cleaner to do something like this instead:

```js
let isESM = x => /\.esm\.js$/.test(x);
let isMatch = (x, y) => isESM(x) === y;

//---

      if (/route-/.test(filename)) {
        // both ESM & regular match here
        isMatch(filename, isESMBuild) && scripts.push(filename);
      } 
      // ...
      else if (/^bundle(.+)\.js$/.test(filename)) {
        // both bundles match here
        if (isMatch(filename, isESMBuild)) {
          mainJs = filename;
        }
      }
```

packages/cli/lib/lib/webpack/create-load-manifest.js
prateekbh: oh nice catch, this was an intermittent thing when i was trying stuff.
I can put back 
```
manifest: {
  '/': defaults
}
```

Will that address your concern?

packages/cli/lib/resources/template.html
prateekbh: addressed

packages/cli/lib/lib/webpack/create-load-manifest.js
lukeed: I _think_ so, tho I'm not 100% sure

packages/cli/lib/lib/webpack/create-load-manifest.js
prateekbh: done. Thanks for the suggestion

packages/cli/lib/resources/template.html
```diff
@@ -10,9 +10,10 @@
 		<% if (htmlWebpackPlugin.options.manifest.theme_color) { %>
 			<meta name="theme-color" content="<%= htmlWebpackPlugin.options.manifest.theme_color %>">
 		<% } %>
-		<% for (const file in htmlWebpackPlugin.options.createLoadManifest(compilation.assets)[htmlWebpackPlugin.options.url]) { %>
-			<% if (file) { %>
-				<link rel="preload" href="<%= htmlWebpackPlugin.files.publicPath + file %>" as="<%= file.match(/\.css$/)?'style':'script' %>" <%= file.match(/\.esm\.js$/)?'crossorigin="anonymous"':'' %>>
+		<% for (const file in htmlWebpackPlugin.options.createLoadManifest(compilation.assets, htmlWebpackPlugin.options.config.esm)[htmlWebpackPlugin.options.url]) { %>
+			<% if (file && file.match(/\.(css|js)$/)) { %>
+				<!-- crossorigin for main bundle as that is loaded from <script type=module tag, other lazy loaded bundles are from webpack so its not needed -->
```
developit: probably can skip the comment or put it in `<% /* these gross tags */ %>`

developit:
Quick question: is this going to push/preload both the esm and non-esm bundle?

prateekbh:
@developit no it will push only esm bundles

ForsakenHarmony:
How many browsers have push support but not module support?

prateekbh:
@forsakenharmony, there are a good bunch of such browsers. But all last 3 version support both IIRC

prateekbh:
Is this good to merge?

packages/cli/lib/lib/webpack/create-load-manifest.js
```diff
@@ -1,15 +1,22 @@
-module.exports = assets => {
+const isESM = filename => /\.esm\.js$/.test(filename);
```
developit: Not entirely scoped to this PR, but would you be open to using `.mjs`?

packages/cli/lib/lib/webpack/push-manifest.js
```diff
@@ -1,9 +1,12 @@
 const createLoadManifest = require('./create-load-manifest');
 
 module.exports = class PushManifestPlugin {
+  constructor(env = {}) {
```
developit: Spacing seems off here (or it's my phone). Also the underscore likely trips our lint rules.

packages/cli/lib/lib/webpack/create-load-manifest.js
```diff
@@ -34,7 +41,10 @@ module.exports = assets => {
     obj = Object.assign({}, defaults);
     obj[filename] = { type:'script', weight:0.9 };
     if (css) obj[css] = { type:'style', weight:0.9 };
-    path = filename.replace(/route-/, '/').replace(/\.chunk(\.\w+)?\.js$/, '').replace(/\/home/, '/');
+    path = filename.replace(/route-/, '/')
+      .replace(/\.chunk(\.\w+)?\.js$/, '')
```
developit:
```
.replace(/\.chunk(\.\w+)?(\.esm)?\.js$/, '')
```

developit:
I didn't run this locally yet to verify but otherwise just the above nitpicks. 

Also I'm not sure if it's elsewhere but it might be nice to expose `BUNDLE_FORMAT="module"` via `env` in case folks need to change behaviour in response to the format in use.

packages/cli/lib/lib/webpack/create-load-manifest.js
prateekbh: I am open to do this in another PR. IIRC webpack's uglify only runs on .just files by default so I switched.

packages/cli/lib/lib/webpack/push-manifest.js
prateekbh: done.

packages/cli/lib/resources/template.html
prateekbh: done

packages/cli/lib/lib/webpack/create-load-manifest.js
prateekbh: this one isn't working hence keeping as is

prateekbh: nevermind... fixed it

prateekbh:
@developit fixed the nits.
> Also I'm not sure if it's elsewhere but it might be nice to expose BUNDLE_FORMAT="module" via env in case folks need to change behaviour in response to the format in use.

This is exactly what [this](https://github.com/developit/preact-cli/blob/next/packages/cli/lib/lib/webpack/webpack-client-config.js#L112) does, just a different var name.

developit:
Ah!

PR #621, opened 2018-08-22
Bugfix for hydrating async components prerendered to a Text node

developit:
https://preact.slack.com/archives/C3M9NTD16/p1534898944000100


developit:
according to @cj this fixes the issue with hydrating Text nodes 👍

PR #623, opened 2018-08-25
Fix SW registration

### LINKED ISSUE #622: sw-esm.js not found when just running preact build. (next.7)
bug

When I build by running `NODE_ENV=production preact build` and then serve the static files I get this error `"GET /sw-esm.js" Error (404): "Not found"`.  I checked and that file indeed doesn’t exist.  I’m not building it with the `--esm` flag, so shouldn’t it be looking for `/sw.js`?  If I run it with the flag `--esm` everything works just fine as the `/sw-esm.js` file gets built.  I’m just wondering why it’s looking for that file without the build flag.

```
npm i -g preact-cli@next http-server@lates
preact create default preact-next
npm i -D preact-cli@next preact-render-spy@latest
npm i -S preact@latest preact-compat@latest preact-router@latest
NODE_ENV=production preact build
http-server build -g -p 5000
```

It to serve `/build` and look for `/sw.js` and not `/sw-esm.js`

lukeed:

bugfix, Closes #622 

**Summary**

The previous code was producing this within `preact build` output:

```js
navigator.serviceWorker.register(n.p+!1?"sw-esm.js":"sw.js")
```

It is now this:

```js
navigator.serviceWorker.register(n.p+"sw.js");
```

### DIFF
--- a/packages/cli/lib/lib/entry.js
+++ b/packages/cli/lib/lib/entry.js
@@ -6,7 +6,7 @@ if (process.env.NODE_ENV==='development') {
 }
 else if (process.env.ADD_SW && 'serviceWorker' in navigator) {
 	// eslint-disable-next-line no-undef
-	navigator.serviceWorker.register(__webpack_public_path__ + process.env.ES_BUILD? 'sw-esm.js' : 'sw.js');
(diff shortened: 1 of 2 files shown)


prateekbh:
tenor
WHAT!!!!

btw is esm SW working fine too?

lukeed:
:laughing: yup, checked both

prateekbh:
Thank you so much.... 😊

PR #625, opened 2018-08-27
Inline PRERENDER=true env var during prerendering

developit:
Makes a `PRERENDER` env var defined that indicates whether the bundle was compiled for prerendering or client-side usage.  This is useful for avoiding SSR-specific code ending up in client bundles.

### DIFF
--- a/src/lib/webpack/webpack-base-config.js
+++ b/src/lib/webpack/webpack-base-config.js
@@ -204,7 +204,8 @@ export default function (env) {
 		plugins: [
 			new webpack.NoEmitOnErrorsPlugin(),
 			new webpack.DefinePlugin({
-				'process.env.NODE_ENV': JSON.stringify(isProd ? 'production' : 'development')
+				'process.env.NODE_ENV': JSON.stringify(isProd ? 'production' : 'development'),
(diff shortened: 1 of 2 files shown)


src/lib/webpack/webpack-base-config.js
```diff
@@ -204,7 +204,8 @@ export default function (env) {
 		plugins: [
 			new webpack.NoEmitOnErrorsPlugin(),
 			new webpack.DefinePlugin({
-				'process.env.NODE_ENV': JSON.stringify(isProd ? 'production' : 'development')
+				'process.env.NODE_ENV': JSON.stringify(isProd ? 'production' : 'development'),
+				'PRERENDER': env.ssr ? 'true' : 'false'
```
lukeed: `env.ssr.toString()` 🏌️

src/lib/webpack/webpack-server-config.js
```diff
@@ -24,6 +24,7 @@ function serverConfig(env) {
 }
 
 export default function (env) {
+	env = Object.assign({}, env, { ssr: true });
```
lukeed: `Object.assign(env, { ssr:true })` ⛳️

PR #632, opened 2018-09-07
Feature/promote runtime caching

harshitkumar31:
Feature

No
**Summary**
https://github.com/developit/preact-cli/issues/607

**Other information**


packages/cli/lib/resources/preact-cli-sw.js
```diff
@@ -0,0 +1,16 @@
+self.addEventListener('fetch', function(event){
+    const isJsonRequest = event.request.headers.get('content-type') === 'application/json';
+    const isPostRequest = event.request.method === 'POST';
+    event.respondWith(fetch(event.request).then(function(response){
+        return response;
+    }).catch(function(err){
+        console.error(err);
+        if(!navigator.onLine && isJsonRequest) {
```
reznord: Tests will fail while running `eslint` 🙈 

```
packages/cli/lib/resources/preact-cli-sw.js
   7:9   warning  Unexpected console statement  no-console
   8:9   error    Expected space(s) after "if"  keyword-spacing
   9:13  error    Expected space(s) after "if"  keyword-spacing
  10:17  warning  Unexpected console statement  no-console
  12:17  warning  Unexpected console statement  no-console
  16:3   error    Missing semicolon             semi
```

prateekbh:
how do we make sure that this does not ship to production for the web apps and only while developing it?

harshitkumar31:
We only generate the service worker in prod env right?  If we want to ship this only for developers, we might need to add another argument to differentiate.

prateekbh:
yep exactly my concern. I would never want this on my production site because

- extra code which is actually for debugging purpose
- I would wanna my consumers that I am lazy enough to not use this.


Yeah its kinda fizzy how we would want to tackle this TBH, lets echo this back on the issue maybe?

prateekbh:
@harshitkumar31 in case you're still interested in doing this, here's one approach:

1. Add another entry point as `sw_debug` in the webpack config.
2. register the service worker here: https://github.com/developit/preact-cli/blob/6742f94664cf88ca0159f809e715fd441495e653/src/lib/entry.js#L5

P.S: This service worker will be a super simple `fetch` handler which listens to `.catch` and shouts in console if failed request is a POST one.

Also this can be extended to point at a page talking about runtime caching strategies once we have a website.

CC: @lukeed SG?

harshitkumar31:
@prateekbh , Sure I'll update the PR.

packages/cli/lib/lib/webpack/sw_debug.js
```diff
@@ -0,0 +1,16 @@
+self.addEventListener('fetch', function(event){
+    const isJsonRequest = event.request.headers.get('content-type') === 'application/json';
+    const isPostRequest = event.request.method === 'POST';
+    event.respondWith(fetch(event.request).then(function(response){
```
prateekbh: we don't really need the `.then` part at all

packages/cli/lib/lib/webpack/sw_debug.js
```diff
@@ -0,0 +1,16 @@
...
+        return response;
+    }).catch(function(err){
+        console.error(err);
+        if (!navigator.onLine && isJsonRequest) {
+            if (isPostRequest) {
+                console.log('May be you should look into backgroundSync');
+            } else {
+                console.log('Maybe you should look into runtimeCaching');
+            }
+        }
```
prateekbh: return the `error`

packages/cli/lib/lib/webpack/sw_debug.js
```diff
@@ -0,0 +1,16 @@
+self.addEventListener('fetch', function(event){
+    const isJsonRequest = event.request.headers.get('content-type') === 'application/json';
+    const isPostRequest = event.request.method === 'POST';
+    event.respondWith(fetch(event.request).then(function(response){
+        return response;
+    }).catch(function(err){
+        console.error(err);
+        if (!navigator.onLine && isJsonRequest) {
```
prateekbh: `navigator.onLine` check should not be here, it is highly unlikely that a dev machine will go offline

packages/cli/lib/lib/webpack/sw_debug.js
```diff
@@ -0,0 +1,16 @@
...
+            if (isPostRequest) {
```
prateekbh: Nice 👍

packages/cli/lib/lib/webpack/sw_debug.js
```diff
@@ -0,0 +1,16 @@
...
+                console.log('May be you should look into backgroundSync');
```
prateekbh: Lets improve both messages like "PREACT-CLI-FRIENDLY-MESSAGE(IMPROVEMENT): We saw that your GET call failed. If this is due to a network error, may be you should look into runtimeCaching`.

harshitkumar31:
@prateekbh , 
   Agreed. One question though, after adding sw_debug in the entry webpack adds code which I feel is unnecessary. Why can't we just copy this file to the dist folder?

prateekbh:
@harshitkumar31 we can, however this automatically does it for you. 
I don't know if we have something already established to copy some file in our source code to user land. But yes, what you suggest is a better thing to do.

harshitkumar31:
@prateekbh ,
   preact-cli is already using CopyWebpackPlugin to copy the manifest file, we could do the same here.

prateekbh:
Ah yes, makes sense... lets do that!
Also can you post a screen shot of both output for other reviewers?

harshitkumar31:
Adding as entry in webpack config:

screen shot 2018-10-09 at 12 41 25 am

Copying using Copy Webpack plugin:
screen shot 2018-10-09 at 1 03 18 am

packages/cli/lib/lib/webpack/webpack-client-config.js
```diff
@@ -93,6 +92,8 @@ function clientConfig(env) {
 				),
 				// copy any static files
 				existsSync(source('assets')) && { from:'assets', to:'assets' },
+				// copy sw_debug
+				{ from: resolve(__dirname,'../webpack/sw_debug.js'), to:'sw_debug.js' },
```
prateekbh: the source file would be better placed here 'https://github.com/developit/preact-cli/tree/next/packages/cli/lib/resources'

prateekbh:
@harshitkumar31 sorry i meant the screen shot of these console error from an app.
sorry if I wasn't clear

harshitkumar31:
@prateekbh , my bad. Updated the screenshots.

packages/cli/lib/lib/entry.js
```diff
@@ -3,6 +3,8 @@ import { h, render } from 'preact';
 if (process.env.NODE_ENV==='development') {
 	// enable preact devtools
 	require('preact/debug');
+	// eslint-disable-next-line no-undef
+	navigator.serviceWorker.register(__webpack_public_path__ + 'sw_debug.js');
```
developit: should we unregister any existing SW in this case?

packages/cli/lib/lib/webpack/sw_debug.js
```diff
@@ -0,0 +1,13 @@
+self.addEventListener('fetch', function (event) {
+    var isPostRequest = event.request.method === 'POST';
+    event.respondWith(fetch(event.request).catch(function (err) {
+        if (err instanceof TypeError) {
+            if (isPostRequest) {
+                console.log('PREACT-CLI-FRIENDLY-MESSAGE(IMPROVEMENT): We saw that your POST API call failed. This might fail for your users as well due to a network error, may be you should look into backgroundSync');
```
developit: Suggestion:

```js
console.log('⚛️Preact CLI development tip: A POST request just failed. This might fail for your users as well due to a network error. It may be worth exploring the backgroundSync API.');
```

```js
console.log('⚛️Preact CLI development tip: A GET request just failed. This might fail for your users as well due to a network error. It may be worth adding runtimeCaching to your Service Worker.');
```

developit:
This looks neat, but I am wondering: are we duplicating the existing network error logging done by DevTools?

packages/cli/lib/lib/entry.js
prateekbh: makes sense... and just in case someone wants to add their own SW at dev time may be add a flag to disable this?

cc: @harshitkumar31

prateekbh:
@harshitkumar31 can you confirm ☝️. I had that question as well, forgot to ask though 😛

harshitkumar31:
@developit , 
    I'm not sure about which devtools you are talking about. Is it preact/devtools  or chrome devtools?

prateekbh:
chrome devtools. are there double logging of the failure?

harshitkumar31:
@prateekbh ,
   Yes there are double logs,
screen shot 2018-10-10 at 1 22 48 am

prateekbh:
@harshitkumar31 I can merge this once you have a flag to disable this feature but on by default.

prateekbh:
@harshitkumar31 In case you're not having free cycles to take this forward/have other engagements, I'd like to merge this as is and add the flags as a separate PR.

cc: @lukeed @ForsakenHarmony

ForsakenHarmony:
You know you can push commits to this?

prateekbh:
with my code editor??? no i don't actually

ForsakenHarmony:
you can add the remote and should have push permission to the branch

harshitkumar31:
@prateekbh ,
   A little confused about where to add the flag, should I add it webpack-base-config.js ?

prateekbh:
@harshitkumar31 I've added the flag condition, can u just move the file in under [resources](https://github.com/developit/preact-cli/tree/next/packages/cli/lib/resources) as per my comment?
that'll be all.

prateekbh:
I dont know why the error!!!

prateekbh:
fixed it... @ForsakenHarmony @lukeed would someone mind taking another look?

packages/cli/lib/lib/entry.js
```diff
@@ -3,6 +3,14 @@ import { h, render } from 'preact';
 if (process.env.NODE_ENV==='development') {
 	// enable preact devtools
 	require('preact/debug');
+	// only add a debug sw if webpack service worker is not requested.
+	if (!process.env.ADD_SW && 'serviceWorker' in navigator) {
+		// eslint-disable-next-line no-undef
+		navigator.serviceWorker.register(__webpack_public_path__ + 'sw_debug.js');
+	} else if (process.env.ADD_SW && 'serviceWorker' in navigator) {
+		// eslint-disable-next-line no-undef
+		navigator.serviceWorker.register(__webpack_public_path__ + (process.env.ES_BUILD ? 'sw-esm.js' : 'sw.js'));
```
ForsakenHarmony: Oh no, OCD

sw_debug
sw-esm

packages/cli/lib/resources/sw_debug.js
```diff
@@ -0,0 +1,13 @@
+self.addEventListener('fetch', function (event) {
+    var isPostRequest = event.request.method === 'POST';
+    event.respondWith(fetch(event.request).catch(function (err) {
+        if (err instanceof TypeError) {
+            if (isPostRequest) {
+                console.log('⚛️Preact CLI development tip: A POST request just failed. This might fail for your users as well due to a network error. It may be worth exploring the backgroundSync API.');
+            } else {
+                console.log('⚛️Preact CLI development tip: A GET request just failed. This might fail for your users as well due to a network error. It may be worth adding runtimeCaching to your Service Worker.');
```
ForsakenHarmony: Maybe we could link to mdn/chrome dev thing here

packages/cli/lib/lib/entry.js
prateekbh: fixed

packages/cli/lib/resources/sw_debug.js
prateekbh: May be.. not so sure

prateekbh:
Thank you @harshitkumar31 🙂

PR #631, opened 2018-09-07
chore(cli): Fix create test image

reznord:
**What kind of change does this PR introduce?**
Attempt to fix the failing tests

**Did you add tests for your changes?**
No, but fixed the failing tests

**Summary**
Fixes the `create.test.js` static image to map the current output.

**Does this PR introduce a breaking change?**
No

PR #639, opened 2018-09-07
Add "preact/babel

lukeed:

**Summary**

**Other information**


marvinhagemeister:
@lukeed Really appreciate this. Was looking for a dedicated preact babel preset to recommend the other day that works with babel v7 👍

lukeed:
Me too! I want to slice up what I have on PWA's [`preset-preact`](https://github.com/lukeed/pwa/blob/master/packages/preset-preact/index.js) (and [7.x here](https://github.com/lukeed/pwa/pull/30)) into these bite-sized chunks.

Repeating config over & over again across build chains is silly

packages/cli/tests/images/create.js
```diff
@@ -22,8 +22,8 @@ exports.default = [
 	'src/routes/profile/index.js',
 	'src/routes/profile/style.css',
 	'src/style/index.css',
-	'src/tests/__mocks__/browserMocks.js',
-	'src/tests/__mocks__/fileMocks.js',
-	'src/tests/header.test.js'
+	'tests/__mocks__/browserMocks.js',
+	'tests/__mocks__/fileMocks.js',
+	'tests/header.test.js'
```
reznord: You fixed my PR changes as well 😛

reznord: Closing #631

packages/cli/lib/lib/webpack/transform-config.js
```diff
@@ -15,11 +15,6 @@ module.exports = async function (env, config, ssr=false) {
 		throw new Error(`preact-cli config could not be loaded!\nFile ${env.config} not found.`);
 	}
 
-	require('@babel/register')({
-		presets: [[require.resolve('@babel/preset-env'), {
-			"targets": { "node": "current" }
-		}]]
-	});
```
ForsakenHarmony: This will break `preact.config.js` pretty much everywhere, are we gonna change something about that as well?

packages/cli/lib/lib/webpack/webpack-base-config.js
```diff
@@ -29,11 +28,24 @@ module.exports = function (env) {
...
+
+	// support `async` – CLI feature, not Preact-specific
+	babel.plugins.push(
+		[require.resolve('fast-async'), { spec: true }]
+	);
+	babel.presets.forEach(x => {
+		if (Array.isArray(x) && x[0].includes('@babel/preset-env')) {
+			x[1].exclude.push('transform-async-to-generator');
+		}
+	});
```
ForsakenHarmony: also not sure if this is the best idea, especially if someone wants to extend the config and then async breaks for them

lukeed: If they're overwriting, it gets lost anyway. The extend/overwrite happens _after_ this point.. same as before

packages/cli/lib/lib/webpack/transform-config.js
lukeed: Oh, derp. I've always done `module.exports` and forgot we allowed that. If it were up to me, we'd drop it entirely. 

Will add it back.

packages/cli/lib/lib/webpack/webpack-base-config.js
ForsakenHarmony: but if you extend the config you won't have `fast-async` because that will override the default one here

packages/cli/package.json
```diff
@@ -46,6 +50,7 @@
   },
   "dependencies": {
     "@babel/core": "^7.0.0",
+    "@babel/register": "^7.0.0",
```
ForsakenHarmony: what about `@babel/preset-env`? 👀

lukeed: Babel package

packages/cli/lib/lib/webpack/webpack-base-config.js
lukeed: Fast async isn't included in preact preset because it's not a part of getting Preact to run. It's a CLI feature not a preact requirement

ForsakenHarmony: yes, but now if someone makes a custom `.babelrc` and extends from `@preact/babel` they'll get errors

lukeed: Why? This writes into `@preact/babel` 100% of the time.

If someone declares a `.babelrc` file to overwrite what PreactCLI offered, that was ***already*** going to be the case.

Not changing (or looking to change) any logic here. If it should be changed, it should be separate from this PR.

ForsakenHarmony: hmm, makes sense I guess

but they have to completely override `preset-env` to get it back

lukeed: As of now, that's their choice. And is the current behavior

ForsakenHarmony: in theory you are changing the behaviour from before though, `fast-async` was in the extendable config from the cli

lukeed: With this package, one can do this:

```js
{
  "presets": [
    "@preact/babel",
    ["@babel/preset-env", {
      "exclude": [
        "transform-async-to-generator"
       ]
     }]
  ],
  "plugins": [
    ["fast-async", { "spec": true }]
  ]
}
```

And they'd be at the same settings. In ^this state, it's a pointless `.babelrc` file, but ofc this assumes the user had something worth doing.

ForsakenHarmony: is that gonna take the rest of the `preset-env` config from `@preact/babel`?

ForsakenHarmony: so the only thing I want to know is if your example `.babelrc` would basically get turned into this by babel

```js
["@babel/preset-env", {
	"loose": true,
	"modules": false,
    "targets": {
		"browsers": [
			">0.25%",
			"not ie 11'"
			"not op_mini all"
		]
	}
	"exclude": [
		"transform-regenerator",
		"transform-async-to-generator",
		"transform-typeof-symbol"
	]
}]
```

and our config doesn't get completely overriden

lukeed:
Does this pass @ForsakenHarmony inspection now?

ForsakenHarmony:
still don't know how I feel about `fast-async` being excluded

also have `babel-plugin-macros` in there now 👀 

I feel like it's best to just have this as like `cli-babel-config` or smth

tomasdev:
IE11 has more user share than Edge 17/18 as the time of writing this. I would not recommend going with that "better" not so better coverage.

ForsakenHarmony:
the difference is, ie 11 keeps us back a lot further than edge, which with that browser list probably doesn't even get special treatment

tomasdev:
I completely understand that, then the question I should have asked is: what's the purpose of preact-cli? towards production applications used by corporations, or more towards prototypes future-thinking that don't care about usage?
Of course it's probably none of the extremes, and in the end it's configurable so... up to the owners. I just threw my opinion.

PR #642, opened 2018-09-08
Inline CSS with Critters Plugin

### LINKED ISSUE #637: Inline critical css
Feature

Currently we ship `<link rel=stylesheet ` the webpack's bundle stylesheet, which has to be parsed before first paint/layout.

- serve any preact-cli app

We can use `critters` to inline just the critical css and then defer load the `bundle.css`

Web perf

- Node version: any
- npm version: any
- Operating system: next
- CLI version: any
- Browser: any

lukeed:

Feature

No

**Summary**

Uses [Critters](https://github.com/GoogleChromeLabs/critters) to inline critical CSS on the client bundle(s) for production _and_ development.

The default settings are good 👌 but can be changed by end-user thru `preact.config.js` if/when needed.

Closes #637 

No


prateekbh:
Quick thing, I'd like the bundle.css to defer loaded since this is inlined anyways. WDYT?

lukeed:
Unless I've missed something, both `defer` and `async` are `<script>`-specific attributes... they won't do anything (for now) on other tags.

The default Critters setup uses `rel=preload as=style` which is as close to `defer` as you can get me thinks, before bringing in any JS

prateekbh:
Hmmm, how about we inline what critters give us and put bundle.css as 
`<line rel='defer-stylesheet' href='...'>` and use js to load this later on?

lukeed:
I think our first-paint is already super fast with default setting. Don't think we should be doing anything custom/beyond what Critters offers, as that's the point/goal of Critters, no?

Either way, sounds like you may want the [`{ preload: 'swap' }`](https://github.com/GoogleChromeLabs/critters#preloadstrategy) option.

prateekbh:
Oh I just read: 
>**default:** Move stylesheet links to the end of the document and insert preload meta tags in their place.

This seems fine, if it is indeed moving it to the end. Please just check that we don't have double preload css tags with `--preload`

prateekbh:
Also, tried this right now and running into: https://github.com/GoogleChromeLabs/critters/issues/5

packages/cli/lib/lib/webpack/webpack-client-config.js
```diff
@@ -75,6 +76,7 @@ function clientConfig(env) {
 		},
 
 		plugins: [
+			new CrittersPlugin(),
```
prateekbh: this should be after `...RenderHTMLPlugin(env),` to resolve the error from https://github.com/GoogleChromeLabs/critters/issues/5

prateekbh:
btw there should be an opt-out/opt-in flag for critters because based on your css it is running into edge cases.

lukeed:
Fixed the plugin ordering 👍 

There _are_ double `link rel=preload` tags, but that shouldn't matter? It still only fires off a single request to `bundle.*.css` & I'm not getting any unused warnings.

screen shot 2018-09-11 at 5 29 40 pm

lukeed:
I think I see what you're talking about. 

The `default` mode append a `<link rel=stylesheet href="bundle.*.css"/>` to the end of the document – is that what you were referring to?

The tag is removed with `js` and `js-lazy`, but has no effect on the Lighthouse score w/ our Preact templates since the CSS is so little.

prateekbh:
>The tag is removed with js and js-lazy, but has no effect on the Lighthouse score w/ our Preact templates since the CSS is so little.

This has affect when the app grows e.g. material.preact.com

preaload: swap works pretty perfect with just a few characters even for bigger sites.

>There are double link rel=preload tags, but that shouldn't matter?

Chrome throws a warning for this in console IIRC

lukeed:
Gotcha. It warns if unused, but not on double preload tag.

Feel free to change it to swap, I don't have any objections! I'll likely be away from desk for a few hours

prateekbh:
How about along with a opt-out flag?

lukeed:
Personally I don't like opt-out flags, so I'd be hesitant. I also not sure why you'd need one?

Arguably, if we're making something a default behavior I feel that we should stand behind it 100©

Unlike the esm teacher, there's no compatibility or support issues.

prateekbh:
Because I feel that it's not very stable, I made like 4-5 changes to the source code in my node_modules before I could get it to work

prateekbh:
@lukeed @developit should this be concerning that even after using preload(swap) lighthouse still complains about unused for the link tag and even the inlined css is being reported as extra css supplied

e.g. https://www.webpagetest.org/lighthouse.php?test=180913_MG_dc5a8260f5c19115230143e77c18527d&run=1

prateekbh:
Also, another use case of opt-out flag/question: if I use a case in July aren't they better at giving me my critical css rather than critters?

lukeed:
If it's not stable, then:

a) we shouldn't include it
b) we should fix it so that it's stable; or
c) _enable_ the feature with a flag, not disable with flag

lukeed:
Oops, sorry, I meant that those were three possible choices; not to do all three 😅 

I'd be totally stoked with inline-css as a default feature, but we need to be confident in it. If we're not, then it either can't be a default behavior or shouldn't be there at all

prateekbh:
+1, does an opt in flag sound like a decent approach now?

ForsakenHarmony:
We coulc do it like the esm stuff

Opt in now and if we solve all the problems we can make it the default

prateekbh:
added a flag, and conditional preload of css if critters is used

packages/cli/lib/index.js
```diff
@@ -31,7 +31,8 @@ prog
 	.option('--analyze', 'Launch interactive Analyzer to inspect production bundle(s)')
 	.option('--prerenderUrls', 'Path to pre-rendered routes config', 'prerender-urls.json')
 	.option('-c, --config', 'Path to custom CLI config', 'preact.config.js')
-	.option('--esm', '[EXPERIMENTAL] Builds ES-2015 bundles for your code.', false)
+	.option('--esm', 'Builds ES-2015 bundles for your code.', false)
+	.option('--critical-css', 'Adds critical cssto the prerendered markup.', true)
```
lukeed: I would make this `--inline` or `--inline-css`, but maybe that's just me /shrug

packages/cli/lib/resources/template.html
```diff
@@ -10,8 +10,9 @@
 		<% if (htmlWebpackPlugin.options.manifest.theme_color) { %>
 			<meta name="theme-color" content="<%= htmlWebpackPlugin.options.manifest.theme_color %>">
 		<% } %>
+		<% const filesRegexp = options.criticalCss ? /\.(js)$/ : /\.(css|js)$/;%>
```
lukeed: I don't understand – we'd still want to preload _some_ CSS, especially the routes' chunks.

packages/cli/lib/index.js
prateekbh: Nah that sounds better

lukeed:
@prateekbh FYI - failing tests are now actually relevant to your changes

prateekbh:
fixed everything, good to merge. plz review

developit:
Critters issue should be fixed in [1.2.0](https://github.com/GoogleChromeLabs/critters/releases/tag/1.2.0).

prateekbh:
@developit I have no idea why critters 1.2.0 complains JSDOM is not a constructor, hence keeping this PR @ 1.1.0 and will upgrade when we fix that

prateekbh:
@ForsakenHarmony can you review this one :)

packages/cli/lib/resources/template.html
```diff
@@ -10,8 +10,9 @@
 		<% if (htmlWebpackPlugin.options.manifest.theme_color) { %>
 			<meta name="theme-color" content="<%= htmlWebpackPlugin.options.manifest.theme_color %>">
 		<% } %>
+		<% const filesRegexp = htmlWebpackPlugin.options.inlineCss ? /\.(chunk\.\w{5}\.css|js)$/ : /\.(css|js)$/;%>
```
reznord: `template.html` is already complicated and messed up. @developit any plans on making it similar to what CRA does?

prateekbh: Well I can take a look at this in next iteration, but not for this PR

PR #674, opened 2018-11-27
Adds brotli support for modern javascript

### LINKED ISSUE #751: [next] Service Worker precaches HTML files, but they aren't used
In CLI@next, the ServiceWorker setup pre-caches all HTML files produced by `prerender-urls.json`. This is great, except we never use those cache entries. They all have `/index.html` appended for our pretty URL handling, so they are cache misses.

Even with that stripped out though, routing around a website built with Preact CLI doesn't do full page navigations such that the prerendered HTML is used.

Instead, it seems like we need to pre-cache the route JS but not the HTML. The only HTML we should precache is `/index.html`, and it should be served up for all requests that are cache misses when offline.

/cc @prateekbh

prateekbh:

Feature

WIP

**Summary**
This PR does the following

- [x] Adds `brotli` support for precaching first party modern javascript, hence reducing the total cost/ total download amount from network to the end user.

- [x] Puts this behind a flag as one needs to add headers to their jS serving infra, before enabling this(see the added console.log)

- [x] Switches the underlying `sw` builder from sw-precache to workbox, for better extendable code.(next PR png -> webP)

- [x] Adds the capability to copy sw.js into user land and add custom logic to it.

- [x] Fixes #751 

We only add brotli support modern js because we can be sure without a test that browser supporting esm supports brotli.

Yes, the existing [preact-cli-sw-precache](https://www.npmjs.com/package/preact-cli-sw-precache) plugins wont work anymore.

But the users will be able to just copy our `sw.js` file into their src root and add custom logic to it.


packages/cli/lib/lib/sw.js
```diff
@@ -0,0 +1,54 @@
...
+      asset.url = asset.url.replace(/.esm.js$/, ".esm.js.br");
+    }
+    return asset;
+  });
+
+  class BrotliRedirectPlugin {
+    // Before saving the response in cache, we need to treat the headers.
+    async cacheWillUpdate({request, response}) {
+      if (/.js.br$/.test(request.url)) {
+        const headers = response.headers;
```
developit: Would this work instead of manually cloning?

```js
response = response.clone();
response.headers.set('content-type', 'application/javascript')
return response;
```

packages/cli/lib/lib/sw.js
```diff
@@ -0,0 +1,54 @@
...
+        });
+      }
+      return response;
+    }
+  }
+  workbox.precaching.addPlugins(new BrotliRedirectPlugin());
+}
+
+const precacheOptions = {};
+if (process.env.ENABLE_BROTLI) {
```
developit: Should this wrap the plugin definition+invocation so that it gets stripped of out non-brotli builds?

packages/cli/lib/lib/webpack/webpack-client-config.js
```diff
@@ -193,26 +175,23 @@ function isProd(config) {
...
-						/polyfills(\..*)?\.js$/,
-						/\.map$/,
-						/push-manifest\.json$/,
-						/.DS_Store/,
-						/\.git/
-					]
-				}),
-			);
-		}
+	if (config['sw']) {
```
developit: Any reason this isn't `config.sw`?

packages/cli/lib/lib/webpack/webpack-sw-config.js
```diff
@@ -0,0 +1,62 @@
...
+    swSrc = resolve(`${src}/sw.js`);
+  }
+
+  return {
+    context: src,
+    entry: {
+      sw: swSrc,
+    },
+    target: 'webworker',
+		output: {
```
developit: spacing (https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig)

packages/cli/lib/lib/webpack/webpack-sw-config.js
```diff
@@ -0,0 +1,62 @@
...
+      'process.env.NODE_ENV': 'production',
+    }),
+  ];
+
+  esm && plugins.push(
+    new BabelEsmPlugin({
+      filename: '[name]-esm.js',
+      beforeStartExecution: (plugins) => {
+        plugins.forEach(plugin => {
+          if (plugin.constructor.name === 'DefinePlugin' && plugin.definitions) {
```
developit: What's the reason for only mutating values in-place here - does this fix `DefinePlugin` definitions leaking into the parent compiler?

Otherwise this could just be

```js
plugins.forEach(plugin => {
  if (plugin.constructor.name === 'DefinePlugin') {
    if (!plugin.definitions) throw Error('ESM Error:  DefinePlugin found without definitions.');
    plugin.definitions['process.env.ES_BUILD'] = 'true';
  }
});
```

packages/cli/lib/lib/webpack/webpack-sw-config.js
```diff
@@ -0,0 +1,62 @@
...
+            throw new Error('WebpackDefinePlugin found but not `process.env.ES_BUILD`.');
+          }
+        });
+      }
+    })
+  );
+
+  let swSrc = resolve(__dirname, './../sw.js');
+  if (fs.existsSync(resolve(`${src}/sw.js`))) {
+    console.log(yellow('⚛️ Info - CLI found sw.js in source root folder, using that to compile the final service worker instead'));
```
developit: Nit: can we make this a different color like blue? yellow seems like a warning, this is informational.

Also maybe shorten the message since it's a heads-up that will be shown every build? Or only show the message when things change (first build or sw.js added/removed):

```js
  // only warn on first build, OR on rebuild if sw.js is added/removed:
  let swSrc = resolve(__dirname, './../sw.js');
  const exists = fs.existsSync(resolve(`${src}/sw.js`));
  if (exists !== global.hasSeenSw) {
    global.hasSeenSw = exists;
    if (exists) {
      console.log(blue('⚛️ Detected custom sw.js: compiling instead of default Service Worker.'));
    }
    else {
      console.log(blue('⚛️ No custom sw.js detected: compiling default Service Worker.'));
    }
  }
```

packages/cli/lib/lib/webpack/webpack-client-config.js
```diff
@@ -273,9 +260,13 @@ function isDev(config) {
...
 module.exports = function (env) {
-	return merge(
+	const config = [merge(
 		baseConfig(env),
 		clientConfig(env),
 		(env.isProd ? isProd : isDev)(env)
-	);
-};
+	)];
+	if (env.sw) {
```
developit: Is the Array return here done with the intention of moving to a MultiCompiler?  This seems like it'll break in future Workbox versions that use child compilation.

packages/cli/lib/lib/sw.js
lukeed: This should – it's what my service workers are doing, tho I just pray that they work

packages/cli/lib/lib/webpack/webpack-client-config.js
prateekbh: How is the child compiler affected by the its parent compilation mode?

packages/cli/lib/lib/webpack/webpack-client-config.js
prateekbh: nups, fixed

packages/cli/lib/lib/webpack/webpack-sw-config.js
prateekbh: good point, i can chk that

packages/cli/lib/lib/sw.js
prateekbh: It kinda is already being stripped out

With `--brotli`
screen shot 2018-12-09 at 8 34 49 pm


Without --brotli
screen shot 2018-12-09 at 8 36 09 pm

packages/cli/lib/lib/webpack/webpack-sw-config.js
prateekbh: done.. i guess?

packages/cli/lib/lib/webpack/webpack-sw-config.js
prateekbh: works, fixed

packages/cli/lib/lib/webpack/webpack-sw-config.js
prateekbh: done

prateekbh:
@developit I re-did this with a child compiler, please review

packages/cli/lib/lib/sw.js
developit: my bad - this would work though:

```js
const headers = new Headers(clonedResponse.headers);
newHeaders.set('content-type', 'application/javascript');
const newResponse = new Response(await clonedResponse.text(), { headers })
```

prateekbh: fixed

packages/cli/lib/lib/webpack/sw-plugin.js
```diff
@@ -0,0 +1,113 @@
...
+					this.constructor.name
+				);
+
+				childCompiler.context = compiler.context;
+				childCompiler.options = Object.assign({}, outputOptions);
+				childCompiler.options.entry = {
+					sw: swSrc,
+				};
+				childCompiler.options.target = 'webworker';
+				childCompiler.options.output = JSON.parse(
```
prateekbh: @developit @ForsakenHarmony do we have any dep already which can help me do this?

prateekbh: nvm fixed it

packages/cli/lib/commands/build.js
```diff
@@ -22,6 +23,15 @@ module.exports = async function(src, argv) {
 		);
 	}
 
+	if (argv.brotli) {
+		console.log(
```
ForsakenHarmony: guess this should use the util log fn

packages/cli/lib/lib/webpack/sw-plugin.js
```diff
@@ -0,0 +1,115 @@
...
+		const exists = fs.existsSync(resolve(`${this.src_}/sw.js`));
+		if (exists) {
+			if (exists) {
+				console.log(
+					blue(
+						'⚛️ Detected custom sw.js: compiling instead of default Service Worker.'
+					)
+				);
+			} else {
+				console.log(
```
ForsakenHarmony: same thing here?

prateekbh:
- Will fix these 
- Merge this
- Release RC
- Write tests

prateekbh:
Fixed comments!

PR #678, opened 2018-12-07
remove script-ext-html-webpack-plugin

yaodingyd:

Refactoring

No

**Summary**

After this fix https://github.com/developit/preact-cli/pull/127 script tag is added by template instead of using `script-ext-html-webpack-plugin`. Also the `defaultAttribute` is not used as it is hardcoded in template.

I propose adding a new option in `html-webpack-plugin` called `scriptLoading` and it is configurable.

NO

### DIFF
--- a/package.json
+++ b/package.json
@@ -146,7 +146,6 @@
     "raw-loader": "^0.5.1",
     "require-relative": "^0.8.7",
     "rimraf": "^2.6.1",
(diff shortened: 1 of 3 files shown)


ForsakenHarmony:
Could you do this for the next branch?

yaodingyd:
@ForsakenHarmony it's under #681 now. thanks!

Issue #679, opened 2018-12-08
MiniCssExtractPlugin uses '[id]' rather than '[name]' for 'chunkFilename'


johnhaitas:
**Do you want to request a *feature* or report a *bug*?**
bug
**What is the current behaviour?**
`next` branch names css chunk files with `[id]` rather than `[name]`

**If the current behaviour is a bug, please provide the steps to reproduce.**
production build using `preact build` using `preact-cli@next`

**What is the expected behaviour?**
files should be named according to their chunk name rather than with a number.

**If this is a feature request, what is motivation or use case for changing the behaviour?**

**Please mention other relevant information.**
- Node version
- npm version
- Operating system
- CLI version: 3.0.0-next.14
- Browser

johnhaitas:
My short-term local workaround is to add the following to my preact config transform
```javascript
helpers.getPluginsByName(config, 'MiniCssExtractPlugin')
	.forEach(({ plugin }) => plugin.options.chunkFilename = plugin.options.chunkFilename.replace('[id]', '[name]'));
```

reznord:
Yep, cuz it is already merged in `next`

johnhaitas:
Thanks all

PR #680, opened 2018-12-08
Addresses issue #679 by replacing `[id]` with `[name]` for the `chunkFilename` in `MiniCssExtractPlugin`

johnhaitas:
bugfix - adjusts how css chunk files are named

No

**Summary**
Per issue #679 - the `MiniCssExtractPlugin` names files numerically with `[id]` rather than using `[name]`

It shouldn't

**Other information**

`preact-cli@3.0.0-next.14`

### DIFF
--- a/packages/cli/lib/lib/webpack/webpack-base-config.js
+++ b/packages/cli/lib/lib/webpack/webpack-base-config.js
@@ -203,7 +203,7 @@ module.exports = function (env) {
 			// Extract CSS
 			new MiniCssExtractPlugin({
 				filename: isProd ? "[name].[contenthash:5].css" : "[name].css",
-				chunkFilename: isProd ? "[id].chunk.[contenthash:5].css" : "[id].chunk.css"
+				chunkFilename: isProd ? "[name].chunk.[contenthash:5].css" : "[name].chunk.css"
 			}),
 			new ProgressBarPlugin({
 				format: '\u001b[90m\u001b[44mBuild\u001b[49m\u001b[39m [:bar] \u001b[32m\u001b[1m:percent\u001b[22m\u001b[39m (:elapseds) \u001b[2m:msg\u001b[22m',


packages/cli/tests/images/build.js
```diff
@@ -91,9 +91,8 @@ exports.preload.head = `
 	<link rel=\\"manifest\\" href=\\"\\/manifest\\.json\\">
 	<link rel=\\"preload\\" href=\\"\\/bundle\\.\\w{5}\\.js\\" as=\\"script\\">
 	<link rel=\\"preload\\" href=\\"\\/route-home\\.chunk\\.\\w{5}\\.js\\" as=\\"script\\">
-	<link rel=\\"preload\\" href=\\"\\/2\\.chunk\\.\\w{5}\\.css\\" as=\\"style\\">
 	<link rel=\\"preload\\" href=\\"\\/route-home\\~route-route66\\~route-route89\\.chunk\\.\\w{5}\\.js\\" as=\\"script\\">
-	<link rel=\\"preload\\" href=\\"\\/1\\.chunk\\.\\w{5}\\.css\\" as=\\"style\\">
+	<link rel=\\"preload\\" href=\\"\\/route-home\\.chunk\\.\\w{5}\\.css\\" as=\\"style\\">
```
johnhaitas: I am concerned that one of the css chunk preloads needs to be removed rather than adjusted here. I expected there to be a line that looks like
```javascript
<link rel=\\"preload\\" href=\\"\\/route-home\\~route-route66\\~route-route89\\.chunk\\.\\w{5}\\.css\\" as=\\"style\\">
```

prateekbh: i don't understand the `~route-route66\\~route-route89` part?
so for this page, the home css and the route css should be loaded... why multiple route name in the same tag?

johnhaitas: perhaps i misunderstand the question. isn't that some new code splitting introduced on the `next` branch? in my local project the name represents the routes that share the code - all those names are joined by `~`.

developit: hmm - this is strange. Unless contenthash is producing slashes, it seems like the URL format here should still just be `/xx.chunk.12345.css`:

https://github.com/developit/preact-cli/blob/next/packages/cli/lib/lib/webpack/webpack-base-config.js#L203-L207

prateekbh: the code splitting happns only under some conditions and i see that 30kb condition wont match here.. so the above code looks fine

johnhaitas: so we can leave as is?

prateekbh: I Guess so

PR #691, opened 2019-01-08
[SW] Don't invoke skipWaiting

developit:
This fixes the issue where lazy-loaded splitted chunks fail because the SW has been updated and the previous URLs are now 404's.

Docs: https://github.com/GoogleChromeLabs/sw-precache#skipwaiting-boolean


ForsakenHarmony:
probably want this for next as well?

Issue #692, opened 2019-01-09
Allow prerenderUrls to be a javascript file


StephanBijzitter:
**Do you want to request a *feature* or report a *bug*?**
I want to request an improvement to an existing feature.

**What is the current behaviour?**
prerenderUrls currently requires a JSON file. Any javascript file that has a default export of an object or a JSON string is ignored silently.

**If the current behaviour is a bug, please provide the steps to reproduce.**
Not a bug, but take this `prerender-urls.js` file:
```js
// perhaps get more names from a database or a glob pattern
const profiles = ['john', 'jane'].map((name) => ({
	url: `/profile/${name}`
}));

module.exports = [
	{
		url: '/',
		title: 'Homepage'
	}, {
		url: '/profile'
	},
	...profiles
];
```
**What is the expected behaviour?**
The file is executed and the resulting object is used.

**If this is a feature request, what is motivation or use case for changing the behaviour?**
Adding a new route to my website currently requires me to also update the JSON file, which is essentially a duplication of my router. Adding a `prebuild` script to my `package.json` allows me to generate a dynamic JSON file, but this is a workaround.

**Please mention other relevant information.**
- Node version: 10.10.0
- Yarn version: 1.13.0
- Operating system: OS X Mojave
- CLI version: 2.2.1
- Browser: Opera

PR #693, opened 2019-01-09
Allow prerenderUrls to be a javascript file

StephanBijzitter:
An improvement to an existing feature for the build steps.

No, I could not find any tests related to the `prerenderUrls` option so I'm not sure how you want to handle this.

**Summary**
See #692 for more details, but in short this change allows for greater flexibility and more dynamic route building.

No.

**List of supported formats for prerenderUrls**
* JSON file (unchanged)

* JS file with `module.exports = [{url: '/'}, ...]`
* JS file with `export default [{url: '/'}, ...]`

* JS file with `module.exports = () => [{url: '/'}, ...]`
* JS file with `export default () => [{url: '/'}, ...]`

* JS file with `module.exports = '"[{"url":"/"}, ...]"'`
* JS file with `export default '"[{"url":"/"}, ...]"'`

* JS file with `module.exports = () => '"[{"url":"/"}, ...]"'`
* JS file with `export default () => '"[{"url":"/"}, ...]"'`

**Promises and async functions**
These take a bit more to support, so I kept it small for this pull request. Support can always be improved later.

### DIFF
--- a/packages/cli/lib/lib/webpack/render-html-plugin.js
+++ b/packages/cli/lib/lib/webpack/render-html-plugin.js
@@ -72,7 +72,7 @@ module.exports = function(config) {
 				pages = result;
 			}
 		} catch (error) {
-			warn(error.message);
+			warn(`Failed to load prerenderUrls file, using default!\n${error.stack}`);
 		}
 	}
 
--- a/packages/cli/lib/lib/webpack/render-html-plugin.js
+++ b/packages/cli/lib/lib/webpack/render-html-plugin.js
@@ -6,6 +6,7 @@ const HtmlWebpackPlugin = require('html-webpack-plugin');
 const prerender = require('./prerender');
 const createLoadManifest = require('./create-load-manifest');
 const template = resolve(__dirname, '../../resources/template.html');
+const { warn } = require('../../util');
 
 module.exports = function(config) {
 	const { cwd, dest, isProd, src } = config;
@@ -71,7 +72,7 @@ module.exports = function(config) {
 				pages = result;
 			}
 		} catch (error) {
-			console.error(error);
+			warn(error.message);
 		}
 	}
 
--- a/packages/cli/lib/lib/webpack/render-html-plugin.js
+++ b/packages/cli/lib/lib/webpack/render-html-plugin.js
@@ -54,9 +54,21 @@ module.exports = function(config) {
 		});
 	};
 
-	const pages = readJson(resolve(cwd, config.prerenderUrls || '')) || [
-		{ url: '/' },
-	];
+	let pages;
+	const {prerenderUrls} = config;
+
+	if (!prerenderUrls) {
+		pages = [{ url: '/' }];
+	} else if (/\.json$/i.test(prerenderUrls)) {
+		pages = readJson(resolve(cwd, prerenderUrls));
+	} else if (/\.js$/i.test(prerenderUrls)) {
+		let result = require(resolve(cwd, prerenderUrls));
+		result = result.default ? result.default : result;
+		if (typeof result === 'function') {
+			result = result();
+		}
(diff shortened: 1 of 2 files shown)


ForsakenHarmony:
can you do this for the next branch?

developit:
Seconded! I would definitely use this.
Also - anything we can do to allow the exported function to be async here?

StephanBijzitter:
@ForsakenHarmony 
>can you do this for the next branch?

Yeah, no problem :-)

@developit
>Seconded! I would definitely use this.
>Also - anything we can do to allow the exported function to be async here?

That is my intention, but I'd have to dig deeper into the code and probably require a bit of a refactoring. You'll see a second pull request once that's ready :-)

StephanBijzitter:
Just changed the base branch to `next`, will rebase sometime today.

ForsakenHarmony:
you have another error though with the array being undefined, not sure if that's what you're talking about

If it is, we'd need a better error there

StephanBijzitter:
Yeah that's the result of specifying a JSON file that doesn't exist. I agree, that should give a much clearer error.

ForsakenHarmony:
there's logging helpers in `utils.js`

ForsakenHarmony:
https://github.com/developit/preact-cli/blob/next/packages/cli/lib/util.js

ForsakenHarmony:
Sorry it's `util.js`

prateekbh:
@StephanBijzitter we merged a PR which should fix some false breaking tests, can you please rebase this branch?

StephanBijzitter:
My pleasure!

packages/cli/lib/lib/webpack/render-html-plugin.js
```diff
@@ -54,9 +54,27 @@ module.exports = function(config) {
...
+				result = result();
+			}
+			if (typeof result === 'string') {
+				result = JSON.parse(result);
+			}
+			if (result instanceof Array) {
+				pages = result;
+			}
+		} catch (error) {
+			warn(error.message);
```
ForsakenHarmony: A custom error message along with it might be nice, not sure if the returned message will be obvious to users

StephanBijzitter: Yeah but that should be additional in that case. It'd be a shame to lose the extra information provided by the error message. In case of a syntax error it's easy to see what the syntax error was and in case of an incorrect path you can see the path that has been tried, which may have been resolved differently than a user would have expected.

Maybe: `Failed to load prerenderUrls file: ${error.message}`?

ForsakenHarmony: yeah, that's what I meant

maybe even `error.stack`

StephanBijzitter: I'll play around a bit with different error messages when I get home.

Having vrijmibo right now :-)
<sub><sup>*Google it, should be sfw</sup></sub>

StephanBijzitter: Changed it to:
screenshot 2019-01-21 at 10 39 30

prateekbh:
@ForsakenHarmony @developit do we want to change this PR as per Jason's latest comment or merge this and do that later or in 4.0?

ForsakenHarmony:
if @StephanBijzitter wants to do it then sure, otherwise I suppose we should finally release this

StephanBijzitter:
Would love to do it... in a different pull request? 😇

prateekbh:
This PR was merged with broken tests: https://travis-ci.org/developit/preact-cli/builds/482330463

Plz fix

prateekbh:
Also, our tests dont fail if they break. 😢

ForsakenHarmony:
that's just a warning

prateekbh:
oh so that's expected!!! looks ugly TBH

PR #695, opened 2019-01-12
Fix the markup for the broken tests

reznord:

Fixes the broken tests in the `next` branch.

Yes, it goes and fixes the broken tests 😛

**Summary**
The markup for the build tests is messed up and because of that the tests for the `next` branch are broken.

Naaaah

**Other information**
Node version: v10.15.0
npm version: 6.4.1
CLI Version: next
Operating System: macOS Mojave v.10.14.2

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "private": true,
   "scripts": {
     "pretest": "lerna bootstrap",
+    "clean": "lerna clean --yes",
     "test": "eslint packages && lerna run test"
   },
   "devDependencies": {
--- a/packages/cli/tests/images/build.js
+++ b/packages/cli/tests/images/build.js
@@ -55,9 +55,9 @@ exports.prerender.heads.home = `
 	<link rel="preload" href="\\/route-home\\.chunk\\.\\w{5}\\.js" as="script">
 	<link rel="preload" href="\\/route-home\\.chunk\\.\\w{5}\\.css" as="style">
 	<link rel="shortcut icon" href="\\/favicon\\.ico">
+	<style>html{padding:0}<\\/style>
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -2,6 +2,7 @@
   "private": true,
   "scripts": {
     "pretest": "lerna bootstrap",
+    "clean": "lerna clean --yes",
```
reznord: adds a clean command, which will remove all the `node_modules` of the packages.

packages/cli/tests/images/build.js
```diff
@@ -73,9 +73,7 @@ exports.prerender.heads.route66 = `
 	<link rel="preload" href="\\/route-route66\\.chunk\\.\\w{5}\\.js" as="script">
 	<link rel="preload" href="\\/route-route66\\.chunk\\.\\w{5}\\.css" as="style">
 	<link rel="shortcut icon" href="\\/favicon\\.ico">
-	<link href=\\"/bundle.\\w{5}.css\\" rel=\\"preload\\" as=\\"style\\" onload=\\"this.rel='stylesheet'\\"><noscript>
-		<link rel=\\"stylesheet\\" href=\\"\\/bundle.\\w{5}.css\\"><\\/noscript>
-	<style>html{padding:0;}<\\/style>
+	<link href=\\"\\/bundle.\\w{5}.css\\" rel=\\"stylesheet\\">
```
prateekbh: Where did the noscript tag go?
Isn't it needed with critters

reznord: I guess not. The `noscript` tag is the reason why the tests were failing. Either ways, closed this PR in favor of #696 (which covers these changes as well)

PR #696, opened 2019-01-12
Yarn Workspaces & Prettier

ForsakenHarmony:
waiting for https://github.com/prateekbh/babel-esm-plugin/pull/9


ForsakenHarmony:
well I guess the fix is mostly for windows

packages/cli/lib/lib/webpack/webpack-base-config.js
```diff
@@ -22,6 +23,24 @@ function resolveDep(dep, cwd) {
 	return dep;
 }
 
+function findAllNodeModules(startDir) {
```
ForsakenHarmony: this is necessary, because it would otherwise not resolve `node_modules` above the current folder, which is something I ran into when trying to use `preact-cli` in a project of mine with workspaces

packages/cli/tests/images/build.js
```diff
@@ -55,10 +55,12 @@ exports.prerender.heads.home = `
 	<link rel="preload" href="\\/route-home\\.chunk\\.\\w{5}\\.js" as="script">
 	<link rel="preload" href="\\/route-home\\.chunk\\.\\w{5}\\.css" as="style">
 	<link rel="shortcut icon" href="\\/favicon\\.ico">
-	<link href=\\"/bundle.\\w{5}.css\\" rel=\\"preload\\" as=\\"style\\" onload=\\"this.rel='stylesheet'\\"><noscript>
-		<link rel=\\"stylesheet\\" href="\\/bundle.\\w{5}.css"><\\/noscript>
-	<style>html{padding:0;}<\\/style>
-<\\/head>
+	<style>html{padding:0}</style>
```
reznord: whoops! I have fixed the same test images in #695 😅

reznord:
@ForsakenHarmony closing https://github.com/developit/preact-cli/pull/695 since the same changes are done here.

.prettierrc.json
```diff
@@ -0,0 +1,8 @@
+{
+	"trailingComma": "es5",
```
reznord: Do we really need this? (Not a huge fan of this cuz it is annoying at times) :P

packages/cli/tests/images/build.js
ForsakenHarmony: Oof

.prettierrc.json
ForsakenHarmony: Well, prettier will do it for you and it's better for diffs

packages/cli/tests/images/build.js
```diff
@@ -73,10 +75,8 @@ exports.prerender.heads.route66 = `
 	<link rel="preload" href="\\/route-route66\\.chunk\\.\\w{5}\\.js" as="script">
 	<link rel="preload" href="\\/route-route66\\.chunk\\.\\w{5}\\.css" as="style">
 	<link rel="shortcut icon" href="\\/favicon\\.ico">
-	<link href=\\"/bundle.\\w{5}.css\\" rel=\\"preload\\" as=\\"style\\" onload=\\"this.rel='stylesheet'\\"><noscript>
-		<link rel=\\"stylesheet\\" href=\\"\\/bundle.\\w{5}.css\\"><\\/noscript>
-	<style>html{padding:0;}<\\/style>
-<\\/head>
+	<link href=\\"/bundle.\\w{5}.css\\" rel="stylesheet">
```
prateekbh: Why is the noscript tag removed here?
Critters should be adding it here as well right?

reznord: Maybe the noscdipt tag is removed in critters2

reznord: The tests started failing on next branch after we updated critters version

ForsakenHarmony: maybe critters deemed it small enough? not sure

prateekbh: @developit.. please comment

prateekbh:
@ForsakenHarmony can you bump the esm-plugin version?

packages/cli/tests/images/build.js
prateekbh: also this does not have `rel=\\"preload\\" as=\\"style\\" onload=\\"this.rel='stylesheet'\\"` which makes it a blocking request

prateekbh: ! this is a valid bug.. the original content should be as is

developit: Critters does inline stylesheets below a threshold, but we don't have that enabled in configuration (https://github.com/GoogleChromeLabs/critters/blob/master/src/index.js#L75-L76). Seems like it's erroring out?

ForsakenHarmony:
I'll revert when the bug in critters is fixed and we can test against that

prateekbh:
@ForsakenHarmony or we can revert to Critters 1.x.x till then.. I'll defer the decision to @developit

ForsakenHarmony:
I'll do that

packages/async-loader/async.js
```diff
@@ -15,15 +15,20 @@ export default function(req) {
 		this.shouldComponentUpdate = (_, nxt) => {
 			nxt = nxt.child === void 0;
 			if (nxt && old === void 0 && b) {
-				old = b.nodeType===3 ? b.data : h(b.nodeName, { dangerouslySetInnerHTML: { __html: b.innerHTML } });
+				old =
+					b.nodeType === 3
+						? b.data
+						: h(b.nodeName, {
```
developit: ^ This is one of the things I dislike about Prettier. 
I don't think anyone would argue this is easier to read :P

developit:
@ForsakenHarmony do you think it's worth double checking that we're going to end up with more than 2 packages?  TBH I have been feeling like it'd be worth moving async-component / async-component-loader out into their own package(s).  If we were to do that, what packages would we have here?  Maybe CLI and core?

packages/async-loader/async.js
prateekbh: isnt there any setting to turn this off?

PR #701, opened 2019-01-20
Babel macros

ForsakenHarmony:


developit:
Should we have a demo for this one?

ForsakenHarmony:
maybe, wanted to try using them on my gql lib to do some stuff at compile time 🙃

developit:
SGTM. I have some tiny fears around magic syntax but at least this is a pretty common approach.

PR #702, opened 2019-01-20
fix unhandledRejection handler

ForsakenHarmony:


prateekbh:
what was this fix?

ForsakenHarmony:
I added that handler with the workspace pr and just realized I also broke it in that pull request

PR #709, opened 2019-01-25
fix prerender error

ForsakenHarmony:


ForsakenHarmony:
I don't think it was really necessary given that we exit right after, but probably best to do it properly either way

PR #711, opened 2019-01-25
Fix deprecations

ForsakenHarmony:
just missing babel-esm


lukeed:
there's `webpack-plugin-replace@1.2.0` too

ForsakenHarmony:
second commit?

ForsakenHarmony:
🤔

PR #713, opened 2019-01-25
remove script-ext-html-webpack-plugin and add script loading option

ForsakenHarmony:
Accidentally force pushed #681, should look up what commands do before I run them


ForsakenHarmony:
just gonna merge this, because jason approved the old one

PR #714, opened 2019-01-25
remove `swap` because it's broken in firefox

### LINKED ISSUE #710: CSS preload causes Firefox to fail loading styles
Bug
CSS fails to load in Firefox when using `rel="preload"` for CSS
1. Create a static build using preact-cli that has a external stylesheet
2. Serve the built content
3. Navigate to page with Firefox
4. Observe how the non in-lined CSS fails to load on the page

Styles should load fine in every browser with the defaults shipped by the browser devs.

Currently Firefox only has experimental support for the `<link rel="preload">` attribute, which is locked behind the `network.preload` flag in `about:config`. In order for preact-cli's async CSS loading to function in Firefox, you have to manually enable the flag.

It would probably be best to remove this so users don't have to go dig into browser configs to get stylesheets to load.

- CLI version: 3.0.0-next.14
- Browser: Firefox Nightly 66.0a1 (2019-01-25) 

With flag enabled: 
image

Default Firefox (Flag at default value of false):
image

ForsakenHarmony:
closes #710

### DIFF
--- a/packages/cli/lib/lib/webpack/webpack-client-config.js
+++ b/packages/cli/lib/lib/webpack/webpack-client-config.js
@@ -241,9 +241,7 @@ function isProd(config) {
 
 	if (config['inline-css']) {
 		prodConfig.plugins.push(
-			new CrittersPlugin({
-				preload: 'swap',
-			})
+			new CrittersPlugin()
 		);
 	}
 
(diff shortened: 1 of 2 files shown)


prateekbh:
sad to see this go.... also this might affect our lighthouse score may be.

prateekbh:
can you check what affect does "js-lazy" have https://github.com/GoogleChromeLabs/critters#preloadstrategy

ForsakenHarmony:
I think that's the same thing, but not inline, I'd say we look at other options later

PR #715, opened 2019-01-25
re-enable prettier flag

prateekbh:
Bugfix: Re-enables preload flag which got removed during the esm preload PR

**Did you add tests for your PR?**
WIP

No


ForsakenHarmony:
kinda wondering if we should have a test for this

it's auto enabled from here https://github.com/developit/preact-cli/blob/next/packages/cli/tests/lib/cli.js#L14

prateekbh:
yep switching it off... and then will add tests

ForsakenHarmony:
am I missing the one where it's enabled? 👀

PR #729, opened 2019-02-12
Update / clean up deps

ForsakenHarmony:


packages/cli/lib/fs.js
```diff
@@ -0,0 +1,10 @@
+const { readFile, writeFile, copyFile, stat, statSync } = require('fs');
+const { promisify } = require('util');
+
+module.exports = {
+	copyFile: promisify(copyFile),
+	readFile: promisify(readFile),
+	writeFile: promisify(writeFile),
+	stat: promisify(stat),
+	statSync,
+};
```
ForsakenHarmony: was kinda annoyed by the abandoned dependency and we'll get native promise support 🔜™

ForsakenHarmony:
This is now waiting for https://github.com/terkelg/tiny-glob/issues/35

lukeed:
Node 6 is still intentionally supported. Can't merge this without major bump.

lukeed:
_Removed my comment...I was distracted, sorry. Thought I was looking at a PR for an entirely different repo_ 🙈

reznord:
Need to change the lerna version in `lerna.json` as well.

packages/cli/package.json
```diff
@@ -58,67 +64,57 @@
...
     "html-webpack-plugin": "^3.2.0",
-    "inquirer": "^5.2.0",
+    "inquirer": "^6.2.2",
     "ip": "^1.1.5",
     "isomorphic-unfetch": "^3.0.0",
+    "kleur": "^3.0.1",
     "loader-utils": "^1.1.0",
-    "log-symbols": "^2.1.0",
-    "mini-css-extract-plugin": "^0.4.0",
+    "loglevelnext": "^3.0.0",
```
lukeed: Where does this get used?

packages/cli/lib/commands/create.js
```diff
@@ -1,12 +1,10 @@
...
-const glob = require('glob');
+const glob = require('tiny-glob');
 const gittar = require('gittar');
-const fs = require('fs.promised');
-const copy = require('ncp');
-const { green } = require('chalk');
+const fs = require('../fs');
+const { green } = require('kleur');
 const { resolve, join } = require('path');
 const { prompt } = require('inquirer');
```
lukeed: I would vote for `prompts` here – waaay smaller but up to you

packages/cli/package.json
ForsakenHarmony: oh, accidentally added it, will remove it again

packages/cli/lib/commands/create.js
ForsakenHarmony: I don't mind

reznord:
Seems like https://github.com/terkelg/tiny-glob/pull/41 is fixed, so this sounds good to me

reznord:
On the other hand, should we do a beta with ceviche?

lukeed:
@reznord It's not fixed. I merged it to continue fixing it in another PR. One of its sub-dependencies was/is broken.

PR #730, opened 2019-02-13
Add support for webp images

developit:
Adds support for webp images.


prateekbh:
Just a note: after #674, we'll should be auto generating webP and redirecting jpg to webP auto matically

Issue #733, opened 2019-02-15
ESM Build Causes error


hassanbazzi:
Getting a bug when esm build is used. Opting out using the esm flag fixes it.

Error:

Uncaught TypeError: Cannot read property 'webpackJsonp' of undefined
    at bootstrap:230
    at bootstrap:239

This is using the latest version of preact-cli.

prateekbh:
I am still trying to debug this... Some funky code splitting has happened which has caused this...

prateekbh:
preact-www has this line
https://github.com/developit/preact-www/blob/next-cli/preact.config.js#L15

which makes the global object to `this` and `this` is undefined in `<script type=module` thus the undefined error.

Please explain why we need this line.

PR #734, opened 2019-02-18
Show child compilation errors

developit:
This will probably help show the error from #733. For me, it was that I didn't have one of my deps installed.

### DIFF
--- a/packages/cli/lib/commands/create.js
+++ b/packages/cli/lib/commands/create.js
@@ -12,7 +12,7 @@ const { info, isDir, hasCommand, error, trim, warn } = require('../util');
 const { addScripts, install, initGit, isMissing } = require('../lib/setup');
 
 const ORG = 'preactjs-templates';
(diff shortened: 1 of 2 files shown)


prateekbh:
Nice...

ForsakenHarmony:
lol why am I in that commit? 👀

hassanbazzi:
One does not choose when they are summoned, but one must be ready, and you were.

psabharwal123:
@developit in code below forEach passes index as 2nd param to callback warn, warn's second param is exit code so build will fail if there is more than 1 warning which i think it should not. Also same issue in forEach(error), few lines above

```
allFields(stats, 'warnings')
			.map(stripLoaderPrefix)
			.forEach(warn);
```
should be 
```
allFields(stats, 'warnings')
			.map(stripLoaderPrefix)
			.forEach(msg => warn(msg));
```

ForsakenHarmony:
ja🅱️ascript

PR #738, opened 2019-02-21
Upgrade to Critters 2

developit:
Update to Critters 2.3.0 and configure it with optimal settings.

The `pruneSource` thing basically disables the removal of CSS Rules from lazy-loaded stylesheets for multipage sites, because doing so relies on the inlined styles being the same on all pages (which is almost never the case).


prateekbh:
Broken tests?

reznord:
We need to modify the test images as well to support `critters2`

reznord:
Can run the travis tests on this again after this gets merged https://github.com/developit/preact-cli/pull/739

reznord:
Closing this one since critters is already on latest version.

PR #739, opened 2019-02-21
fix(tests): Modify build test images for supporting critter v2

reznord:
Fixes build images

Yusss

**Summary**
Modify the existing build images for fixing supporting critters-2 #738 

Nupes

### DIFF
--- a/packages/cli/tests/images/build.js
+++ b/packages/cli/tests/images/build.js
@@ -52,8 +52,9 @@ exports.prerender.heads.home = `
 	<meta name="apple-mobile-web-app-capable" content="yes">
 	<link rel="manifest" href="\\/manifest\\.json">
 	<link rel="shortcut icon" href="\\/favicon\\.ico">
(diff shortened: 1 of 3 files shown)


packages/cli/tests/images/build.js
```diff
@@ -66,8 +67,7 @@ exports.prerender.heads.route66 = `
 	<meta name="apple-mobile-web-app-capable" content="yes">
 	<link rel="manifest" href="\\/manifest\\.json">
 	<link rel="shortcut icon" href="\\/favicon\\.ico">
-	<link href=\\"/bundle.\\w{5}.css\\" rel=\\"preload\\" as=\\"style\\">
-	<style>html{padding:0;}<\\/style>
```
developit: I think this might be a bug, need to check.

prateekbh: Yes this is the bug we need to fix in critters. With V2 it stops adding inline does to other routes than index

reznord:
This is no longer valid, closing this one.

Issue #736, opened 2019-02-19
prerender file that returns a promise and passes props to route


wulfmann:
**Do you want to request a *feature* or report a *bug*?**

Feature

**What is the current behaviour?**

Hardcoded values

**If this is a feature request, what is motivation or use case for changing the behaviour?**

I would like to be able to return a promise / route list (like is currently supported) to preact build and have arbitrary props passed to the route.

This would allow me to fetch data from an api to allow for a fully rendered build without js (potentially).

--------------

Api idea:

Example Gist:
https://gist.github.com/wulfmann/258bc8189ea246ba1e18bdc5797b2c98

Is this something that has been considered or is already being worked on? I would be willing to work on a solution, just wanted to hear your thoughts! @developit

wulfmann:
Some similar ideas in other projects:

Next:
https://nextjs.org/learn/excel/static-html-export/exporting-other-pages

Nuxt:
https://nuxtjs.org/api/configuration-generate/#speeding-up-dynamic-route-generation-with-code-payload-code-

Gatsby:
https://www.gatsbyjs.org/tutorial/part-seven/

wulfmann:
I think preact (and preact-cli) are especially handy for small projects (a few pages) and in cases where data isn't updated very often, i'd love to utilize it more of as a generator.

I imagine you could do something like this with webpack but I wanted to see if you would be interested in extending the prerender feature.

prateekbh:
Yes we definitely want this.
This was also once suggested by @developit

The main idea is to enable `prerender-urls.js` to have a default export of a function which could return a promise.

This promise can then resolve with a json value that we expect today.

That should be cover the abovesaid requirements, right?

wulfmann:
@prateekbh That solves the first half of what I was suggesting. Is there currently a way to pass props to the route via the json object? I wasn't able to get that to do anything when I tried it.

prateekbh:
See #430

prateekbh:
// @lukeed

wulfmann:
@prateekbh If you look at the gist i linked, you will see the usage I had in my head.

I was more thinking of passing actual props to the component being rendered itself, rather than interpolating the value into the template.

This obviously just may not be possible, but it would be pretty cool.

prateekbh:
@wulfmann the gist exactly aligns with what we expect the shape of API

function():<Promise<Array<Object{url, title,.....}>>>

wulfmann:
[...]

PR #740, opened 2019-02-21
Allow attributes from prerender file to be passed through

wulfmann:

Feature

Yes

**Summary**

Allows arbitrary props and attributes to pass down into the components.

#736 

No

**Other information**


packages/cli/lib/lib/webpack/render-html-plugin.js
```diff
@@ -50,8 +50,8 @@ module.exports = function(config) {
 			config,
 			url,
 			ssr(params) {
-				Object.assign(params, { url });
-				return config.prerender ? prerender({ cwd, dest, src }, params) : '';
+				Object.assign(props, params, { url });
+				return config.prerender ? prerender({ cwd, dest, src }, props) : '';
```
ForsakenHarmony: https://github.com/developit/preact-cli/blob/next/packages/cli/lib/resources/template.html#L23

we could possibly just remove `params` and pass the whole `values` object from above as props here

wulfmann: @ForsakenHarmony I did have that at first actually, but wasn’t sure if you’d want the entire configuration passed in. I don’t mind either way.

ForsakenHarmony: it would potentially take that away yes, but I think the `prerender-urls` file is more flexible in general

wulfmann: I updated it here, but wasn't sure if you wanted me to remove params altogether, or if you wanted me to merge those objects also.

https://github.com/developit/preact-cli/pull/740/files#diff-b26e9c1b1185752601692f2207489852R54

ie:

```Object.assign(values, params, { url })```

File: packages/cli/lib/lib/webpack/render-html-plugin.js [hunk 2]
```diff
@@ -50,8 +50,8 @@ module.exports = function(config) {
 			config,
 			url,
 			ssr(params) {
-				Object.assign(props, params, { url });
-				return config.prerender ? prerender({ cwd, dest, src }, props) : '';
+				Object.assign(params, { url });
```
ForsakenHarmony: well now this line is kinda pointless 👀

wulfmann: The only reason I didn't remove was I just wanted to know if that was your intent.

I just didn't know if you wanted to remove the ability to pass params to ssr entirely or just extras beyond url.

ForsakenHarmony: yeah just remove it

wulfmann: Alright i've made those changes. I also removed the option that gets passed in the template, let me know if you want me to bring that back. Just didn't want to leave dead code there.

packages/cli/tests/subjects/multiple-prerendering/index.js
```diff
@@ -12,9 +13,10 @@ export default class App extends Component {
 	render(props) {
 		return (
 			<div id="app">
-				<Router url={props.url} onChange={this.handleRoute}>
+				<Router url={props.url} onChange={this.handleRoute} {...props}>
 					<Home path="/" />
 					<Route66 path="/route66" />
+					<Custom path="/custom" {...props} />
```
prateekbh: Ohhhhhhhhhhhhh this needs to be documented...
I was figuring out how to do this for days... and then remembered that we already have this

prateekbh:
@ForsakenHarmony @wulfmann I have a question on this, although this generates the values on pre-render but as soon as this is hydrated on the client the prop would be unavailable right?
And thus clearing anything which was prerendered on the server isnt it?

PR #742, opened 2019-02-22
Modnomod in watch

### LINKED ISSUE #721: Option to use esm build with dev server
feature

no option to use esm build with dev server

--esm flag for dev server (preact watch)

we have been noticing issues that show up only in esm builds and would like to be able to fix them while using preact watch for eg. this fails in esm build only -
```
render({
		image = requireGameImage(`${gameId}/eventTile.jpg`),
		game: gameName,
		gameId
	 }.....
```

- Node version latest
- npm version latest
- Operating system windows
- CLI version 3.0.0-next.16
- Browser chrome

prateekbh:
Fixes #721 

Adds module/nomodule in `preact watch`

No

**Summary**

Adds the esm capability in `preact watch`.

No

### DIFF
--- a/packages/cli/lib/lib/webpack/webpack-client-config.js
+++ b/packages/cli/lib/lib/webpack/webpack-client-config.js
@@ -230,26 +230,24 @@ function isProd(config) {
 		);
 	}
 
-	if (config.esm) {
-		if (config.sw) {
-			prodConfig.plugins.push(
-				new SWPrecacheWebpackPlugin({
-					filename: 'sw-esm.js',
-					navigateFallback: 'index.html',
-					navigateFallbackWhitelist: [/^(?!\/__).*/],
-					minify: true,
-					stripPrefix: config.cwd,
-					staticFileGlobsIgnorePatterns: [
-						/(\.[\w]{5}\.js)/,
-						/polyfills(\..*)?\.js$/,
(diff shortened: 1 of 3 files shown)


packages/cli/lib/lib/webpack/webpack-client-config.js
```diff
@@ -180,44 +231,6 @@ function isProd(config) {
 	}
 
 	if (config.esm) {
```
ForsakenHarmony: guess this can be simplified now

prateekbh: fixed

packages/cli/lib/lib/webpack/webpack-client-config.js
```diff
@@ -109,6 +113,53 @@ function clientConfig(env) {
 	};
 }
 
+function getBabelEsmPlugin(config) {
```
ForsakenHarmony: why did you move this out btw?

prateekbh: Oh right.. This was intermittent decision which should have not been moved out

ForsakenHarmony: oh 👀

PR #743, opened 2019-02-25
fix not generating sw.js when it's disabled

arash16:
config.serviceWorker is string 'false' and condition was still hold true.


prateekbh:
Are you using `--no-service-worker`

arash16:
@prateekbh nope, I was using --service-worker=false :grin:

lukeed:
Thank you, but this is for 2.x 

In any event, `--no-foo` is always the convention.

PR #744, opened 2019-02-25
[next] fix keyframe name collisisions

developit:
This fixes `@keyframes` rules from multiple CSS chunks being minified to the same name (generally "a" or "b"). Currently, `optimize-css-assets-webpack-plugin` does not have a way to minify global identifiers across assets without colliding.


hassanbazzi:
We're using this on the new Preact website. Can confirm it works.

PR #747, opened 2019-02-27
Typescript support

ForsakenHarmony:


packages/cli/lib/lib/webpack/webpack-base-config.js
```diff
@@ -105,18 +112,27 @@ module.exports = function(env) {
 
 		module: {
 			rules: [
+				{
+					enforce: 'pre',
+					test: /\.tsx?$/,
+					loader: 'awesome-typescript-loader',
+					options: {
+						silent: true,
+						useBabel: true,
```
developit: I think this will break BabelEsmPlugin, which looks for babel-loader. 

Would it be possible to run awesome-ts-loader as `{enforce:'pre'}` and then pass `/\.[tj]sx?/` files to babel-loader after?

ForsakenHarmony: hmm good question

packages/cli/package.json
```diff
@@ -108,6 +110,9 @@
     "stack-trace": "0.0.10",
     "style-loader": "^0.23.1",
     "sw-precache-webpack-plugin": "^0.11.2",
+    "terser-webpack-plugin": "^1.1.0",
+    "typescript": "^3.3.3333",
+    "unfetch": "^4.0.1",
```
sync: Just curious, what are we doing with unfetch ?

sync:
Would be great to get “plug and play” ts support

packages/cli/package.json
ForsakenHarmony: good question, no idea where that came from, probably something old that slipped in

packages/cli/lib/lib/webpack/webpack-base-config.js
```diff
@@ -61,11 +72,29 @@ module.exports = function(env) {
 	let cliNodeModules = findAllNodeModules(__dirname);
 	let nodeModules = [...new Set([...userNodeModules, ...cliNodeModules])];
 
+	let compat = 'preact-compat';
+	try {
+		requireRelative.resolve('preact/compat', cwd);
```
developit: Should we be trying to resolve this only to `node_modules/preact/compat`? I think this will resolve to `node_modules/preact/compat/dist/compat.js`.

PR #754, opened 2019-03-05
Better Prerender Errors

ForsakenHarmony:


prateekbh:
is this still active?

reznord:
I can work on this and we need to improve our logging much better.

ForsakenHarmony:
Ran into issues with reliably getting the right source map and stuff like that

prateekbh:
Got it

PR #756, opened 2019-03-11
More Preact X compat things™

ForsakenHarmony:


packages/cli/package.json
```diff
@@ -92,8 +102,6 @@
     "optimize-css-assets-webpack-plugin": "^5.0.1",
     "ora": "^3.0.0",
     "postcss-loader": "^3.0.0",
-    "preact": "^8.1.0",
```
developit: Does this break prerendering?

packages/cli/lib/lib/webpack/webpack-base-config.js
```diff
@@ -6,7 +6,7 @@ const SizePlugin = require('size-plugin');
 const autoprefixer = require('autoprefixer');
 const requireRelative = require('require-relative');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
-const FixStyleOnlyEntriesPlugin = require('webpack-fix-style-only-entries');
```
developit: Bah. Will the fix still let this strip out the Webpack modules that have nothing in them? They should have no originalModule since they're not from an ESM source?

ForsakenHarmony: should fix it, just commented it out for now, because it broke the tests

packages/cli/package.json
ForsakenHarmony: it shouldn't, as a peerDep it should resolve fine

packages/cli/package.json
```diff
@@ -64,21 +76,21 @@
...
     "console-clear": "^1.0.0",
-    "copy-webpack-plugin": "^5.0.0",
+    "copy-webpack-plugin": "^5.0.1",
     "critters-webpack-plugin": "^1.3.3",
     "cross-spawn-promise": "^0.10.1",
-    "css-loader": "^2.1.0",
+    "css-loader": "^2.1.1",
     "ejs-loader": "^0.3.0",
-    "esm": "^3.2.7",
+    "esm": "<=3.2.14",
```
ForsakenHarmony: https://github.com/standard-things/esm/issues/747

jdalton: `esm` [v3.2.17](https://github.com/standard-things/esm/releases/tag/3.2.17) is released :tada:

packages/cli/lib/lib/webpack/webpack-base-config.js
```diff
@@ -61,11 +61,17 @@ module.exports = function(env) {
 	let cliNodeModules = findAllNodeModules(__dirname);
 	let nodeModules = [...new Set([...userNodeModules, ...cliNodeModules])];
 
+	let compat = 'preact-compat';
+	try {
+		requireRelative.resolve('preact/compat', cwd);
```
developit: Should we be trying to resolve this only to `node_modules/preact/compat`? I think this will resolve to `node_modules/preact/compat/dist/compat.js`.

PR #764, opened 2019-03-21
Allow async prerender urls

prateekbh:

It allows prerender-urls.js to return a promise

Yes

**Summary**
Currently `prerender-urls.js` can return a function but only sync. This allows that function to be async as well. Allowing `preact-cli` to be a static site generator

No

### DIFF
--- a/packages/cli/lib/lib/webpack/render-html-plugin.js
+++ b/packages/cli/lib/lib/webpack/render-html-plugin.js
@@ -5,9 +5,11 @@ const HtmlWebpackPlugin = require('html-webpack-plugin');
 const prerender = require('./prerender');
 const createLoadManifest = require('./create-load-manifest');
 const { warn } = require('../../util');
+const { yellow } = require('kleur');
+
 let template = resolve(__dirname, '../../resources/template.html');
 
-module.exports = function(config) {
+module.exports = async function(config) {
 	const { cwd, dest, isProd, src } = config;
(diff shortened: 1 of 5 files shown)


packages/cli/lib/lib/webpack/webpack-client-config.js
```diff
@@ -84,7 +84,7 @@ function clientConfig(env) {
 
 		plugins: [
 			new PushManifestPlugin(env),
-			...RenderHTMLPlugin(env),
+			...(await RenderHTMLPlugin(env)),
```
ForsakenHarmony: maybe we should lowercase this, looks like a class here

prateekbh: Done

packages/cli/lib/lib/webpack/render-html-plugin.js
```diff
@@ -66,7 +68,9 @@ module.exports = function(config) {
 					result = result.default();
 				}
 				if (typeof result === 'function') {
-					result = result();
+					console.log(yellow(`Fetching URLs from ${config.prerenderUrls}`));
+					result = await result();
+					console.log(yellow(`Fetched URLs from ${config.prerenderUrls}`));
```
ForsakenHarmony: shouldn't this use our log functions from util?

prateekbh: done

PR #767, opened 2019-03-22
temporarily bypass async loader for preact x

ForsakenHarmony:


prateekbh:
But wouldn't this disable coffee splitting?

reznord:
In `Preact X`, yes it will disable code splitting

developit:
I don't think it's a good idea to ship code splitting in a disabled state.

prateekbh:
+1 ^

ForsakenHarmony:
Was meant as a temporary workaround so people can use preact cli with x until it's fixed

developit:
one option here would be to remove the base reuse stuff and just have the async loader's component render null while waiting. It'd just have the flash of no content problem we had before @lukeed's fix, but still async/lazy.

prateekbh:
Which works but still is ugly 😐

developit:
Indeed. I spent a few hours playing around with using componentDidCatch for this, but I'm running into some alpha.2 bugs.

ForsakenHarmony:
I've been using `yarn link` a bunch recently for testing

developit:
that's my usual approach yeah.

reznord:
I think @developit has a PR on the way for async component in Preact X.

PR #770, opened 2019-04-07
Do not force a favicon

MasterRennor:
This PR removes the default/fallback favicon. (https://github.com/developit/preact-cli/issues/643)

The specs are updated to reflect the no-icon-by-default markup. This could be improved, as the case where an asset _does_ exist is no longer covered by the specs.

### DIFF
--- a/packages/cli/lib/lib/webpack/render-html-plugin.js
+++ b/packages/cli/lib/lib/webpack/render-html-plugin.js
@@ -28,7 +28,7 @@ module.exports = async function(config) {
 			},
 			favicon: existsSync(resolve(src, 'assets/favicon.ico'))
 				? 'assets/favicon.ico'
-				: resolve(__dirname, '../../resources/favicon.ico'),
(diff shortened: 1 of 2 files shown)


prateekbh:
But why?
This is a default favicon. You can always override it.

Is that a problem?

MasterRennor:
Well, it sort of looked like there was a green light for that:
image

Personally, I also don't mind a fallback favicon, but I guess the issue for the original ticket poster is that it might interfere with their branding.

reznord:
@MasterRennor instead of removing it, can we make it optional? 

If only favicon is present, then can we add `<link rel='favicon' href='favicon.ico' />` to the html?

MasterRennor:
@reznord, if I'm not mistaken, I am only removing the fallback icon which is added when there is no favicon provided by the user. So, effectively, this PR is doing just that: only adding a favicon, if there is one in `src/assets`.

PR #775, opened 2019-04-13
Fix weird highlighting in README

zaccolley:
Looks like a `'` in the README was adding highlighting we don't want.

__Before:__

__After:__


zaccolley:
Sorry made two commits, can close and do another if you can't squash?

Issue #443, opened 2017-12-21
Documentation on how to use SASS


jeanlucaslima:
**Do you want to request a *feature* or report a *bug*?**
Feature/improvement. 

**If this is a feature request, what is motivation or use case for changing the behavior?**
The documentation won't explain easily how to use things like sass, and even lead to incorrect behaviours like using the flag `--sass` as mentioned in #411

guido4000:
This is how I added it:

* $ npm install --save-dev node-sass sass-loader
* create helpers.scss file
* rename index and component style file extensions to .scss
* import helpers.scss in index and component scss files: @import "src/style/helpers"

jeanleonino:
@guido4000 yeah, it indeed does that and is solved on #411. This issue is about having this documented.

developit:
I would really love to fix our auto-installation for `sass`/`sass-loader`.  Then we could add automatic renaming during project init (`preact create default foo --sass`) and there'd be no more steps!

jeanleonino:
@developit that sounds the best approach, actually. 

Why was the `--sass` removed? How can this be changed? I am willing to try a PR solving it.

lukeed:
It removed as a reaction to having way too many flags for the CLI haha. We probably went a little too far in the other direction, but now we have plans to move back towards Jason's comment. 🙌

dsky1990:
same idea with you @jeanleonino

vutran:
Any plans on getting this reimplemented (along with other loaders)? Otherwise, we should just update the documentation with the manual steps of installing the appropriate packages and loaders.

Another idea would be automatically installing these required packages/loaders on the `watch` command upon demand if a one of these alternative CSS lang (`stylus`, `less`, `sass`, etc.) is present in `src/style`.

lukeed:
Yeah! Right now auto-installation is on the roadmap. Unfortunately doesn't help right now, so you still need to rename your file extensions and install `node-sass` and `sass-loader` manually.

vutran:
Lmk where I can help.

Side note, it would make sense to document it in the meantime in the `README`. If so, maybe @jeanleonino can create the PR as originally suggested.

MattSidor:
Can this issue please be re-opened? I am struggling to figure out the correct config changes I need to make to get `*.scss` files to transpile/load correctly...

MattSidor:
I think I've figured this out for my project.

[...]

Issue #411, opened 2017-10-18
How to create a project with sass?


expdevelop:
Sorry guys, maybe I'm stupid, but I'm literally CAN'T find any guide to create a project on preact-cli with sass usage instead of css.

So, could anyone help me with that? 
`preact create default app --sass` is not working.

SaraVieira:
Hi !

Can you tell me what is the error you are getting ?

expdevelop:
@SaraVieira `preact create default app --sass` -> `Unknown argument: sass`

lukeed:
There's no `--sass` flag anymore. Just install `node-sass` and `sass-loader` in your app and you're ready to go.

```sh
$ preact create default app
$ npm install --save-dev node-sass sass-loader
```

expdevelop:
@lukeed I guessed it by myself but I got some errors in console.

`⚠ WARN ./node_modules/preact-cli/lib/lib/webpack/proxy-loader.js?{"__proxy_loader__":{"loader":"sass-loader","cwd":"/${projectpath}/app"},"sourceMap":true}!./src/components/Input/Input.sass
(Emitted value instead of an instance of Error) 

 ⚠️  PostCSS Loader

Previous source map found, but options.sourceMap isn't set.
In this case the loader will discard the source map entirely for performance reasons.
See https://github.com/postcss/postcss-loader#sourcemap for more information.


 @ ./src/components/Input/Input.sass 4:14-226 18:2-22:4 19:20-232
 @ ./src/components/Input/Input.js
 @ ./src/components/FeedbackForm/FeedbackForm.js
 @ ./src/routes/Home/Home.js
 @ ./src/components/App/App.js
 @ ./src/index.js
 @ ./node_modules/preact-cli/lib/lib/entry.js`

lukeed:
@expdevelop Great! That's just a warning, and it's solved by #407 

For the meantime (since it's annoying) you can do this:

```js
let sass = helpers.getLoadersByName(config, 'sass-loader')[0];
sass.loader.options.sourceMap = true;
```

Hope that helps!

botteu:
Where would I add this:
`let sass = helpers.getLoadersByName(config, 'sass-loader')[0];
sass.loader.options.sourceMap = true;`

lukeed:
@botteu Inside a [`preact.config.js`](https://github.com/developit/preact-cli/blob/master/README.md#webpack)

expdevelop:
@lukeed I got error with that

[...]

PR #783, opened 2019-05-01
Add information on SASS and LESS preprocessing

stefanpl:

A simple attempt to provide some documentation on using css preprocessors, as discussed (for almost 1,5 years) in #443 and #411. I only know SASS and LESS, so I only added paragraphs for those. Feel free to add stylus/css modules/anything else.


ForsakenHarmony:
Thanks 🎉

PR #790, opened 2019-05-08
Added .wasm extension to base webpack config for support of importing .wasm files

matthewharwood:
Feature: An update to the webpack-base-config.js file

No
**Summary**
Currently preact-cli will not allow the import of .wasm files. This update should afford the opportunity to do so.

It shouldn't

**Other information**
N/A


ForsakenHarmony:
why not 🙃

PR #803, opened 2019-05-20
Using same version of preact for pre-rendering

### LINKED ISSUE #795: Duplicate preact instance while prerendering breaking options
Ported over from: https://github.com/developit/preact-render-to-string/issues/86#issuecomment-492325953

bug

Prerendering will throw an undefined error:

```txt
TypeError: Cannot read property '__H' of undefined
```

This is caused by loading multiple instances of preact which can't see each others `option`-hooks. There are a total of 2 instances at play:

1) included in the generated `ssr-bundle.js`
2) one inside `prerender.js`: https://github.com/developit/preact-cli/blob/bd464043dac8646183ed9a8cfab677778f73b4f7/packages/cli/lib/lib/webpack/prerender.js#L28

Both instances have their own `options` variable which ultimately leads to this error.

```json
{
  "scripts": {
    "build": "preact build"
  },
  "devDependencies": {
    "preact-cli": "^3.0.0-rc.1",
    "preact-render-to-string": "^5.0.3"
  },
  "dependencies": {
    "preact": "^10.0.0-beta.1"
  }
}
```

```jsx
import { useState } from "preact/hooks";

export default () => {
  const [a, b] = useState("c");
  return <h1>Hello, World! {a}</h1>;
};
```

It should not throw.

- Node version: 11.7
- npm version 6.5.0
- Operating system: Linux Debian Sid
- CLI version: 3.0.0-rc.1
- Browser: doesn't matter

prateekbh:
Fixes #795 

Fixes duplicate preact problem.

- Uses `preact` . and `preact-render-to-string` from user land.
- Fall back to local `preact-render-to-string` to local copy but wont work with preact-X
- **Note:** After this we need to include `preact-render-to-string` in all our template's `package.json`

Already exist

No

### DIFF
--- a/packages/cli/lib/lib/webpack/prerender.js
+++ b/packages/cli/lib/lib/webpack/prerender.js
@@ -1,6 +1,6 @@
 const { red, yellow } = require('kleur');
 const { resolve } = require('path');
-const { readFileSync } = require('fs');
+const { readFileSync, existsSync } = require('fs');
 const stackTrace = require('stack-trace');
 const { SourceMapConsumer } = require('source-map');
 
@@ -24,10 +24,15 @@ module.exports = function(env, params) {
 			);
 			return '';
 		}
-
-		let preact = require('preact'),
-			renderToString = require('preact-render-to-string');
-
+		const { cwd } = env;
+		const preactPath = require.resolve(`${cwd}/node_modules/preact`);
+		const renderToStringPath = require.resolve(
+			`${cwd}/node_modules/preact-render-to-string`
+		);
+		let preact = require(preactPath),
+			renderToString = require(existsSync(renderToStringPath)
+				? renderToStringPath
+				: 'preact-render-to-string');
 		return renderToString(preact.h(app, { ...params, url }));
 	} catch (err) {
 		let stack = stackTrace.parse(err).filter(s => s.getFileName() === entry)[0];
--- a/packages/cli/lib/lib/webpack/prerender.js
+++ b/packages/cli/lib/lib/webpack/prerender.js
@@ -1,6 +1,6 @@
 const { red, yellow } = require('kleur');
 const { resolve } = require('path');
-const { readFileSync, existsSync } = require('fs');
+const { readFileSync } = require('fs');
 const stackTrace = require('stack-trace');
 const { SourceMapConsumer } = require('source-map');
 
@@ -25,14 +25,10 @@ module.exports = function(env, params) {
 			return '';
 		}
 		const { cwd } = env;
-		const preactPath = require.resolve(`${cwd}/node_modules/preact`);
-		const renderToStringPath = require.resolve(
-			`${cwd}/node_modules/preact-render-to-string`
-		);
-		let preact = require(preactPath),
-			renderToString = require(existsSync(renderToStringPath)
-				? renderToStringPath
-				: 'preact-render-to-string');
+		let preact = require(require.resolve(`${cwd}/node_modules/preact`)),
+			renderToString = require(require.resolve(
+				`${cwd}/node_modules/preact-render-to-string`
+			));
 		return renderToString(preact.h(app, { ...params, url }));
 	} catch (err) {
 		let stack = stackTrace.parse(err).filter(s => s.getFileName() === entry)[0];
--- a/packages/cli/lib/lib/webpack/prerender.js
+++ b/packages/cli/lib/lib/webpack/prerender.js
@@ -25,10 +25,12 @@ module.exports = function(env, params) {
 			return '';
 		}
 		const { cwd } = env;
-		let preact = require(require.resolve(`${cwd}/node_modules/preact`)),
-			renderToString = require(require.resolve(
-				`${cwd}/node_modules/preact-render-to-string`
-			));
+
+		const preact = require(require.resolve(`${cwd}/node_modules/preact`));
+		const renderToString = require(require.resolve(
+			`${cwd}/node_modules/preact-render-to-string`
+		));
+
 		return renderToString(preact.h(app, { ...params, url }));
 	} catch (err) {
 		let stack = stackTrace.parse(err).filter(s => s.getFileName() === entry)[0];
--- a/packages/cli/lib/lib/webpack/webpack-server-config.js
+++ b/packages/cli/lib/lib/webpack/webpack-server-config.js
@@ -15,11 +15,7 @@ function serverConfig(env) {
 			libraryTarget: 'commonjs2',
 		},
 		externals: {
-			/** This makes pre-rendered bundle use the same preact as preact-cli.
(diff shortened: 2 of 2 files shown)


hassanbazzi:
Maybe we should add a note on that somewhere in the FAQ maybe? or Readme?

prateekbh:
about the comments i added?

hassanbazzi:
Well..... For users. I think users will expect the user-land preact to affect pre-rendering Preact.

prateekbh:
**Note:** I really wanted both bundle and `pre-rendering.js` to use preact from user land but sadly doesn't work that way. Can anyone help tell me why

// @marvinhagemeister

prateekbh:
@hassanbazzi I updated the PR to use modules from user land instead its just that it has an added step of installing `preact-render-to-string` in user land

packages/cli/lib/lib/webpack/prerender.js
```diff
@@ -24,10 +24,15 @@ module.exports = function(env, params) {
...
-
+		const { cwd } = env;
+		const preactPath = require.resolve(`${cwd}/node_modules/preact`);
+		const renderToStringPath = require.resolve(
+			`${cwd}/node_modules/preact-render-to-string`
+		);
+		let preact = require(preactPath),
+			renderToString = require(existsSync(renderToStringPath)
+				? renderToStringPath
+				: 'preact-render-to-string');
```
ForsakenHarmony: kinda wondering if we actually need this, it should already resolve from cwd?

and we don't have it as a fixed dep anymore https://github.com/developit/preact-cli/blob/master/packages/cli/package.json#L64-L67

prateekbh: preact/preact-render-to-string both should be used from same source otherwise `__hooks` fail.

Thats why we need to force require this from user land. This will allow user to have the flexibility of using any version of preact/render-to-string they want.

If they don't have, then we can fallback to `cwd` but any usage of hooks will fail.

This'll make it less of a breaking change for everyone as the old ones continue to work as is and the new one will install one more dep and can use hooks

ForsakenHarmony: yeah I'm saying it should only exist in user land either way

prateekbh: that breaks old users, no? just trying to minimize the breaking change impact

ForsakenHarmony: it's already this way for cli v3

prateekbh: well... its also not used anymore.. so ummm makes sense, will make the change

ForsakenHarmony: I meant we shouldn't need the `${cwd}` part? or am I wrong?

prateekbh: we actually do need ${cwd}. Otherwise it uses `preact/render-to-string` from preact-cli's `node_modules` folder instead of the created app's folder

prateekbh: Also `${cwd}` points to the working directory of the created project and not the where `preact-cli` is installed globally.

ForsakenHarmony: well `preact-cli` doesn't have them anymore

but yes it could get in the way if you're not using it via an npm script and or it's just installed globally

prateekbh: Oh sorry for not conveying that.. in my head everyone uses it like that.

Anyways, i'll fix the test

ForsakenHarmony: yeah I kinda assumed having it locally installed, you'll want that for automated builds either way

packages/cli/lib/lib/webpack/prerender.js
```diff
@@ -24,10 +24,11 @@ module.exports = function(env, params) {
 			);
 			return '';
 		}
-
-		let preact = require('preact'),
-			renderToString = require('preact-render-to-string');
-
+		const { cwd } = env;
+		let preact = require(require.resolve(`${cwd}/node_modules/preact`)),
```
ForsakenHarmony:
```suggestion
		const preact = require(require.resolve(`${cwd}/node_modules/preact`));
```

packages/cli/lib/lib/webpack/prerender.js
```diff
@@ -24,10 +24,11 @@ module.exports = function(env, params) {
...
+			renderToString = require(require.resolve(
```
ForsakenHarmony:
```suggestion
		const renderToString = require(require.resolve(
```

packages/cli/lib/lib/webpack/prerender.js
```diff
@@ -24,10 +24,11 @@ module.exports = function(env, params) {
...
+				`${cwd}/node_modules/preact-render-to-string`
+			));
```
ForsakenHarmony:
```suggestion
			));
			
```

ForsakenHarmony:
Not entirely sure how to write a test for it

prateekbh:
:| this was breaking the current tests....
i'll add the fix for tests ASAP

ForsakenHarmony:
I already have it

packages/cli/lib/lib/webpack/prerender.js
```diff
@@ -24,9 +24,12 @@ module.exports = function(env, params) {
 			);
 			return '';
 		}
+		const { cwd } = env;
 
-		let preact = require('preact'),
-			renderToString = require('preact-render-to-string');
+		const preact = require(require.resolve(`${cwd}/node_modules/preact`));
```
tomsiwik: Currently, using a monorepo manager like lerna this workaround breaks the build process and returns:

`✖ ERROR Template execution failed: Error: Cannot find module './node_modules/preact'`

The `cwd` is located inside `root/packages/mypack` and is looking for a `root/packages/mypack/node_modules`, which however is located in `root/node_modules`. Node won't resolve correctly thanks to lerna and thus this will break.

I reverted the changes and logged destination. A revert worked but I think changing the destination to: `${__dirname}/../../../../preact` and `${__dirname}/../../../../preact-render-to-string` would work too.

alexjsdev: is there any workaround for this? I just moved a preact app into my monorepo to not have to maintain it in a separate repo, and I'm running into this exact error.

tomsiwik: With Lerna and Yarn workspaces no. If you disable yarn workspaces this should work with Lerna though.

merceyz: Fixed in https://github.com/preactjs/preact-cli/pull/1418

PR #806, opened 2019-05-29
Include readme in npm tarball

developit:
Currently it's not included, since the readme is in the repo root.
I wasn't sure if the publishing process happens from within `packages/cli` or the root.
https://www.npmjs.com/package/preact-cli/v/3.0.0-rc.2

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,9 @@
     "pretest": "yarn lint",
     "test": "lerna run test --stream",
     "prettier": "prettier --write **/*.{js,ts,tsx,json,css,scss,md,yml}",
-    "lint": "eslint ."
+    "lint": "eslint .",
+    "prepack": "cp README.md packages/cli/",
+    "postpack": "rm packages/cli/README.md"
   },
   "devDependencies": {
     "babel-eslint": "^10.0.1",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -7,7 +7,9 @@
     "pretest": "yarn lint",
     "test": "lerna run test --stream",
     "prettier": "prettier --write **/*.{js,ts,tsx,json,css,scss,md,yml}",
-    "lint": "eslint ."
+    "lint": "eslint .",
+    "prepack": "cp README.md packages/cli/",
+    "postpack": "rm packages/cli/README.md"
```
ForsakenHarmony: don't think cp and rm work properly on windows

hassanbazzi: They're usually used throughout package.json scripts. Should be fine imo

mseddon: A lot of package.json scripts blow up on windows due to Unix shell commands. I'd recommend testing.

mseddon: Adding a `cp` script under windows 10 bails for me with:
```
'cp' is not recognized as an internal or external command,
operable program or batch file.
```

ForsakenHarmony: `&&` works because npm uses cmd in the background afaik

we're already using `rimraf` so might as well use that here and then some non bloated copy cli (looking at `cpy-cli`)

ForsakenHarmony: I'll push to this later

ForsakenHarmony:
Not even a new dep with that

PR #810, opened 2019-06-09
Update autoprefixer to 9.6.0

filoozom:

Updates `autoprefixer` to `9.6.0`

No

**Summary**

The `browsers` option is now deprecated as of `9.6.0` and will be removed in `10.0.0`. The latest version of `autoprefixer` now shows a big red warning when using its latest version:

```
Replace Autoprefixer browsers option to Browserslist config.
Use browserslist key in package.json or .browserslistrc file.

Using browsers option cause some error. Browserslist config 
can be used for Babel, Autoprefixer, postcss-normalize and other tools.

If you really need to use option, rename it to overrideBrowserslist.

Learn more at:
https://github.com/browserslist/browserslist#readme
https://twitter.com/browserslist
```

I don't think there's a way to implicitly use the parents `browserlist` file or configuration in `package.json`, so I think using `overrideBrowserslist` is the best solution for now.

No

**Other information**

None

### DIFF
--- a/packages/cli/lib/lib/webpack/webpack-base-config.js
+++ b/packages/cli/lib/lib/webpack/webpack-base-config.js
@@ -208,7 +208,7 @@ module.exports = function(env) {
 							options: {
 								ident: 'postcss',
 								sourceMap: true,
-								plugins: [autoprefixer({ browsers })],
+								plugins: [autoprefixer({ overrideBrowserslist: browsers })],
 							},
 						},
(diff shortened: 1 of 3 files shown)


ForsakenHarmony:
does yarn.lock not need to be updated?
as in run yarn once in the root dir

filoozom:
Ah yes indeed, don't really use `yarn` myself.

Issue #816, opened 2019-06-18
Is service worker customization tougher than v2


prateekbh:
**Do you want to request a _feature_ or report a _bug_?**
Question

**What is the current behaviour?**
In V3 we moved from SwPrecache to Workobx jS. 
While moving we added a functionality to be able to create both module/nomodule service worker from the same source service worker file defined [here](https://github.com/preactjs/preact-cli/blob/master/packages/cli/lib/lib/sw.js).
This uses InjectManifest plugin from Workbox which injects precache file in a handwritten service worker. 

This also gives us extra functionality to do brotli/webP redirect whenever user opts in.

While all this is great, any additional changes to the service worker needs to be done in code instead of changing some config in v2.

I just wonder if that might come across as a tougher choice for the users and whether we should just stick to the config based [GenerateSwWebpackPlugin](https://developers.google.com/web/tools/workbox/modules/workbox-webpack-plugin#generatesw_plugin) from Workbox and may be write blogs/wiki on how to do the brotli/webP with preact-cli.

prateekbh:
// @developit @hassanbazzi @ForsakenHarmony @lukeed

ForsakenHarmony:
was it easy to customise it before?

ideally it'd be easy to extend, not just replace the service worker

and we could maybe add options to `preact.config.js`

prateekbh:
Its like with current setup you need to write code to change the functionality, vs in earlier setup you needed to change config of plugin.

I guess its just more tough for ppl to write code for SWs, thats why i opened this issue

prateekbh:
If we agree that this is the case #819 will fix it

ForsakenHarmony:
yeah I don't like modifying config of existing plugins in `preact.config.js` I'd rather have something more intuitive

prateekbh:
@ForsakenHarmony  just to confirm, so you feel writing workbox code for your service worker would be more intuitive than editing an object in `preact.config.js`?

just a nit, our users would need to learn workbox APIs to stuff like runtime caching etc vs adding them in config.

ForsakenHarmony:
No I'm saying changing to a differnt plugin and not changing our public api (I don't count modifying the webpack config) doesn't change anything

prateekbh:
Ok a quick show of hands please to sort this out 

**Reply on this comment with 👍 / 👎** 


👍 - Yes lets shift to config based service worker generator as is done in next/nuxt/gatsby etc

[...]

PR #817, opened 2019-06-20
Fix issue with not using custom sw.js even if recognized

hesselbom:
Bugfix

No, not sure how to write tests for it so if someone could help me it'd be great.

**Summary**
When providing a custom sw.js in _/src/sw.js_ or _/sw.js_ the console log said "Detected custom sw.js: compiling instead of default Service Worker." but didn't actually use that file, instead always using the default sw.js.

There was also a redundant if statement (`if (exists) { if (exists) {`)

No

**Other information**
N/A

### DIFF
--- a/packages/cli/lib/lib/webpack/sw-plugin.js
+++ b/packages/cli/lib/lib/webpack/sw-plugin.js
@@ -15,13 +15,12 @@ class SWBuilderPlugin {
 		let swSrc = resolve(__dirname, '../sw.js');
 		const exists = fs.existsSync(resolve(`${this.src_}/sw.js`));
 		if (exists) {
-			if (exists) {
-				info(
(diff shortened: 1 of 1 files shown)


prateekbh:
I literally discovered this yesterday... 
one of the facepalm moments

hesselbom:
:D

Any ETA for when this will be released on NPM, i.e. bumped preact-cli version? I'd love to use this now but afaik I can't just reference this repo as npm dependency because of the lerna setup, if I'm not mistaken.

prateekbh:
@hesselbom I would suggest to use `--no-sw` for now, and register it manually. As we're not sure if we want to continue with this service worker for 3.0 or switch to a config based as was in v2 see #816

PR #818, opened 2019-06-24
update readme commands

AEnterprise:
fixes the name for the option that controls if the serviceworker should be enabled on the production or not as well as adding a note on the --no-<option> form of the params.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ Note: If you don't specify enough data to the `preact create` command, it will p
 
 Create a production build
 
-For the boolean fields you can pass --<option> to set to true or --no-<option> to set to false.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -81,14 +81,16 @@ Note: If you don't specify enough data to the `preact create` command, it will p
 
 Create a production build
 
+For the boolean fields you can pass --<option> to set to true or --no-<option> to set to false.
```
lukeed:
```suggestion
You can disable `default: true` flags by prefixing them with `--no-<option>` – for example. `--no-sw`
```

AEnterprise:
@lukeed copied them all from the cli now, server no longer seems to exist though so i removed that one

PR #826, opened 2019-07-08
ServiceWorker / Workbox Tweaks

developit:
Just a couple of tweaks:

- fixed a few regexes that were missing escapes (they'd match `foobarjs` as if it were a JS file)
- only precache the index route, since all navigation requests get routed to it once the SW has booted.

### DIFF
--- a/packages/cli/lib/lib/sw.js
+++ b/packages/cli/lib/lib/sw.js
@@ -9,8 +9,8 @@ self.__precacheManifest = [].concat(self.__precacheManifest || []);
 if (process.env.ENABLE_BROTLI && process.env.ES_BUILD) {
 	// Alter the precache manifest to precache brotli files instead of gzip files.
 	self.__precacheManifest = self.__precacheManifest.map(asset => {
-		if (/.*.js$/.test(asset.url)) {
-			asset.url = asset.url.replace(/.esm.js$/, '.esm.js.br');
+		if (/\.js$/.test(asset.url)) {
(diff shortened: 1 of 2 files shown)


developit:
Alright I think I found a fix for the navigation route.  We can replace the `workbox.routing.setNavigationRoute` usage with this:

```js
const isNav = event => event.request.mode === 'navigate';

workbox.routing.registerRoute(
  ({ url, event }) => isNav(event),
  workbox.strategies.networkOnly()
});

workbox.routing.setCatchHandler(({event}) => {
  if (isNav(event)) return caches.match('/index.html');
  return Response.error();
});
```

prateekbh:
The suggested changes sound great, do you wanna do this in the same PR @developit ?


Also can you open a separate issue for 200.html thing? Sounds nice would wanna think through though

prateekbh:
Also let's make that networkFirst/StaleWhileRevalidate instead of NetworkOnly and add a timeout of say 5s.
So if network doesn't respond withing 5s fallback to cache but also update the cache

developit:
Isn't it already basically doing NetworkFirst? When networkOnly() throws (due to timeout or offline), it should fall through to the catch handler, which responds with the `index.html`.

prateekbh:
That makes sense in that case should we just add the timeout as well?

prateekbh:
@developit solution in https://github.com/preactjs/preact-cli/pull/826#issuecomment-509274883
work for everything but homepage as its already precached so always returns from cache

PR #830, opened 2019-07-08
Always enable CSS sourcemaps

jonathantneal:

This change allows CSS source maps to be generated in development.

The current behavior is that CSS source maps are **not** generated in development. This PR changes that because source maps are intended to improve the _development_ experience. They are most often leveraged by _Developer_ Tools to make it “_much easier to debug the original source, rather than the source in the transformed state that the browser has downloaded_” ([mdn](https://developer.mozilla.org/en-US/docs/Tools/Debugger/How_to/Use_a_source_map)).

Admittedly, I would need to understand the rationale for the current behavior to be sure of this change. The current behavior is that CSS source maps are generated in production _only_. This seems odd to me, because I’ve understood production to be when smaller file sizes are more desirable, and debugging features are more often removed — for instance this project removes prototype debugging in production (https://github.com/preactjs/preact-cli/blob/master/packages/cli/lib/lib/babel-config.js#L30-L31).

No, the current source map behavior is also not tested.

**Summary**

I was excited to use CSS Modules, but found it difficult to debug in development without source maps.

No, this does not introduce a breaking change.

### DIFF
--- a/packages/cli/lib/lib/webpack/webpack-base-config.js
+++ b/packages/cli/lib/lib/webpack/webpack-base-config.js
@@ -210,7 +210,7 @@ module.exports = function(env) {
 								modules: true,
 								localIdentName: '[local]__[hash:base64:5]',
 								importLoaders: 1,
-								sourceMap: isProd,
+								sourceMap: true,
 							},
 						},
 						{
@@ -232,7 +232,7 @@ module.exports = function(env) {
 						{
 							loader: 'css-loader',
 							options: {
-								sourceMap: isProd,
+								sourceMap: true,
 							},
 						},
 						{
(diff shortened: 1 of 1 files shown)


ForsakenHarmony:
I think it might be the wrong way around currently, maybe it should have been a `!isProd`

jonathantneal:
Thanks of the quick feedback, @ForsakenHarmony. I will follow your lead and gladly make the changes if you decide that you would like this PR modified to work the other way around.

ForsakenHarmony:
I think this has to be explicitly enabled for `style-loader` as well

ForsakenHarmony:
https://github.com/webpack-contrib/style-loader#sourcemap

jonathantneal:
I’ve explicitly enabled `sourceMap` for `style-loader` as well, @ForsakenHarmony, and put it into its own commit (see https://github.com/preactjs/preact-cli/pull/830/commits/b73fc711784c2b199513b89552fb92e00b2aa069). Let me know if that is how you wanted it done, or if you would it done another way. 😄

ForsakenHarmony:
not sure what the failure is, tests seem to pass 🙃

PR #832, opened 2019-07-08
Support PostCSS plugin configuration

jonathantneal:

This change allows [the existing PostCSS Loader](https://github.com/preactjs/preact-cli/blob/v3.0.0-rc.2/packages/cli/lib/lib/webpack/webpack-base-config.js#L206-L213) to respect userland `postcss.config.js` configuration files in their ability to define the PostCSS plugins used in a project.

See https://github.com/postcss/postcss-loader#usage

No, and I did not see similar tests for PostCSS plugins in the existing project.

**Summary**

I would like to use [PostCSS Normalize](https://github.com/csstools/postcss-normalize) and [PostCSS Preset Env](https://github.com/csstools/postcss-preset-env), but I am unable to do this in my `postcss.config.js` configuration file due to how **PostCSS Loader** is currently configured.

My workaround has been to detect and delete the existing **PostCSS Loader** `plugins` option, which then allows the configuration file to be followed. However, I think it would be better if the zero-configuration (i.e Autoprefixer) is preserved _unless_ a configuration file explicitly defines its own plugins.

No, this PR does not introduce a breaking change.

**Other information**

This change adds a `postcss-load-config` dependency, but it should de-dupe along with the other nested `postcss-load-config` dependency. See these results for `npm ls postcss-load-config`:

```
preact-cli@3.0.0-rc.2 /preactjs/preact-cli/packages/cli
├── postcss-load-config@2.1.0 
└─┬ postcss-loader@3.0.0
  └── postcss-load-config@2.1.0  deduped
```

### DIFF
--- a/packages/cli/lib/lib/webpack/webpack-base-config.js
+++ b/packages/cli/lib/lib/webpack/webpack-base-config.js
@@ -11,6 +11,7 @@ const ProgressBarPlugin = require('progress-bar-webpack-plugin');
 const ReplacePlugin = require('webpack-plugin-replace');
 const ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin');
 const createBabelConfig = require('../babel-config');
+const loadPostcssConfig = require('postcss-load-config');
 
 function readJson(file) {
 	try {
@@ -83,6 +84,14 @@ module.exports = function(env) {
 	);
 
 	let tsconfig = resolveTsconfig(cwd, isProd);
+
+	let postcssPlugins;
+
+	try {
+		postcssPlugins = loadPostcssConfig.sync(cwd).plugins;
+	} catch (error) {
+		postcssPlugins = [autoprefixer({ overrideBrowserslist: browsers })];
+	}
 
 	return {
 		context: src,
@@ -208,7 +217,7 @@ module.exports = function(env) {
 							options: {
 								ident: 'postcss',
 								sourceMap: true,
-								plugins: [autoprefixer({ overrideBrowserslist: browsers })],
+								plugins: postcssPlugins,
 							},
 						},
 					],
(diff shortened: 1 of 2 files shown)


jonathantneal:
I see some positive feedback in the form of comment reactions. However, is this PR under consideration by the Preact CLI team?

I am hopeful, as my similarly themed PRs were merged — #829 to support Browserslist configuration files, and #831 to support pcss files.

packages/cli/lib/lib/webpack/webpack-base-config.js
```diff
@@ -84,6 +85,14 @@ module.exports = function(env) {
 
 	let tsconfig = resolveTsconfig(cwd, isProd);
 
+	let postcssPlugins;
+
+	try {
+		postcssPlugins = loadPostcssConfig.sync(cwd).plugins;
```
ForsakenHarmony: sounds good to me, just wondering if we should load more than just the plugins?

jonathantneal: Awesome!

Okay, forgive the long response, but here are my thoughts on loading more than just the plugins from a PostCSS configuration.

The [other options](https://github.com/michael-ciniawsky/postcss-load-config#options) are `to`, `map`, `from`, `parser`, `syntax`, and `stringifier`.

I am confident that `to` and `from` can be ignored because webpack determines those locations. I am somewhat confident that `parser`, `syntax`, and `stringifier` can be ignored at this time, because `css-loader` and `style-loader` can still choke on non-CSS syntaxes.

I am on the fence about `map`. On the one hand, it can be ignored, because source map configuration needs to be unified across all CSS loaders in order to be useful. However, if PostCSS configuration enables or disables CSS source maps for a project, then that does sound useful. The question is whether PostCSS configuration is the most logical place to put control over CSS source maps.

PR #835, opened 2019-07-13
Fix hard to read cli label

marvinhagemeister:

This PR fixes a usability issue where the cli output was hard to read due to very low contrast. Previously we had a dark gray font on a blue background which is too close together on the brightness scale.

Also added a space around the label.

not needed

**Summary**

It's best illustrated with a few screenshots. I'm using the default `Breeze` color scheme that ships with KDE.

Before:

preact-cli-ascii-before

After:

preact-cli-ascii-after

no


marvinhagemeister:
I don't have a windows machine to test on, so I'm not sure how it looks there :S

PR #839, opened 2019-07-15
Inline pre-render data

prateekbh:

Bugfix + Feature

Yes

**Summary**

#740 added the capability to pass custom props to the route components so that custom data or data from external sources can be pre-rendered.

However it lacked handling things on client side
Although the data was passed as props and was pre-rendered. The same props go missing on client side and thus preact removes the data. This PR solves it by inlining that data and making it available to the root node.

This data can be used to 
- hydrate a store
- pass data to routes to show them conditionally in UI

No


packages/async-loader/async.js
```diff
@@ -29,7 +38,13 @@ export default function(req) {
 			return !nxt;
 		};
 
-		this.render = (p, s) => (s.child ? h(s.child, p) : old);
+		this.render = (p, s) => {
+			if (!s.urlData || s.urlData.url !== window.location.pathname) {
```
prateekbh: This doesn't match `/profile/john` to `/profile/john/` which `preact-router` does.
@developit is there an export from `preact-router` to validate if it would consider 2 given URLs as same?

ForsakenHarmony:
I'd much rather do this at the root level and not add magic to the async component

prateekbh:
@ForsakenHarmony do you mean something like `render(document.body, <App/>, {...prerenderProps});`?

ForsakenHarmony:
Why not? We render the default export from index.js already, adding the parsing to the entry shouldn't be too hard

packages/cli/lib/pre-render-data-context.js
```diff
@@ -0,0 +1,26 @@
+import { Component } from 'preact';
+import { createContext } from 'preact-compat';
```
ForsakenHarmony: this will break X?

ForsakenHarmony: also force includes compat I guess?

prateekbh: Force includes only the context bits of it. Hoping tree shaking is still a thing.

Also I guess this is solvable if we release this as a separate package and ask our users to implement it instead of us doing it magically. 
Or if it sounds like too much work then this should just be a PreactX only feature.

WDYT?

prateekbh: no u're right increases the base bundle by 13kb ungzip and 4kb gzip...🙀

prateekbh:
After some thought I think we should just inline the data and leave it for the user to decide what to do.

Also we should create a separate lerna sub package `@preact-cli/prerender-data-provider` and let user use it so that 

a.) 2 different versions of `@preact-cli/prerender-data-provider` can address both preact 8 and preact X
b.) User does not need to take the weight of this package if they aren't using custom pre-render data.

packages/cli/lib/pre-render-data-context.js
prateekbh: If we move everything to preact X this delta comes down to 0.3 kb but then every one will be required to be on X and preact 8 wont work

packages/cli/lib/lib/app.js
```diff
@@ -0,0 +1,9 @@
+import { h } from 'preact';
+import { Provider } from '../../pre-render-data';
+
+const interopDefault = m => (m && m.default ? m.default : m);
+
+export default ({ CLI_DATA: { preRenderData } = {}, ...props }) => {
+	let app = interopDefault(require('preact-cli-entrypoint'));
+	return <Provider value={preRenderData}>{h(app, props)}</Provider>;
```
ForsakenHarmony: I think we still want to pass the prerender data to the root component myb?

also this still includes compat?

prateekbh: Yes if we this magic `<Provider.....>` compat will be included.
I am in favor of removing the whole `<Provider value=...>` from CLI code and give it as a package to the user to use it

ForsakenHarmony: sounds good

packages/cli/lib/lib/webpack/prerender.js
```diff
@@ -30,7 +30,6 @@ module.exports = function(env, params) {
 		const renderToString = require(require.resolve(
 			`${cwd}/node_modules/preact-render-to-string`
 		));
-
 		return renderToString(preact.h(app, { ...params, url }));
```
ForsakenHarmony: I think the behaviour differs between client side in the entry above and prerender here

prateekbh: params has everything coming from https://github.com/preactjs/preact-cli/pull/839/files#diff-b26e9c1b1185752601692f2207489852R17

prateekbh: I wish this was typescript 🙃.

ForsakenHarmony: oh, ok

cortopy:
I've just opened https://github.com/preactjs/preact-cli/issues/844 and then I stumbled upon this PR. I would also like the possibility of custom plugins. However, I think a prerender plugin would be more appropriate since data is not the only part involved in the process. I'd love to hear your thoughts on the idea

PR #841, opened 2019-07-16
feat: alias compat for preact X

ForsakenHarmony:


prateekbh:
do we need compat for X?

ForsakenHarmony:
wdym?

prateekbh:
i thought all the stuff in compat are now directly in `preactX` so we can alias them directly to `preact`

ForsakenHarmony:
no? 👀

prateekbh:
ohhh i might be completely wrong here

ForsakenHarmony:
it's now in the main package imported as `preact/compat`, but it's still a thing to keep core small

PR #842, opened 2019-07-17
Update README.md links

jonathantneal:

This change fixes broken links in **README.md**.

**README.md** badges now display correctly:
   1. `developit` becomes `preactjs` to reference the correct organization
   2. `preact-cli.svg?branch=master` becomes `preact-cli/master.svg` to match shields.io documentation
   3. `?style=flat` is omitted to match shields.io documentation (flat is the default style)
   4. **Build Status** badge uses `shields.io` to match the other badges

**README.md** reference links now use the right URLs:
   1. `developit` becomes `preactjs` to reference the correct organization
      - Note: https://github.com/developit/unfetch is correctly preserved :smile:
   2. `preact-cli/blob/master/src/lib` becomes `preact-cli/blob/master/packages/cli/lib` to correctly reference the new, lerna project structure
  3. `async!` link uses `1.4.1` tag to better date the information

No, this does not add tests, because these are documentation updates.

No, this does not introduce a breaking change, because these are documentation updates.


ForsakenHarmony:
Thanks 👌

PR #843, opened 2019-07-19
Service worker tweaks

prateekbh:
- Makes the documents go to network first
- Disable scaffolding service worker until we have preact api for service worker

No

**Summary**

With this change service worker response for documents only comes in if the user is offline completely.
Otherwise you always grab the fresh content from the network.

Yes


prateekbh:
Yeah, it's not really a silver bullet. 
Its should make a lot of people happy by making web pages more like web pages and not native apps.
But there are certainly some cases where people would want the earlier behaviour, for them we'll add preact cli API in a minor release after 3.0

marvinhagemeister:
Sounds good to me :100: :+1:

PR #848, opened 2019-07-24
Prerender provider

prateekbh:
This PR adds support for retrieving pre-render data for every route. If the user directly land on the route its, pre-render data is provided synchronously via the inlined script tag.
On client side navigation to another route its pre-render data is then fetched and provided the same way.

The PR provides a render prop component(for class based routes) and a hook to achieve the same.

**API**

in app.js
```js
import { Provider } from '@preact/prerender-data-provider';
...

export default class App extends Component {
	render(props) {
		return (
			<Provider value={props}>
				<div id="app">
					<Router url={props.url} onChange={this.handleRoute} {...props}>
						<Home path="/" />
						<Route66 path="/route66" />
					</Router>
				</div>
			</Provider>
		);
	}
}
```

1. hooks
```js
import { usePrerenderData } from '@preact/prerender-data-provider';
const Custom = props => {
	const [value, isLoading, error] = usePrerenderData(props);
        if (isLoading) return <div>Loading</div>;
        if (error) return <div>error</div>;
	return <div>{value.myProp}</div>;
};
```

2. Render Prop
```js
import { PreRenderDataSource } from '@preact/prerender-data-provider';
class Custom extends Component {
	render(props) {
		return (
			<PreRenderDataSource
				{...props}
				render={({ value, isLoading, error }) => {
                                           if (isLoading) return <div>Loading</div>;
                                           if (error) return <div>error</div>;
					return <div>{(value || {}).myProp}</div>;
				}}
			/>
		);
	}
}

```

Yes

**Summary**
There is no good way to consume pre-rendered data right now. 
This PR makes preact-cli a much more stronger static site generator as the user can now just get the data from its CMS and return it in `prerender-urls.js`. The same data would then be available for pre-rendering and on the client with the same API. 

So the user does not have to do any extra `if()... else()...` checks for data availability and fetch it manually if not available.

No

P.S.: This only works for components directly inside router.
e.g.
```jsx
<Router>
   <Home path...../> // <- Works only for these components because it exposes data only when the url is correct.
</Router>
```


StephanBijzitter:
This, plus the added support for async prerender urls already in v3, would be AMAZING

packages/prerender-data-provider/index.js
```diff
@@ -0,0 +1,41 @@
+import { Component } from 'preact';
+import { createContext } from 'preact-compat';
```
ForsakenHarmony: just really not happy about requiring compat to use prerender data

prateekbh: me too... dont know a better way though.
Thats y extracted it out in a different pckg

ForsakenHarmony: only reasonable way I see would be to require preact x, we have it in core there

prateekbh: Like only supported for preactX?

ForsakenHarmony: Yes

prateekbh: fixed.
added compatibility note in readme

prateekbh:
ummm @ForsakenHarmony any idea on whats going on with the test?
I upgraded preact/ render to string/ preact router

prateekbh:
Ok I need some help for this API.
Here's what the API currently does. Imagine a pre-render-urls.js

```js
export.default = [
  {url: '/', data: {description: 'homepage', author: 'john'}},
  {url: '/profile', data: {description: 'profilepage', author: 'jane'}}
]
```

Now when u land on `/` or `/profile` **directly** the `usePrerenderData` will give you the associated data synchronously.
But when you do a client side navigate to the other page, the pre-render data is not available for that.
Currently, the user is then expected to check if the data form `usePrerenderData` is null then make API call to get the data

e.g.

```jsx
function ProfilePage(props) {
  const [value] = usePrerenderData(props);
  if (!value) {
    // perform API fetch here
  }
  return (<div>{value.description}</div>)
}
```

**Question:**
Is this what you would expect because the pre-render data would only be inlined for the page you landed on? Or you would want a mechanism to fetch the data for other page's pre-render data as well

e.g.
```
function ProfilePage(props) {
  const [preRenderData, getPreRenderData] = usePrerenderData(props);
  if (!value) {
    getPreRenderData();
  }
  return (<div>{value.description}</div>)
}
```

ForsakenHarmony:
I guess it depends, if you have an api only getting the data for the current page would be fine

If you want to deploy it as a static page maybe not, we could output json files in that case myb

prateekbh:
Considering that the pre-render data work is to support more of static site generator scenario, do you think it'd be more intuitive to just give a function `getPreRenderData(url)` to the user which could fetch the data for that next page as well?

prateekbh:
@ForsakenHarmony @developit I have updated the PR description to reflect the current state of the PR.

prateekbh:
❤️❤️❤️

PR #850, opened 2019-07-25
Adding an offline test

prateekbh:

Adds tests

Yes

**Summary**
Adds a test that our service workers actually work.

No


prateekbh:
I have no idea what i did there, but the tests seems to be flaky now

prateekbh:
So a single test of `watch.test.js` works

prateekbh:
Also is fine without `service-worker.test.js`

packages/cli/tests/service-worker.test.js
```diff
@@ -0,0 +1,67 @@
...
+		dir = join(dir, 'build');
+		server = getServer(dir);
+	});
+
+	afterAll(async () => {
+		await server.server.stop();
+		await browser.close();
+	});
+
+	it('builds the default output', async () => {
```
ForsakenHarmony: I guess this name isn't right 👀

prateekbh: lol fixing it

Issue #844, opened 2019-07-22
Custom render-html-plugin method


cortopy:
**Do you want to request a _feature_ or report a _bug_?**
Request a feature

**What is the current behaviour?**
Preact-cli uses its own [html-render-plugin](https://github.com/preactjs/preact-cli/blob/master/packages/cli/lib/lib/webpack/render-html-plugin.js) and it's very difficult to extend. For example, in [L54](https://github.com/preactjs/preact-cli/blob/master/packages/cli/lib/lib/webpack/render-html-plugin.js#L54) it seems like the prerender method can be customised. But how does one get to pass the configuration for it?

**If the current behaviour is a bug, please provide the steps to reproduce.**

**What is the expected behaviour?**

**If this is a feature request, what is motivation or use case for changing the behaviour?**

I would like to be able to extend `render-html-plugin` for my own purposes like:
- Using a css-in-js library that supports SSR
- Injecting data as initial state
etc.

**Please mention other relevant information.**

I would like to start a discussion on which would be the best way to accomplish this. Would a PR that aims to do the following look good?

- [ ] extract `render-html-plugin` as a separate package in the preact-cli monorepo (and use this plugin as default at this stage)
- [ ] as a next step, make `render-html-plugin` configurable by initialising it in `preact.config.js`. If not initialised then behaviour can be using default settings maybe
- [ ] alternative plugins can also be initialised in `preact.config.js`, in which case `render-html-plugin` is not used

Please paste the results of `preact info` here.

cortopy:
Ah!! I've just realised something like this is already discussed in https://github.com/preactjs/preact-cli/pull/839

prateekbh:
Hi!
Yes, the cases u listed seems like valid use cases. The approach also seems fine.
>alternative plugins can also be initialized in `preact.config.js`, in which case render-html-plugin is not used

Is the only part i am not sure about as the setting in the the `template.html` are the key to the performance promised by preact-cli

cortopy:
> Hi!
> Yes, the cases u listed seems like valid use cases. The approach also seems fine.
> 
> > alternative plugins can also be initialized in `preact.config.js`, in which case render-html-plugin is not used
> 
> Is the only part i am not sure about as the setting in the the `template.html` are the key to the performance promised by preact-cli

[...]

PR #855, opened 2019-07-29
extract render-html-plugin

cortopy:

It extracts `render-html-plugin` into a separate package. 

These changes do not include any change in the logic of the application. This PR limits itself to move existing files into separate packages. I have tested the changes by creating and building an app using this version.

The only change in codebase is the currying of the plugin with a config object and the introduction of a `find` logic for the template file.

If this PR looks ok to reviewers, I'm happy to add more tests.

**Summary**

The main goal for doing this is to allow more customisation of the prerender process as per https://github.com/preactjs/preact-cli/issues/844

For final users, this PR does not change much as it doesn't expose an easy way to configure the plugin. I think this would be better addressed in a future PR, since this one already introduces some changes that I would like to have reviewed first:

- Move the files related to the prerender process (in `util/webpack`) into a separate package. This would require publishing this package.
- These files depend on an `util` file which now becomes a shared dependency between the main `cli` package and `@preact/render-html-plugin`. For this reason, I have also created a `@preact/cli-util` package so that both can share the code. This would also require publishing the package

No

**Other information**


ForsakenHarmony:
I'm still not sure what exactly you'd like to be able to customise

cortopy:
> I'm still not sure what exactly you'd like to be able to customise

Instead of css modules, I use a CSS-in-JS [library](https://github.com/streamich/nano-css). As a result, I need the prerendering to output two strings, one for the HTML body and another for the styles.

I've started using my [forked solution](https://github.com/cortopy/preact-cli/blob/render-html/packages/cli/lib/lib/webpack/prerender.js). But anybody that uses js libraries for styles, regardless of which one, would have to do something similar. Otherwise the prerendered pages would have no styles at all.

This is just one example of why prerendering can't fit into one-size-for-all solution. I wouldn't expect preact-cli to cover for all possible scenarios. My case is about styles, but I'm pretty sure others will have other needs.

Extracting the process as a plugin is, in my opinion, the first step towards achieving customisation using the webpack `preact.config.js` file

ForsakenHarmony:
While I do think we should support your use case I think we might want to do so at a different point, maybe make the ssr function itself customisable

cortopy:
@ForsakenHarmony that's great to hear. The main point of this PR and the steps I tried to plan in https://github.com/preactjs/preact-cli/issues/844 are already aimed in that direction.
For what you're saying I understand that you would rather have a different approach to this PR? If so, could you please explain so that I can modify accordingly?

If this is something that the core team would prefer doing themselves, please also let me know so that I close this PR and carry on using my fork.

PR #856, opened 2019-08-02
Add missing properties in location mock

marvinhagemeister:
bugfix/feature (depending on the point of view)

yup

**Summary**

This PR adds all properties of the `Location` object and not just `href` and `pathname`.

no

**Other information**
The `url` module is available in node since basically forever. Tested with node 10 and 12 to be sure.


ForsakenHarmony:
🙏

developit:
This seems good, though I'm a tiny bit worried folks will rely on `location.hostname` even though it's unlikely to be correct.

marvinhagemeister:
It's still `null` on SSR like it should be. The difference is that the property exists.

Issue #858, opened 2019-08-04
No clear console option


krzepah:
Hello,

**Do you want to request a _feature_ or report a _bug_?**
Feature

**What is the current behaviour?**
Clears console after build (preact watch)

**What is the expected behaviour?**
Don't clear the console !

**If this is a feature request, what is motivation or use case for changing the behaviour?**

npm-run-all two servers at the same time (and on the same stdout) and have them to play nicely together.

**Please mention other relevant information.**

--no-clear ?

https://github.com/preactjs/preact-cli/blob/master/packages/cli/lib/lib/webpack/run-webpack.js#L49 👿

I couldn't try to put it to false tough, I didn't understand get how the build process worked and the software is 'piled when installed : /

Thanks for your time !

krzepah:
Thanks !

PR #859, opened 2019-08-05
Don't clear console option

harshitkumar31:
Enhancement - Issue #858

No
**Summary**
Adds a new option to disable the console clear after webpack build

No

**Other information**

### DIFF
--- a/packages/cli/lib/index.js
+++ b/packages/cli/lib/index.js
@@ -79,6 +79,7 @@ prog
 	.option('--src', 'Specify source directory', 'src')
 	.option('--cwd', 'A directory to use instead of $PWD', '.')
 	.option('--esm', 'Builds ES-2015 bundles for your code.', true)
+	.option('--no-clear', "Don't clear the console", false)
 	.option('--sw', 'Generate and attach a Service Worker', false)
 	.option('--rhl', 'Enable react hot loader', false)
 	.option('--json', 'Generate build stats for bundle analysis')
--- a/packages/cli/lib/index.js
+++ b/packages/cli/lib/index.js
@@ -79,7 +79,7 @@ prog
 	.option('--src', 'Specify source directory', 'src')
 	.option('--cwd', 'A directory to use instead of $PWD', '.')
 	.option('--esm', 'Builds ES-2015 bundles for your code.', true)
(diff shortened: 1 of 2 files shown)


packages/cli/lib/index.js
```diff
@@ -79,6 +79,7 @@ prog
 	.option('--src', 'Specify source directory', 'src')
 	.option('--cwd', 'A directory to use instead of $PWD', '.')
 	.option('--esm', 'Builds ES-2015 bundles for your code.', true)
+	.option('--no-clear', "Don't clear the console", false)
```
prateekbh: Should we rethink if we want to keep this `false`?
// @lukeed @ForsakenHarmony

harshitkumar31: Won't it be too cluttered for users who are just getting started?

lukeed: This doesn't make sense, actually. It should be `('--clear', '...', true)` and then people can disable it thru the `--no-clear` flag

harshitkumar31: @lukeed  Updated.

packages/cli/lib/lib/webpack/run-webpack.js
```diff
@@ -46,7 +46,7 @@ async function devBuild(env) {
 			let serverAddr = `${protocol}://${host}:${bold(port)}`;
 			let localIpAddr = `${protocol}://${ip.address()}:${bold(port)}`;
 
-			clear(true);
+			if (config['clear']) clear(true);
```
lukeed: Here `config` is the webpack config - we want the CLI argv which comes down as `env` here.
```suggestion
			if (env['clear']) clear(true);
```

packages/cli/lib/index.js
prateekbh: isnt it how most CLI's out there work today?

lukeed: They could, yes, but it makes no sense _for us_ because our display format has been to show the `default: true` version -- looking at `esm` directly above.

ForsakenHarmony: I don't necessarily agree with `--clear, true` as long as `--no-clear` doesn't get automatically documented by sade 👀

People get confused all the time

PR #860, opened 2019-08-06
Fix Promise polyfill not working in IE11

marvinhagemeister:
bugfix

No, our tooling is not set up to run on other browsers than Chrome.

**Summary**

A bad import path for the `Promise` polyfill leads to an exception where the polyfill is used. Discovered this while work on fleshing out support for IE11 on [preact-www](https://github.com/preactjs/preact-www/pull/376). Instead of the expected `Promise` function we'd set `global.Promise` to the module object:

```js
// promise-polyfill.js
mode.exports = {
  default: function Promise() {}
};

// polyfills.js
global.Promise = require('promise-polyfill');
console.log(global.Promise);
// Logs: { default: [function Promise] }
```

So instead the `require` path call should be:

```js
global.Promise = require('promise-polyfill').default;
```

no

**Other information**
n/a

### DIFF
--- a/packages/cli/lib/lib/webpack/polyfills.js
+++ b/packages/cli/lib/lib/webpack/polyfills.js
@@ -1,2 +1,2 @@
-if (!global.Promise) global.Promise = require('promise-polyfill');
+if (!global.Promise) global.Promise = require('promise-polyfill').default;
 if (!global.fetch) global.fetch = require('isomorphic-unfetch');


developit:
This was caused by [promise-polyfill 8.0.0](https://github.com/taylorhakes/promise-polyfill/blob/master/CHANGELOG.md#800), which we upgraded to around 6 months ago.

PR #862, opened 2019-08-07
Simplify preact-cli template injection

marvinhagemeister:
feature

yup

**Summary**

This PR moves all custom the webpack html injection we use into easy to use variables. I'm hoping that this prevents issues like https://github.com/preactjs/preact/issues/1841 and allows us to make changes to our template without introducing breaking changes for users. With this PR the user's template can be shortened to:

```html
<!DOCTYPE html>
<html lang="en">
	<head>
		<meta charset="utf-8" />
		<title><% preact.title %></title>
		<% preact.headEnd %>
	</head>
	<body>
		<h1>Guess what</h1>
		<% preact.bodyEnd %>
	</body>
</html>
```

no

### DIFF
--- a/packages/cli/lib/lib/webpack/render-html-plugin.js
+++ b/packages/cli/lib/lib/webpack/render-html-plugin.js
@@ -1,11 +1,13 @@
-const { resolve } = require('path');
-const { existsSync, readFileSync, writeFileSync } = require('fs');
+const { resolve, join } = require('path');
+const os = require('os');
+const { existsSync, readFileSync, writeFileSync, mkdtempSync } = require('fs');
 const HtmlWebpackExcludeAssetsPlugin = require('html-webpack-exclude-assets-plugin');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
 const prerender = require('./prerender');
 const createLoadManifest = require('./create-load-manifest');
 const { warn } = require('../../util');
 const { info } = require('../../util');
+
 let defaultTemplate = resolve(__dirname, '../../resources/template.html');
 
 function read(path) {
@@ -36,7 +38,8 @@ module.exports = async function(config) {
 
 		// Unfortunately html-webpack-plugin expects a true file,
 		// so we'll create a temporary one.
-		template = resolve(__dirname, 'template.tmp.ejs');
+		const tmpDir = mkdtempSync(join(os.tmpdir(), 'preact-cli'));
+		template = resolve(tmpDir, 'template.tmp.ejs');
 		writeFileSync(template, content);
 	}
 
--- a/packages/cli/lib/lib/webpack/render-html-plugin.js
+++ b/packages/cli/lib/lib/webpack/render-html-plugin.js
@@ -1,24 +1,50 @@
 const { resolve } = require('path');
-const { existsSync } = require('fs');
+const { existsSync, readFileSync, writeFileSync } = require('fs');
 const HtmlWebpackExcludeAssetsPlugin = require('html-webpack-exclude-assets-plugin');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
 const prerender = require('./prerender');
 const createLoadManifest = require('./create-load-manifest');
 const { warn } = require('../../util');
 const { info } = require('../../util');
-let template = resolve(__dirname, '../../resources/template.html');
+let defaultTemplate = resolve(__dirname, '../../resources/template.html');
+
+function read(path) {
+	return readFileSync(resolve(__dirname, path), 'utf-8');
+}
 
 module.exports = async function(config) {
 	const { cwd, dest, isProd, src } = config;
 	const inProjectTemplatePath = resolve(src, 'template.html');
+	let template = defaultTemplate;
 	if (existsSync(inProjectTemplatePath)) {
 		template = inProjectTemplatePath;
 	}
+
+	template = config.template || template;
+
+	let content = read(template);
+	if (/preact\.headEnd|preact\.bodyEnd/.test(content)) {
+		const headEnd = read('../../resources/head-end.ejs');
+		const bodyEnd = read('../../resources/body-end.ejs');
+		content = content
+			.replace(
+				/<%\s+preact\.title\s+%>/,
+				'<%= htmlWebpackPlugin.options.title %>'
+			)
+			.replace(/<%\s+preact\.headEnd\s+%>/, headEnd)
+			.replace(/<%\s+preact\.bodyEnd\s+%>/, bodyEnd);
+
+		// Unfortunately html-webpack-plugin expects a true file,
+		// so we'll create a temporary one.
+		template = resolve(__dirname, 'template.tmp.ejs');
+		writeFileSync(template, content);
+	}
+
 	const htmlWebpackConfig = values => {
 		const { url, title, ...routeData } = values;
 		return Object.assign(values, {
 			filename: resolve(dest, url.substring(1), 'index.html'),
-			template: `!!ejs-loader!${config.template || template}`,
+			template: `!!ejs-loader!${template}`,
 			minify: isProd && {
 				collapseWhitespace: true,
 				removeScriptTypeAttributes: true,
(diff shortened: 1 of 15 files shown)


prateekbh:
This is what we've been wanting to do for so long. Thanks

packages/cli/lib/resources/body-end.ejs
```diff
@@ -0,0 +1,18 @@
+<%= htmlWebpackPlugin.options.ssr() %>
```
ForsakenHarmony: wonder if we could just do this in js instead of the template now

marvinhagemeister: I'm not sure if this is possible as `ejs` would escape anything we put in the directly.

ForsakenHarmony: I meant not using `ejs` at all

marvinhagemeister: I'm a little worried that switching to no template engine may cause us more trouble than it's worth. We still have a bit of logic I'm both the `headEnd` and `bodyEnd` template that involves loops and if-statements. Many of those values are filled out by `HtmlWebpackPlugin` at runtime. I'm not sure how that would work, but I may be missing something.

prateekbh:
Can we break
`preact.body` and `preact.script` differently?
This'll help us introduce changes to differential bundling in future without a breaking change

packages/cli/lib/lib/webpack/render-html-plugin.js
```diff
@@ -1,24 +1,50 @@
...
 
 module.exports = async function(config) {
 	const { cwd, dest, isProd, src } = config;
 	const inProjectTemplatePath = resolve(src, 'template.html');
+	let template = defaultTemplate;
 	if (existsSync(inProjectTemplatePath)) {
 		template = inProjectTemplatePath;
 	}
+
+	template = config.template || template;
```
prateekbh: Config template is `--template` option?
If yes I guess we should remove that and just stick to template.html in `src` because we scaffold that now

packages/cli/lib/lib/webpack/render-html-plugin.js
```diff
@@ -1,24 +1,50 @@
...
+			.replace(
+				/<%\s+preact\.title\s+%>/,
+				'<%= htmlWebpackPlugin.options.title %>'
+			)
+			.replace(/<%\s+preact\.headEnd\s+%>/, headEnd)
+			.replace(/<%\s+preact\.bodyEnd\s+%>/, bodyEnd);
+
+		// Unfortunately html-webpack-plugin expects a true file,
+		// so we'll create a temporary one.
+		template = resolve(__dirname, 'template.tmp.ejs');
```
prateekbh: Can we make this in `temp` dir? So that this unnecessary detail is abstracted away

packages/cli/lib/lib/webpack/render-html-plugin.js
marvinhagemeister: I'm not sure I understand, it's the same logic as before. Do you want to remove the cli flag completely and hardcode `<project>/src/template.html`?

packages/cli/lib/lib/webpack/render-html-plugin.js
marvinhagemeister: Can't find any references to a `temp` directory in the current code in `master`. Should we create another directory in the users project folder?

packages/cli/lib/resources/body-end.ejs
```diff
@@ -0,0 +1,18 @@
...
+	%>
+	<script nomodule src="<%= htmlWebpackPlugin.files.chunks["polyfills"].entry %>"></script>
+	<script nomodule defer src="<%= htmlWebpackPlugin.files.chunks['bundle'].entry %>"></script>
+<% } else { %>
+	<script <%= htmlWebpackPlugin.options.scriptLoading %>  src="<%= htmlWebpackPlugin.files.chunks['bundle'].entry %>"></script>
+	<script nomodule src="<%= htmlWebpackPlugin.files.chunks["polyfills"].entry %>"></script>
+<% } %>
+<script type="__PREACT_CLI_DATA__">
+	<%= JSON.stringify(htmlWebpackPlugin.options.CLI_DATA) %>
+</script>
```
ForsakenHarmony: btw can we move this script block up, webpack seems to do some weird stuff with the last script block trying to figure out what the origin is

had an error because this one doesn't have a source

packages/cli/lib/lib/webpack/render-html-plugin.js
prateekbh: Yeah like why to have two ways of doing the exact same thing? i.e. "custom template"
you want a custom template? just edit `template.html` is src. WDYT?

// @ForsakenHarmony

packages/cli/lib/lib/webpack/render-html-plugin.js
prateekbh: I was thinking may be the OS's `temp` dir.

packages/cli/lib/lib/webpack/render-html-plugin.js
ForsakenHarmony: but we can't have it break completely when someone deletes it in the project dir

prateekbh: if they delete from there project i guess we start fallback to use the one in `preact-cli` repo

prateekbh: https://github.com/preactjs/preact-cli/pull/862/files#diff-b26e9c1b1185752601692f2207489852R18-R21

ForsakenHarmony: so we in theory have to check on every build / watch

prateekbh: check on every build? yes.
check on **start** of every watch i guess.

ForsakenHarmony: yeah

packages/cli/lib/lib/webpack/render-html-plugin.js
prateekbh: https://www.npmjs.com/package/tmp

packages/cli/lib/lib/webpack/render-html-plugin.js
prateekbh: So do we have a consensus of deleting `--template`?

marvinhagemeister: I think in either case that'd be something that should be addressed in a separate PR. This one is about changing the template itself and not the template resolution strategy.

If we settle on removing the `--template` argument we need to make sure to document it somehwere.

packages/cli/lib/lib/webpack/render-html-plugin.js
marvinhagemeister: TIL, that node has a native `mkdTemp` function for that :tada: 

https://nodejs.org/api/fs.html#fs_fs_mkdtempsync_prefix_options

packages/cli/lib/resources/body-end.ejs
marvinhagemeister: done :tada:

marvinhagemeister:
> Can we break
> `preact.body` and `preact.scripts` differently?
> This'll help us introduce changes to differential bundling in future without a breaking change

This PR introduces a `preact.headEnd` and a `preact.bodyEnd` variable which inserts the content at the end of the respective tag. For that I took the code in the existing template as is and split it accordingly. At the end the template that's passed to `HtmlWebpackPlugin` is the exact same as in `master` (+ the script ordering suggestion @ForsakenHarmony made in a comment).

packages/cli/lib/lib/webpack/render-html-plugin.js
ForsakenHarmony: it may not be entirely ideal to create a new temporary directory for every build 🤔

marvinhagemeister: Good call :+1: Updated the PR to always reuse the same directory. Let me know what you think :100:

PR #864, opened 2019-08-08
Rename jest-preact-preset -> jest-preset-preact

marvinhagemeister:
Note to self: After this https://github.com/preactjs-templates/default/pull/27 needs to be updated too.


ForsakenHarmony:
🙈

PR #870, opened 2019-08-16
Fix and improve "create" questions

DenysVuika:
bugfix + feature

No, manual tests only

**Summary**

For the time being, running `preact create` without additional parameters causes the tool to crash because of the bug in the input types for user questions.

This PR introduces the following changes:

- upgrade to latest version of `prompts`
- fix the input types
- handle situation when user cancels the questions (`Ctrl+C`), do not continue

Screenshot 2019-08-17 at 20 41 43

- move the question generator closer to the "create" command code (was in the "setup" previously but used only for "create")
- introduce the template selector

Screenshot 2019-08-16 at 10 13 25

- using `custom` template shows extra question to provide the value
- ask about  package manager (yarn/npm) only if `Install dependencies` selected
- confirm the directory overwrite only when it already exists, abort the application correctly

Screenshot 2019-08-17 at 20 40 48

No

**Other information**

### DIFF
--- a/packages/cli/lib/commands/create.js
+++ b/packages/cli/lib/commands/create.js
@@ -8,25 +8,114 @@ const { resolve, join } = require('path');
 const { prompt } = require('prompts');
 const isValidName = require('validate-npm-package-name');
 const { info, isDir, hasCommand, error, trim, warn } = require('../util');
-const { addScripts, install, initGit, isMissing } = require('../lib/setup');
+const { addScripts, install, initGit } = require('../lib/setup');
 
 const ORG = 'preactjs-templates';
 const RGX = /\.(woff2?|ttf|eot|jpe?g|ico|png|gif|webp|mp4|mov|ogg|webm)(\?.*)?$/i;
 const isMedia = str => RGX.test(str);
 const capitalize = str => str.charAt(0).toUpperCase() + str.substring(1);
 
+// Formulate Questions if `create` args are missing
+function requestParams(argv) {
+	return [
+		// Required data
+		{
+			type: argv.template ? null : 'select',
+			name: 'template',
+			message: 'Pick a template',
+			choices: [
+				{
+					value: 'default',
+					title: 'Default (JavaScript)',
+					description: 'Default template with all features',
+				},
+				{
+					value: 'typescript',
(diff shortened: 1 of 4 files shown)


packages/cli/lib/commands/create.js
```diff
@@ -8,27 +8,116 @@ const { resolve, join } = require('path');
...
+			message: 'Directory to create the app',
+		},
+		// Extra data / flags
+		{
+			type: argv.name ? null : 'text',
+			name: 'name',
+			message: 'The name of your application',
+		},
+		{
+			type: argv.force ? null : 'confirm',
```
ForsakenHarmony: should this only prompt if the directory exists?

DenysVuika: That was not in the original code but I can add the check

DenysVuika: added more cases for the question flow, see the PR description

DenysVuika:
@ForsakenHarmony thanks for the review. Does this ticket need anything from me to progress, or it will be merged later?

ForsakenHarmony:
wanted someone else to do a review, but I think it's good, so I'll just merge it now, thanks 🎉

packages/cli/lib/commands/create.js
```diff
@@ -8,27 +8,138 @@ const { resolve, join } = require('path');
...
 const { info, isDir, hasCommand, error, trim, warn } = require('../util');
-const { addScripts, install, initGit, isMissing } = require('../lib/setup');
+const { addScripts, install, initGit } = require('../lib/setup');
 
 const ORG = 'preactjs-templates';
 const RGX = /\.(woff2?|ttf|eot|jpe?g|ico|png|gif|webp|mp4|mov|ogg|webm)(\?.*)?$/i;
 const isMedia = str => RGX.test(str);
 const capitalize = str => str.charAt(0).toUpperCase() + str.substring(1);
 
+function directoryExists(workingDir, destDir) {
```
prateekbh: nit: this should be in `../util`

DenysVuika:
@prateekbh I am working on additional contribution related to CSS preprocessor selection, will move that function to utils there if you don’t mind

PR #873, opened 2019-08-20
removing sw-precache-webpack-plugin

prateekbh:
- cleaning dep as we've moved to workbox


ForsakenHarmony:
I noticed this at some point, but forgot again 🙃

PR #878, opened 2019-08-27
Fixing template path issue

### LINKED ISSUE #876: Build process fails to locate custom template
Bug

Upgraded to version 3.0.0-rc.5 and build command with custom template file fails. Uninstalled and returning to version 3.0.0-rc.4 resolves the problem

The build command I was using is: `preact build --template src/template/defaultindex.html`

Build succeeds

Error on build: 
× ERROR Error: ENOENT: no such file or directory, open 'C:\Users\....\node_modules\preact-cli\lib\lib\webpack\src\template\defaultindex.html'

Environment Info:

  System:
    OS: Windows 10
    CPU: (4) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
  Binaries:
    Node: 10.15.2 - C:\Program Files\nodejs\node.EXE
    npm: 6.4.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: 42.17134.1.0

harshitkumar31:
Fixes #876 

No
**Summary**
Resolving template path from argv

**Other information**

### DIFF
--- a/packages/cli/lib/lib/webpack/render-html-plugin.js
+++ b/packages/cli/lib/lib/webpack/render-html-plugin.js
@@ -22,7 +22,10 @@ module.exports = async function(config) {
 		template = inProjectTemplatePath;
 	}
 
-	template = config.template || template;
+	if (config.template) {
+		const templatePathFromArg = resolve(cwd, config.template);
+		if (existsSync(templatePathFromArg)) template = templatePathFromArg;
+	}
 
 	let content = read(template);
 	if (/preact\.headEnd|preact\.bodyEnd/.test(content)) {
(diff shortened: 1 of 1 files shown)


packages/cli/lib/lib/webpack/render-html-plugin.js
```diff
@@ -22,7 +22,10 @@ module.exports = async function(config) {
 		template = inProjectTemplatePath;
 	}
 
-	template = config.template || template;
+	if (config.template) {
+		const templatePathFromArg = resolve(cwd, config.template);
+		if (existsSync(templatePathFromArg)) template = templatePathFromArg;
```
ForsakenHarmony: I feel like if someone specified a custom template path and it doesn't exist we should at least warn

ForsakenHarmony:
Thanks 🙏

Issue #809, opened 2019-06-06
sw-debug.js script is registering even if --sw=false in development mode


Panya:
**Do you want to request a _feature_ or report a _bug_?**
bug

**What is the current behaviour?**
I'm using `preact watch --sw false` command but `sw-debug.js` script is still trying to register as service worker.

**What is the expected behaviour?**
no `sw-debug.js` service worker when `--sw false` is passed.

**Please mention other relevant information.**

- Node version 10.15.3
- npm version 6.4.1
- Operating system Mac OS 10.14.6
- CLI version 3.0.0-rc.2
- Browser Chrome 74

prateekbh:
it really a no-op service worker thats installed only during `preact watch` to help you tell where you can use what APIs of SW.

does that seem confusing?

Panya:
I prefer not to include the sw if sw=false.

prateekbh:
gotcha

ForsakenHarmony:
I guess we might want to add the option to watch as well

psabharwal123:
https://github.com/preactjs/preact-cli/pull/884

reznord:
This is fixed in #884

DenizBasgoren:
registers sw-debug.js when I build with `preact build --no-sw`

Haugen:
Same issue for me still. `sw-debug.js` is still included in the build folder when running `preact build --no-sw`. Using preact-cli version `3.0.1`.

psabharwal123:
strange, I dont see it being included in the build in our project

RonnyLV:
sw-debug.js still registering on Preact 3.0.3
![image](https://user-images.githubusercontent.com/3844817/100598296-344a3c00-3307-11eb-89b4-c48303d82405.png)

RonnyLV:
When do you plan to release this fix?

rschristian:
> 
> 
> When do you plan to release this fix?
> 
> ![image](https://user-images.githubusercontent.com/3844817/100600377-e2ef7c00-3309-11eb-9b18-b021adc0e5fc.png)

What fix are you referring to? 

Also, `--inline-css` and `--no-brotli` can be removed from your command. CSS is inlined by default, and brotli is disabled by default. The flags have no effect with your usage.

RonnyLV:
> > When do you plan to release this fix?
[...]

PR #884, opened 2019-09-08
Make preact-cli exit with status code 1 if there are errors during build

psabharwal123:
bugfix

no test needed

**Summary**
currently even if there are errors, build will return exit code 0 so CI systems mark the build as successful and proceed to deploy the broken code

no

**Other information**

  System:
    OS: macOS 10.14.3
    CPU: (8) x64 Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz
  Binaries:
    Node: 10.16.3 - /usr/local/bin/node
    Yarn: 1.17.3 - /usr/local/bin/yarn
    npm: 6.9.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 76.0.3809.132
    Firefox: 67.0.2
    Safari: 12.0.3
  npmGlobalPackages:
    preact-cli: 3.0.0-rc.4


psabharwal123:
@prateekbh @ForsakenHarmony

psabharwal123:
also fixed #809

packages/cli/lib/lib/entry.js
```diff
@@ -13,7 +13,7 @@ if (process.env.NODE_ENV === 'development') {
 		hotLoader.preact(interopDefault(require('preact')));
 	}
 	// only add a debug sw if webpack service worker is not requested.
-	if (!process.env.ADD_SW && 'serviceWorker' in navigator) {
+	if (process.env.ADD_SW === undefined && 'serviceWorker' in navigator) {
```
ForsakenHarmony: Does this now mean specifying the option at all will disable it?

psabharwal123: yes that was for https://github.com/preactjs/preact-cli/issues/809

ForsakenHarmony: Yeah but I don't think that would be very intuitive, maybe add a check for `ADD_SW === 'false'`?

psabharwal123: if we check for ADD_SW === 'false', then sw-debug will get added even with --no-sw flag passed in the run command

ForsakenHarmony: Keep the current check as well 🙃

psabharwal123: what should we do if condition ADD_SW === 'false' is true?

ForsakenHarmony:
```suggestion
	if ((!process.env.ADD_SW || process.env.ADD_SW === 'false') && 'serviceWorker' in navigator) {
```

myb

psabharwal123: the intent was that if someone passes --no-sw (process.env.ADD_SW is false or 'false') then sw-debug.js is not registered since user is intentionally disabling it and most likely understands the benefits of service worker. Our use case is that we are using workbox so would like to not add sw-debug.js when developing locally.

psabharwal123: @ForsakenHarmony please let me know if you have more q

psabharwal123:
@prateekbh @ForsakenHarmony please merge this for rc.6

PR #900, opened 2019-10-03
fixing hot-reloading broken due to esm

prateekbh:
Temporarily disables modules in dev mode to fix hot reloading

No


psabharwal123:
this will reintroduce https://github.com/preactjs/preact-cli/issues/721

prateekbh:
thanks for bringing this up, but I guess the webpack hot loader doesnt work in `script type=module`.
This takes away hto reloading for thousands. I guess a better fix would be to just disable in watch mode so you can still do `preact watch --esm`

prateekbh:
oh nvm this should have been fixed by #862

PR #908, opened 2019-10-29
Inline CSS fixes

prateekbh:
**To be merged after: https://github.com/GoogleChromeLabs/critters/pull/45**

Bug fix

Updated the existing tests

**Summary**
Critters only inlines the css from the css files actually present in the html.
This was the reason why home page's css was never inlined on http://cli-demo-next.preactjs.com
This PR adds route's css as well as bundle css

No

### DIFF
--- a/packages/cli/lib/lib/webpack/webpack-client-config.js
+++ b/packages/cli/lib/lib/webpack/webpack-client-config.js
@@ -287,6 +287,7 @@ function isProd(config) {
 			new CrittersPlugin({
 				pruneSource: false,
 				logLevel: 'silent',
+				additionalStylesheets: ['*.css'],
 			})
 		);
 	}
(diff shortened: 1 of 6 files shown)


prateekbh:
Ahhh problem!
We always pre-render home, which means for someone who points all routes to `/` there will always be extra css loaded via `route-home.chunk.css`

prateekbh:
Like deploy to surge and redirect * to /

ForsakenHarmony:
Oh you're talking about the backend always serving `/index.html`, not the route specific html files

prateekbh:
yep, in that case all routes will download an extra `route-home.chunk.xxxxx.css`

prateekbh:
// @developit: ^ request for additional stylesheets options in critters

prateekbh:
@developit @ForsakenHarmony can this be solved by skipping css code splitting for server build?

prateekbh:
also may be skip terser for server bundle for speeding up builds?

PR #911, opened 2019-11-03
Preact X: Zero-Mutation Hydration

developit:
This is a hacky WIP for fixing the automatic progressive hydration we do for code-splitted routes in CLI. With Preact 8, we could falsify `this.nextBase` to trick the diffing algorithm into reusing the prerendered DOM tree. In X we can't do that, so I've tried a few different approaches here to accomplish the same.

The goal, as indicated in the added test (currently failing), is to hydrate with 0 DOM mutations. This means we need to use the `hydrate()` method (in production, in development this would break), but also that the deferred rendering from `async-component` needs to also be a zero-mutation hydrate. Even as clever as our old hack was for Preact 8, we were never able to perform a mutation-free initial render - the first render was actually doing something equivalent to `node.innerHTML = node.innerHTML`, which is definitely not free.

Ideally, Preact's initial support for Suspense could make this all quite seamless. The Suspense variant is commented out in this PR, but it basically amounts to rendering a `Pending` component that just throws a Promise. That Promise resolves once the code-split is loaded and ready, at which time we _should_ (theoretically) be able to use `setState()` on the root of the not-yet-hydrated component to trigger the previously-cancelled hydration.  Unfortunately, doing this right now produces a duplicate DOM tree.

So to summarize: I think we should get this to the point where our technique is at least as decent as the previous innerHTML hack. I've left something along those lines here, but personally I'm not happy with the results. What was previously an effective solution in Preact 8 now causes full DOM tree construction in X since we don't allow DOM tree reuse (component caching) as 8 did.


prateekbh:
@developit can you open another PR with the suspense variant? I have no idea what to uncomment for the suspense variant to work

PR #921, opened 2019-12-04
FAKE progressive hydration

prateekbh:
- FAKES progressive hydration
- Does not let the SSR DOM disappear until preact is ready with a new one.
- Then removes the SSR DOM
- Let preact replace it with fresh DOM

In a nut shell no long White flashes on screen. but THIS IS STILL NOT PROGRESSIVE HYDRATION


packages/async-loader/async.js
```diff
@@ -1,36 +1,80 @@
...
-			});
-		};
+options[f] = function(vnode) {
+	/**
+	 *  3. now Pending is being unmount and its dom is being removed
+	 *  so we detach the dom from the vnode and empty its children so that Pending is unmount
+	 *  but the DOM is never actually removed.
+	 */
+	if (vnode.type === Pending) {
+		vnode.__e = null;
```
ForsakenHarmony: maybe we should have these as a map or just mapping vars

```js
const keys = {
	dom: '__e',
	children: '__k',
	...
};
```

```js
const dom = '__e';
const children = '__k';
...
```

packages/async-loader/async.js
```diff
@@ -1,36 +1,80 @@
-import { h, Component } from 'preact';
+// import { options, h, Component } from 'preact';
+import { h, Component, options, render } from 'preact';
 
-export default function(req) {
-	function Async() {
-		Component.call(this);
+const f = 'unmount';
```
ForsakenHarmony: and maybe a more descriptive name, we are not the minifier 🙃

packages/async-loader/async.js
developit: The constants would be nice since they'd get inlined.

packages/async-loader/async.js
```diff
@@ -1,36 +1,80 @@
...
+options.diffed = function(vnode) {
+	/**
+	 *  4. The route component is now contructed and its DOM will be appended to the browser's DOM.
+	 *  But right before it, we swap its newly contructed DOM with the DOM already present on the browser.
+	 */
+	if (
+		vnode.__e &&
+		vnode.__e.parentNode === null &&
+		vnode.__.__ &&
+		vnode.__.__.type.name === 'AsyncComponent'
```
developit: This might be removed by Terser. Could check for a property, or maybe even use the existing `vnode.__.__.type.preload` as a signal that it's AsyncComponent.

prateekbh: seems to be working fine tested with build command

packages/async-loader/async.js
prateekbh: done

packages/async-loader/async.js
prateekbh: done

packages/cli/package.json
```diff
@@ -54,9 +54,9 @@
     "node-sass": "^4.12.0",
     "p-retry": "^4.1.0",
     "polka": "^0.5.2",
-    "preact": "^10.0.0",
-    "preact-compat": "^3.18.5",
-    "preact-render-to-string": "^5.0.6",
+    "preact": "^10.0.4",
+    "preact-compat": "^3.19.0",
```
cristianbote: Is the old `compat` still needed? 🤔

prateekbh: It's there just so that ppl working with Preact 8 have one less thing to do

cristianbote: oh you are right! Got confused and was thinking this is the `async` package and since the new version won't work with `8.x` thought that we could remove it from there.

All good! 👍 thank you

packages/async-loader/async.js
```diff
@@ -1,36 +1,99 @@
...
-		Component.call(this);
+const DOM = '__e';
+const CHILDREN = '__k';
+const PARENT = '__';
+const UNMOUNT = 'unmount';
+const oldUnmountOpts = options[UNMOUNT];
+const oldDiffed = options.diffed;
+const IS_PRERENDERED =
+	document.querySelector('#app') &&
+	document.querySelector('#app').childElementCount;
```
ForsakenHarmony: should this get casted to a bool?

prateekbh: like `!!(document.querySelector('#app') &&document.querySelector('#app').childElementCount);`

prateekbh: I changed this

packages/async-loader/async.js
```diff
@@ -1,36 +1,113 @@
...
+			vnode[DOM] = hydrationNode;
+			// this hook in options is no more needed once hydration is done.
+			options.diffed = oldDiffed;
+		}
+		oldDiffed && oldDiffed(vnode);
+	};
+}
+
+function Pending() {
+	hydrationNode = appChildren.length > 0 ? appChildren[0] : null;
```
developit: Hmm - searching within the app's root node is going to fail in a bunch of cases, like if there are nodes between App and the Router. Maybe we could inject a prop during prerendering and then look for it during hydration?

prateekbh: can you give a small example of what this failing case would look like?
e.g. this can handle 
```jsx
<App>
  <header/>
  <nav />
  <Router>
    ...
  </Router>
</App>
```

developit:
```js
<App>
  <div>
    <header/>
    <nav />
    <Router>
      ...
    </Router>
  </div>
</App>
```

developit: Right now during prerendering we bypass `async-loader`, but we could instead inject a component that marks AsyncComponent boundaries :

```js
exports.pitch = function(req) {
	this.cacheable && this.cacheable();
	
	return `
		import Sync from '@preact/async-loader/sync';
		function load(cb) {
			cb( require(${stringifyRequest(this, '!!' + req)}) );
		}
		export default Sync(load);
	`;
};
```

```js
import { h, options } from 'preact';

var next, old = options.vnode;
options.vnode = function(vnode) {
  var x = 'type' in vnode;
  var tag = x ? vnode.type : vnode.nodeName;
  var props = x ? 'props' : 'attributes';
  if (next && typeof tag === 'string') {
    next = null;
    (vnode[props] || (vnode[props] = {}))['data-pacr'] = next;
  }
};

export default function sync(load, name) {
  var component;
  return function SyncComponent(props) {
    if (!component) component = load();
    next = name;
    return h(component, props);
  }
}
```

Then the client could pick up the root using `querySelector("[data-pacr]").localName`.

Ideally this wouldn't be necessary. I wonder if there's an options hook we could use that would fire just prior to rendering the AsyncComponent and include the candidate DOM node?  That was the thing I stumbled on when porting this from 8 to X - in 8, we had `component._nextBase`.

developit: I forgot to make the point in my comment, but the reason for the complexity is because technically `async-component`/`async-loader` can be used anywhere in the component tree. In CLI that's anywhere you do `import X from 'async!./x'`.

prateekbh:
As discussed in the last meeting of core team, we wanted to have a common annotation on `vnode` about the DOM element which it will effect. 
There is a potential use case for this in both `Suspense` land and `partial hydration`. While we polish this concept/make PRs/land it/ship it, we all wanted to ship CLI v3 out so that an easy well-lit path for ecosystem is available to use this project.

We can ship this using @developit 's last comment but it is not perfect and still cannot cover all use cases.
So, The partial hydration will work if there is only one `Async` boundary on the top level.
If there is more than one usage of `Async` component directly under `App` we will bail on partial hydration and a flash would be visible as it is today.

We'll also follow this by opening an issue for CLI v3 so that people know the exact limitation and the status of the solution.

Thought? // @developit @ForsakenHarmony @marvinhagemeister

packages/async-loader/async.js
prateekbh: Instead of holding everything back, should we temporarily drop support for `async!` and move ahead for automatic code splitting for `routes`?

Now that almost everyone is more used to Suspense anyways for code splitting?

developit: We could drop support for pre-rendering `async!` and switch it to just boot only on the client. That way we're not dropping the split point and causing a huge change in bundle structure, but we're not holding back on progressive hydration.

prateekbh: Nit: Following code doesn't work
```js
options.vnode = function(vnode) {
  var x = 'type' in vnode;
  var tag = x ? vnode.type : vnode.nodeName;
  var props = x ? 'props' : 'attributes';
  if (next && typeof tag === 'string') {
    next = null;
    (vnode[props] || (vnode[props] = {}))['data-pacr'] = next;
  }
};
```

because the vnodes are constructed bottom up, so `if (next && typeof tag === 'string') {` does not point to the actual AsyncBoundary  but the farthest child of it.

prateekbh:
this does not work even as a fake hydration because entry animations would just appear again

PR #952, opened 2020-01-25
fixing async loader for v3

prateekbh:
TO BE MERGED after: https://github.com/preactjs/preact/pull/2259

Bug fix

No

**Summary**
- Fixes hydration in V3 for preactX

No

### DIFF
--- a/packages/cli/lib/lib/entry.js
+++ b/packages/cli/lib/lib/entry.js
@@ -32,7 +32,8 @@ if (process.env.NODE_ENV === 'development') {
 let app = interopDefault(require('preact-cli-entrypoint'));
 
 if (typeof app === 'function') {
-	let root = document.body.firstElementChild;
+	let root =
+		document.getElementById('preact_root') || document.body.firstElementChild;
 
 	let init = () => {
 		let app = interopDefault(require('preact-cli-entrypoint'));
--- a/packages/cli/lib/lib/entry.js
+++ b/packages/cli/lib/lib/entry.js
@@ -1,6 +1,6 @@
 /* global __webpack_public_path__ */
 
-import { h, render } from 'preact';
+import { h, render, hydrate } from 'preact';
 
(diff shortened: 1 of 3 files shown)


packages/cli/lib/lib/entry.js
```diff
@@ -48,7 +48,11 @@ if (typeof app === 'function') {
 		 * to send other data like at some point in time.
 		 */
 		const CLI_DATA = { preRenderData };
-		root = render(h(app, { CLI_DATA }), document.body, root);
+		const doRender =
+			process.env.NODE_ENV !== 'production' || root.tagName !== 'script'
```
developit: wondering if we need to recommend folks use `id="app"` here. We could have L35 do:

```js
let root = document.getElementById('preact_root') || document.body.firstElementChild;
```

Thoughts? It would help us avoid browser extensions potentially changing the value of firstElementChild on us.

developit:
Looks perfect aside from the one question about assuming firstElementChild. I think we can move forward with this regardless.

packages/cli/lib/lib/entry.js
prateekbh: yes! I really really want to do this. This stream lines everything for us

prateekbh:
> Only other thing would be if we want to consider having something for Preact 8

I guess lets keep it simple `async-component` v3 will be for cli v3 and `async-component` < v3 will be for CLI v1&2

packages/cli/lib/lib/entry.js
reznord: totally agree with it. :) Always wanted to do this one

prateekbh: done

developit:
@prateekbh right, just currently CLI V3 supports Preact 8 and X at the same time.

ForsakenHarmony:
Does this handle route transitions as well now?

developit:
Ahh good point. This needs to only throw during hydration!

PR #930, opened 2019-12-18
fix: Update sass options

chobe:
**What kind of change does this PR introduce?**
fix bug with scss

**Did you add tests for your changes?**
No, they're broken

**Summary**
After installing the default template and following the instructions of the sass preprocessors, the compilation was broken.

```
✖ ERROR ./style/index.scss (../node_modules/preact-cli/lib/lib/webpack/proxy-loader.js??ref--6-0!./style/index.scss)
Module build failed (from ../node_modules/preact-cli/lib/lib/webpack/proxy-loader.js):
ValidationError: Invalid options object. Sass Loader has been initialised using an options object that does not match the API schema.
 - options has an unknown property 'includePaths'. These properties are valid:
   object { implementation?, sassOptions?, prependData?, sourceMap?, webpackImporter? }
    at validate (./node_modules/sass-loader/node_modules/schema-utils/dist/validate.js:85:11)
    at Object.loader (./node_modules/sass-loader/dist/index.js:36:28)
    at Object.proxyLoader (./node_modules/preact-cli/lib/lib/webpack/proxy-loader.js:26:16)
 @ ./style/index.scss 2:14-219 21:1-42:3 22:19-224
 @ ./index.js
 @ ../node_modules/preact-cli/lib/lib/entry.js
 @ multi ../node_modules/preact-cli/lib/lib/entry webpack-dev-server/client webpack/hot/dev-server
✖ ERROR BabelEsmPlugin: ./style/index.scss (../node_modules/preact-cli/lib/lib/webpack/proxy-loader.js??ref--6-0!./style/index.scss)
Module build failed (from ../node_modules/preact-cli/lib/lib/webpack/proxy-loader.js):
ValidationError: Invalid options object. Sass Loader has been initialised using an options object that does not match the API schema.
 - options has an unknown property 'includePaths'. These properties are valid:
   object { implementation?, sassOptions?, prependData?, sourceMap?, webpackImporter? }
    at validate (./node_modules/sass-loader/node_modules/schema-utils/dist/validate.js:85:11)
    at Object.loader (./node_modules/sass-loader/dist/index.js:36:28)
    at Object.proxyLoader (./node_modules/preact-cli/lib/lib/webpack/proxy-loader.js:26:16)
 @ ./style/index.scss 2:14-219 21:1-42:3 22:19-224
 @ ./index.js
 @ ../node_modules/preact-cli/lib/lib/entry.js
[...]

PR #964, opened 2020-02-05
Sass options #930

ForsakenHarmony:
This is #930


ForsakenHarmony:
Thanks for the contribution @chobe

PR #981, opened 2020-02-14
skiping duplicate css warning

prateekbh:
bugfix

no

**Summary**
- Removes unavoidable but non build size impacting warnings

No


ForsakenHarmony:
Just gotta hope this doesn't ignore some other output which might actually be an issue

prateekbh:
upgrading to webpack 5 will ensure that in this case the contents are identical before ignoring the warning

PR #1011, opened 2020-02-25
Hydrate bug fix

prateekbh:
- Only hydrates if pages are pre-rendered


Inviz:
Looks great! thanks for a quick solution

PR #1018, opened 2020-02-26
Check for existince of template.html

reznord:
If template.html exists in the custom template directory, do not replace it with the current exisitng template

Bugfix

Yes, one test is added in `create.test.js`

**Summary**
Whenever a user bootstraps with a custom template, we overrider the `template.html` with the default one. This PR checks if the file exists, if so, it doesn't override it.

Nope

**Other information**

```
Environment Info:

  System:
    OS: macOS 10.15.3
    CPU: (8) x64 Intel(R) Core(TM) i5-8259U CPU @ 2.30GHz
  Binaries:
    Node: 12.16.1 - ~/.nvm/versions/node/v12.16.1/bin/node
    Yarn: 1.22.0 - /usr/local/bin/yarn
    npm: 6.13.4 - ~/.nvm/versions/node/v12.16.1/bin/npm
  Browsers:
    Chrome: 80.0.3987.87
    Firefox: 73.0.1
    Safari: 13.0.5
```

### DIFF
--- a/packages/cli/tests/create.test.js
+++ b/packages/cli/tests/create.test.js
@@ -1,4 +1,5 @@
-const { relative } = require('path');
+const fs = require('fs');
+const { relative, resolve } = require('path');
 const { create } = require('./lib/cli');
 const { expand } = require('./lib/utils');
 const snapshots = require('./images/create');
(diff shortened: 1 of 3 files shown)


prateekbh:
can you add a test using the netlify template?

packages/cli/tests/create.test.js
```diff
@@ -19,6 +20,15 @@ describe('preact create', () => {
 		});
 	});
 
+	it(`Check if custom template is being used or not`, async () => {
```
prateekbh: nit: it(`should use template.html from the github repo`)

reznord: done

PR #1019, opened 2020-02-27
Hydrate bug fix

prateekbh:

bugfix

no

**Summary**
Makes sure hydrate is only used for correct URL

no

### DIFF
--- a/packages/cli/lib/lib/entry.js
+++ b/packages/cli/lib/lib/entry.js
@@ -58,7 +58,7 @@ if (typeof app === 'function') {
 			process.env.PRERENDER &&
 			process.env.NODE_ENV === 'production' &&
 			hydrate &&
-			currentURL === location.pathname;
+			currentURL === location.pathname + location.search;
 		const doRender = canHydrate ? hydrate : render;
 		root = doRender(h(app, { CLI_DATA }), document.body, root);
 	};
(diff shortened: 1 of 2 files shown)


packages/cli/lib/lib/entry.js
```diff
@@ -43,19 +45,21 @@ if (typeof app === 'function') {
...
-			process.env.NODE_ENV !== 'production' ||
-			!hydrate
-				? render
-				: hydrate;
+		const currentURL = preRenderData.url ? normalizeURL(preRenderData.url) : '';
+		const canHydrate =
+			process.env.PRERENDER &&
+			process.env.NODE_ENV === 'production' &&
+			hydrate &&
+			currentURL === location.pathname;
```
developit: wondering if we need to also include `location.search` here? (tbh I think we could make a case both ways)

```suggestion
			currentURL === location.pathname + location.search;
```

reznord: makes sense, or else folks users might complain saying that query params aren't being retained. 😅

reznord: Pushed the tiny change 😬

developit: We may want to check both?  I'm worry about this:

`https://preactjs.com/?lang=en` <-- should match the content for `"/"` but now won't.

PR #1035, opened 2020-03-06
Power template list from github API in `preact create` command

reznord:
Refactoring

Not required

**Summary**
Right now, we have hard coded the template choices for `create` command. With this PR, we fetch the listings from the github API (same API which we use for `preact list` command) and load the template options from static fallback values.

Nope

```
Environment Info:

  System:
    OS: macOS 10.15.3
    CPU: (8) x64 Intel(R) Core(TM) i5-8259U CPU @ 2.30GHz
  Binaries:
    Node: 12.16.1 - ~/.nvm/versions/node/v12.16.1/bin/node
    Yarn: 1.22.0 - /usr/local/bin/yarn
    npm: 6.13.4 - ~/.nvm/versions/node/v12.16.1/bin/npm
  Browsers:
    Chrome: 80.0.3987.132
    Firefox: 73.0.1
    Safari: 13.0.5
```

### DIFF
--- a/packages/cli/lib/commands/create.js
+++ b/packages/cli/lib/commands/create.js
@@ -108,6 +108,11 @@ async function fetchTemplates() {
 			await fs.writeFile(TEMPLATES_CACHE_FILENAME, JSON.stringify(repos, null, 2), 'utf-8');
 		}
 
+		// fetch the API response from cache file
+		const templatesFromCache = await fs.readFile(TEMPLATES_CACHE_FILENAME, 'utf8');
+		const parsedTemplates = JSON.parse(templatesFromCache);
+		const officialTemplates = normalizeTemplatesResponse(parsedTemplates || []);
+
 		templates = officialTemplates.concat(CUSTOM_TEMPLATE);
 	} catch (e) {
 		// in case github API fails to fetch the data, fallback to the hard coded listings
--- a/packages/cli/lib/commands/create.js
+++ b/packages/cli/lib/commands/create.js
@@ -95,6 +95,15 @@ function requestParams(argv, templates) {
 	];
 }
 
+async function updateTemplatesCache() {
+	const repos = await fetch(TEMPLATES_REPO_URL).then(r => r.json());
+	await fs.writeFile(
+		TEMPLATES_CACHE_FILENAME,
+		JSON.stringify(repos, null, 2),
+		'utf-8'
+	);
+}
+
 async function fetchTemplates() {
 	let templates = [];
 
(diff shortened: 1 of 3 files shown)


reznord:
I don't think there is a point in writing to the disk and then serving it from there.

It would make sense if we hit the API only once (when we don't have the data in cache). But if we take that approach, then the cache will never be updated.

I'd prefer leaving it as it is where we hit the API every time.

prateekbh:
Delayed startup time is what I am worried.
May be I am over thinking

packages/cli/lib/commands/create.js
```diff
@@ -119,10 +95,56 @@ function requestParams(argv) {
...
+			TEMPLATES_CACHE_FILENAME,
+			'utf8'
+		);
+		const parsedTemplates = JSON.parse(templatesFromCache);
+		const officialTemplates = normalizeTemplatesResponse(parsedTemplates || []);
+
+		templates = officialTemplates.concat(CUSTOM_TEMPLATE);
+	} catch (e) {
+		// in case github API fails to fetch the data, fallback to the hard coded listings
+		templates = FALLBACK_TEMPLATE_OPTIONS.concat(CUSTOM_TEMPLATE);
```
developit: fantastic, came here to suggest this.

packages/cli/lib/commands/create.js
```diff
@@ -119,10 +95,56 @@ function requestParams(argv) {
...
+			const repos = await fetch(TEMPLATES_REPO_URL).then(r => r.json());
+			await fs.writeFile(
+				TEMPLATES_CACHE_FILENAME,
+				JSON.stringify(repos, null, 2),
+				'utf-8'
+			);
+		}
+
+		// update the cache file without blocking the rest of the tasks.
+		fetch(TEMPLATES_REPO_URL)
```
prateekbh: Why a mix of .then and async in same function?

reznord: If we do `async` here then it will delay the initial bootup since it will stop the execution of the main thread.

The 2nd fetch is only for updating the cache file every time user runs `preact create`

reznord: Will move the caching logic to a separate function.

packages/cli/lib/commands/create.js
```diff
@@ -119,10 +95,57 @@ function requestParams(argv) {
 	];
 }
 
+async function updateTemplatesCache() {
+	const repos = await fetch(TEMPLATES_REPO_URL).then(r => r.json());
+	await fs.writeFile(
+		TEMPLATES_CACHE_FILENAME,
```
ForsakenHarmony: does this put it in the current directory?

packages/cli/lib/commands/create.js
```diff
@@ -119,10 +95,57 @@ function requestParams(argv) {
...
+			const repos = await fetch(TEMPLATES_REPO_URL).then(r => r.json());
+			await fs.writeFile(
+				TEMPLATES_CACHE_FILENAME,
+				JSON.stringify(repos, null, 2),
+				'utf-8'
+			);
+		}
+
+		// update the cache file without blocking the rest of the tasks.
+		updateTemplatesCache();
```
ForsakenHarmony: not sure how obvious this behavior is, also no error handling I suppose

PR #1038, opened 2020-03-09
Migrate to gh-actions and remove travis ci

JoviDeCroock:
## Performance

GH Action: 2 minutes work (uncached) - 1minute 20 (cached)
Travis spawns two processes which both take:  3min 10-20 (cached)

EDIT: consider making these checks required to merge in settings


prateekbh:
can we run these on multiple node versions like latest/lts and lts -1?

JoviDeCroock:
> can we run these on multiple node versions like latest/lts and lts -1?

Seems like the `node-action` does not support lts/latest as keywords so I added the specific versions

PR #1040, opened 2020-03-09
Reduce bundle size in esm-mode by 7.5kb by using native-url

JoviDeCroock:

A feature

No tests, can I test this?

**Summary**

We can reduce 7.5 kb by switching out the webpack url polyfill for [native-url](https://github.com/GoogleChromeLabs/native-url)

Note: The GH-actions aren't added to `branch-protection` yet so we can just merge without them running


ForsakenHarmony:
Thought we already did this before 🤔

lukeed:
What pieces of `url` are in use?

prateekbh:
cc: @janicklas-ralph

prateekbh:
Quick question: if we don't do this, do we ship some other  bigger polyfill via Babel?

lukeed:
Yes, webpack's

prateekbh:
@lukeed, always or when we use `URL`  in the application code?

lukeed:
When `'url'` is used

lukeed:
Although this seems to be the only place that it's used? https://github.com/preactjs/preact-cli/blob/86b530df20cd06baa890a0c9488bc504ad921db0/packages/cli/lib/lib/webpack/prerender.js#L5

That shouldnt be included in webpack bundle... not sure what's going on

janicklas-ralph:
No issues on `native-url` at the moment. 

NextJS has also been using it for sometime.

prateekbh:
@jovidecroock why with env.esm tho?

JoviDeCroock:
@prateekbh it should only be safe to use in modern mode

JoviDeCroock:
> @JoviDeCroock based on your last comment [#1040 (comment)](https://github.com/preactjs/preact-cli/pull/1040#issuecomment-598118172), this has to be placed here: https://github.com/preactjs/preact-cli/blob/master/packages/cli/lib/lib/webpack/webpack-client-config.js#L154

Can we change stuff for webpack through that babel-plugin? That would feel kinda odd, I could check for that env variable as well in webpack?

prateekbh:
^ its just a hook if u wanna make extra changes in the child compilation. Usually changing definitions  upfront would leak them to parent compiler.

JoviDeCroock:
Hey @prateekbh I just noticed this is available on every browser https://caniuse.com/#feat=mdn-api_url

PR #1049, opened 2020-03-12
Implementing the minimal sw API

### LINKED ISSUE #967: Service worker improvements
Feature

To make additions in service worker, users currently have to 
1. copy paste the whole [sw.js](https://github.com/preactjs/preact-cli/blob/master/packages/cli/lib/lib/sw.js) in their `src` folder.
2. find apt place to make the change according to their need
3. make changes amid the 70 lines of code that we have.

1. Ugly details of basic service worker setup should be hidden from the users.
2. Users should be able to import a prebuilt module e.g. `preact/sw` and initialize it in < 5 LOC.
3. Advanced users should be able to add functionality on a clean stale environment

We want to abstract away the basic service worker setup in a module and make improvements to it  while users can customize additional functionality on top of it.

Currently the option users have for modifications in `sw.js` is to copy the entire thing and make changes to it. This makes us loose all control over their service worker for future releases.

**P.S: We don't want to maintain a full fledged SW library. We just want to govern basic setup and be able to ship improvements/library changes to all the basic users without a breaking change in preact-cli. However, this means that it might still be a breaking change for some advanced users, who can pretty much eject from our `sw` setup anyways.**

prateekbh:
Feature

Already in place

**Summary**
Fixes #967
- [x] Introduce minimal API
- [x] Scaffold SW to user land
- [x] Use scaffolded SW

No


packages/cli/lib/lib/webpack/webpack-client-config.js
```diff
@@ -226,61 +225,36 @@ function isProd(config) {
...
-			prodConfig.plugins.push(
-				new InjectManifest({
-					swSrc: 'sw-esm.js',
-					include: [/^\/?index\.html$/, /\.esm.js$/, /\.css$/, /\.(png|jpg)$/],
-					precacheManifestFilename: 'precache-manifest.[manifestHash].esm.js',
-				})
-			);
 	}
 
 	if (config.sw) {
```
developit: Is this going to add a second `InjectManifest` plugin when `config.esm` is `true`?

packages/cli/sw/sw.js
```diff
@@ -0,0 +1,4 @@
+import { getFiles, setupPrecaching, setupRouting } from './index';
+
+setupRouting();
```
developit: I'm wondering if we're locking ourselves into too strict an API here with the two functions. Thoughts on this?

```
import { getFiles, init } from './index';

setup({
  files: getFiles(),
  // other options here
});
```

packages/cli/lib/lib/webpack/webpack-client-config.js
prateekbh: yes that is for the second service worker.

packages/cli/sw/sw.js
prateekbh: Coincidently that was literally my initial design.
The problem i found with this design was that `precaching` and `routing` are 2 very independent pieces.

The problem that we're trying to solve(have a large base of basic users which just use our API and we can change under lying implementation for them without a breaking change) fades away here because
- Pretty much everyone uses precaching
- We introduce a fallback page for error/offline cases
- We have a default strategy for navigation request(Network first currently)

If my app changes any of the above(e.g. a game or a music app can be/should be cache first or i might want to change my catch behavior to a "you are offline" page) then i'll need to do everything inside `init` again.

vs

In the currently implemented design, I can skip routing and just `precaching`

developit: TBH based on that I'm fine with leaving them separate.

packages/cli/lib/lib/webpack/webpack-client-config.js
```diff
@@ -172,7 +171,14 @@ function getBabelEsmPlugin(config) {
...
 function isProd(config) {
 	let limit = 200 * 1000; // 200kb
-
+	const { src } = config;
+	let swPath = join(__dirname, '..', '..', '..', 'sw', 'sw.js');
+	const userSwPath = join(src, 'sw.js');
+	if (existsSync(userSwPath)) {
+		swPath = userSwPath;
+	} else {
+		warn(`Could not find sw.js in ${src}. Using a fallback service worker.`);
```
developit: nit:

```suggestion
		warn(`Could not find sw.js in ${src}. Using the default service worker.`);
```

prateekbh:
BTW this removes the `brotli` plugin. But we can write up a small doc on how to do that manually or publish that class separately.

developit:
@prateekbh what's the reason for dropping the brotli plugin?

prateekbh:
@developit this is implemented to be extended with a plugin system
a follow pr/ a good first issue can look like
```js
import { getFiles, setupPrecaching, setupRouting } from 'preact-cli/sw/';
import {brotliFileProcessing, PRECACHE_OPTIONS} from 'preact-cli/sw/BrotliPlugin';
setupRouting();
setupPrecaching(brotliFileProcessing(getFiles()), PRECACHE_OPTIONS);
```

packages/cli/lib/index.js
```diff
@@ -51,7 +51,7 @@ prog
 	)
 	.option('-c, --config', 'Path to custom CLI config', 'preact.config.js')
 	.option('--esm', 'Builds ES-2015 bundles for your code.', true)
-	.option('--brotli', 'Adds brotli redirects to the service worker.', false)
+	.option('--brotli', 'Builds brotli compressed bundles of javascript.', false)
```
reznord: For now, should we not remove this option?

packages/cli/lib/lib/webpack/webpack-client-config.js
```diff
@@ -172,7 +171,14 @@ function getBabelEsmPlugin(config) {
 
 function isProd(config) {
 	let limit = 200 * 1000; // 200kb
-
+	const { src } = config;
+	let swPath = join(__dirname, '..', '..', '..', 'sw', 'sw.js');
+	const userSwPath = join(src, 'sw.js');
```
reznord: Just curious, can we move the `sw.js` to be in the root folder (`/`) instead of `src`? (just a user preference I guess)

packages/cli/lib/index.js
prateekbh: I dont think so,  this still generates the bundles.  which can be used by people for their custom server setups

packages/cli/lib/lib/webpack/webpack-client-config.js
prateekbh: everything related to code should be inside `src` right?

packages/cli/lib/index.js
reznord: Yep true, forgot about that part.

PR #1051, opened 2020-03-14
Refactor list command

reznord:
refactoring

Not required.

**Summary**
Since the logic for fetching the template listings is the same for both `create` & `list` commands, migrated it to a helper utility and used it from there. Reduces the duplicacy of code. :P 

No


packages/cli/lib/template-listings-helper.js
```diff
@@ -0,0 +1,81 @@
...
+		os.homedir(),
+		TEMPLATES_CACHE_FOLDER,
+		TEMPLATES_CACHE_FILENAME
+	);
+
+	try {
+		const repos = await fetch(TEMPLATES_REPO_URL).then(r => r.json());
+		await fs.writeFile(cacheFilePath, JSON.stringify(repos, null, 2), 'utf-8');
+	} catch (err) {
+		error(`\nFailed to update template cache\n ${err}`);
```
prateekbh: so what happens when i am disconnected from the network?
console logs `Failed to update template cache`?

packages/cli/lib/template-listings-helper.js
```diff
@@ -0,0 +1,81 @@
...
+			const repos = await fetch(TEMPLATES_REPO_URL).then(r => r.json());
+			await fs.writeFile(
+				cacheFilePath,
+				JSON.stringify(repos, null, 2),
+				'utf-8'
+			);
+		}
+
+		// update the cache file without blocking the rest of the tasks.
+		updateTemplatesCache();
```
prateekbh: why not give the `repos` fetched in https://github.com/preactjs/preact-cli/pull/1051/files#diff-983d722f3055735a842ad26c7dcece61R58 as a param?

packages/cli/lib/template-listings-helper.js
```diff
@@ -0,0 +1,81 @@
...
+exports.fetchTemplates = async function() {
+	let templates = [];
+	const cacheFolder = join(os.homedir(), TEMPLATES_CACHE_FOLDER);
+	const cacheFilePath = join(
+		os.homedir(),
+		TEMPLATES_CACHE_FOLDER,
+		TEMPLATES_CACHE_FILENAME
+	);
+
+	try {
```
prateekbh: There are too many things that can fail in this `try...catch` forcing us to go to `FALLBACK_TEMPLATES`.
`FALLBACK_TEMPLATES` should only be rendered if there is no `cache` or it cannot be read. Otherwise always use `cache`

PR #1052, opened 2020-03-15
Support custom babelrc

reznord:
Feature

No (not sure how to test this)

**Summary**
Apps bootstrapped with CLI doesn't honour custom babel config as of now. This fixes the missing functionality from 2.x versions.

Nope


prateekbh:
Re Test: May be build an app with 0 routes and --no-esm and inspect if const/let classes are preserved in generated bundle.js

reznord:
> Do we want to make the `babelrc` location configurable?

Yep, cuz users normally tend to use `.babelrc` & `.babelrc.json`. Babal docs suggest us to use `babel.config.json` tho. So thought it would be better to make the location of `babelrc` configurable.

ForsakenHarmony:
wondering if it would be better if we tried to resolve more of the variants :woman_shrugging:

PR #1083, opened 2020-03-25
Updating babel deps

prateekbh:

Updates babel deps

Yes

No


reznord:
Now that we have updated babel to `v7.9.x`, should we turn on `bugfixes: true` in `preset-env` config? That enables us to use `preset-modules`.

prateekbh:
nice catch

prateekbh:
upgraded `babel-esm-plugin` which takes care of this for the `esm` files

PR #1116, opened 2020-04-06
Add "bugfixes: true" to preset-env in default Babel config

Saeris:
Seems like an easy win to get even smaller bundle sizes from @developit's recent addition to babel-preset-env as detailed here: https://babeljs.io/blog/2020/03/16/7.9.0

Checked package.json for the CLI package, it appears it's already been upgraded to the latest versions for babel.


prateekbh:
this will be enabled for  next release for esm files. do we want it enabled otherwise as well?
// @developit

rschristian:
Looks like this ended up being added in #1580 so I'm going to close this out

PR #1176, opened 2020-05-13
convert to @prefresh/webpack and use react-refresh/babel for hooks

JoviDeCroock:
Converts `preact-refresh` to use the new and improved `@prefresh/webpack`, this will also work for hooks and custom hooks since it uses `react-refresh/babel` to produce function signatures.


ForsakenHarmony:
Thanks for all your work on it

PR #1207, opened 2020-05-28
exclude react-refresh/babel for node_modules

JoviDeCroock:

Fixes a bug where prefresh would get confused because of node_modules (the loader isn't injected for it for perf reasons). This was added in 0.6 and makes prefresh work again, still trying to figure out the call-stack bug.

No


reznord:
LGTM 👍

PR #1230, opened 2020-06-09
Use `lessOptions`

### LINKED ISSUE #1173: Changing to less crashes the default build
_bug_
Creating a new default Project with preact-cli install less and less-loader. Change all .css to .less.
Start the dev server which cashes.

1. `preact create default less-test`
2. `npm install --save-dev less less-loader`
3. `npm run dev`
4. run dev should fail to build and start.

installing less on a clean project should work with the steps described in the docs

----

developit:
this one didn't require the property hiding from sass-loader (#964) since it's within the existing options object.

Should fix #1173.


developit:
@JoviDeCroock it'd be `options.lessOptions` there, and we delete `options`.

Issue #1226, opened 2020-06-08
pre-rendered data should be HTMLescaped


prateekbh:
**Do you want to request a _feature_ or report a _bug_?**
Bug

**What is the current behaviour?**
Any prerendered data is not HTMLEscaped.

**If the current behaviour is a bug, please provide the steps to reproduce.**
Add a `</script> and some test here` to the prerendered data

**What is the expected behaviour?**
- Pre-rendered data should be HTMLEscaped.
- And should be Unescaped before putting into context API

developit:
@prateekbh I've used `devalue` to achieve this in the past. Normally `devalue` produces executable JS, which we don't want because that will block the HTML parser. Passing the data through JSON.stringify() and then through devalue() works around this ([here](https://github.com/preactjs/preact-cli/blob/5dd567867c56510cbb38d7caad86003956d83cc6/packages/cli/lib/resources/body-end.ejs#L2-L4)):

```diff
<script type="__PREACT_CLI_DATA__">
- <%= JSON.stringify(htmlWebpackPlugin.options.CLI_DATA) %>
+ <%= JSON.parse(require('devalue')(JSON.stringify(htmlWebpackPlugin.options.CLI_DATA))) %>
</script>
```

PR #1232, opened 2020-06-09
use serialize-javascript to escape values

JoviDeCroock:

Yes

**Summary**

Resolves: https://github.com/preactjs/preact-cli/issues/1226


packages/cli/lib/lib/webpack/render-html-plugin.js
```diff
@@ -93,7 +94,8 @@ module.exports = async function(config) {
 				return config.prerender ? prerender({ cwd, dest, src }, values) : '';
 			},
 			scriptLoading: 'defer',
-			CLI_DATA: { preRenderData: { url, ...routeData } },
+			CLI_DATA: serialize({ preRenderData: { url, ...routeData } }),
```
prateekbh, line 97: Cant we use `encodeURI`/`decodeURI`?

packages/cli/lib/lib/webpack/render-html-plugin.js
```diff
@@ -148,9 +150,9 @@ module.exports = async function(config) {
 // Adds a preact_prerender_data in every folder so that the data could be fetched separately.
 class PrerenderDataExtractPlugin {
 	constructor(page) {
-		const { url } = page.CLI_DATA.preRenderData;
+		const { url } = page.UNSAFE_CLI_DATA.preRenderData;
```
prateekbh, line 153: As mentioned above just `decodeURI` the `page.CLI_DATA.preRenderData`

prateekbh:
Actually can we just `encodeURI` here: https://github.com/preactjs/preact-cli/blob/master/packages/cli/lib/resources/body-end.ejs#L3

and its consumer

packages/cli/lib/lib/webpack/render-html-plugin.js
prateekbh, line 153: can we `deserialize` here? so that there is only one data source.

packages/cli/lib/lib/webpack/render-html-plugin.js
```diff
@@ -148,9 +148,10 @@ module.exports = async function(config) {
 // Adds a preact_prerender_data in every folder so that the data could be fetched separately.
 class PrerenderDataExtractPlugin {
 	constructor(page) {
-		const { url } = page.CLI_DATA.preRenderData;
+		const cliData = page.CLI_DATA || {};
+		const { url } = cliData.preRenderData;
```
reznord, line 152:
```suggestion
		const { url } = cliData.preRenderData || {};
```

packages/cli/lib/lib/webpack/render-html-plugin.js
reznord: This is already addressed

PR #1235, opened 2020-06-10
Fix CSS preloading in Firefox

### LINKED ISSUE #1199: Async CSS loading is broken in Firefox 77

This is a bug. It seems to be resolved in Firefox nightly but at this moment it's broken on the v77 beta build of Firefox. It could be a beta issue, if someone could test on stable that'd be awesome. As a workaround I've just locked the version at `3.0.0-rc.9` in our projects. 

The stylesheet loaded via the new technique introduced in `3.0.0-rc.13` is broken on Firefox 77. 

Prior to `3.0.0-rc.13` stylesheets were loaded like this:

```html
<head>
  <link href="/bundle.css" rel="preload" as="style">
</head>
<body>
  ...
  <link rel="stylesheet" href="/bundle.css">
</body>
```

In `3.0.0-rc.13` they are now loaded with the following snippet:

```html
<link href="/bundle.css" rel="preload" as="style" onload="this.rel='stylesheet'" />
```

A test case is here:
https://jsfiddle.net/6gnwvfoe/

If the font stays as the default browser serif font, then you're experiencing the issue. Chrome and Safari seem fine. 

The stylesheet loads as normal.

developit:
This should fix #1199 - it reverts back to the old behavior, which was the critters default:

```html
<head>
  <link href="/bundle.css" rel="preload" as="style">
</head>
<body>
  ...
  <link rel="stylesheet" href="/bundle.css">
</body>
```


developit:
Ah, yeah indeed it would.

developit:
I wonder if `preload:"media"` would work. It does `<link rel=stylesheet media="not x" onload="this.media='all'" href="..css">`.

prateekbh:
removed myself from reviewers as i added a commit.

Checked on FF77 and chrome. works as expected

PR #1249, opened 2020-06-18
bump @prefresh/webpack

JoviDeCroock:
The changes:

- Add export maps for Node 13/14
- Correctly return the function type from the refreshSigning, this enables support for `memo` and `forwardRef`
- Change `window` to `self` so prefresh works outside of the browser (example web worker)


reznord:
Lock file isn't updated

Issue #1248, opened 2020-06-17
Watch mode without webpack's dev server


oliverstr:
**Do you want to request a _feature_ or report a _bug_?**
Feature

**What is the current behaviour?**
When developing a custom server for SSR, the process of building the ssr-bundle is a very bad experience.
The `build` command can take a long time and there's no watch mode, meaning every time there's a code change the developer needs to kill the server, build the entire application, and run the server again.

**If the current behaviour is a bug, please provide the steps to reproduce.**

**What is the expected behaviour?**
`build` command could have a watch mode option

**If this is a feature request, what is motivation or use case for changing the behaviour?**
This would greatly benefit developers that have custom servers and want a hot reload option.

**Please mention other relevant information.**
I can create a PR for this.

Please paste the results of `preact info` here.

RyanChristian4427:
As `build` is meant to create a production build (coming from the ReadMe), would it be better to instead add an option to the `watch` command in some capacity? Adding a watch feature to the build command would mean the entire production build/optimization process would be done on every change to the source, no?

oliverstr:
> As `build` is meant to create a production build (coming from the ReadMe), would it be better to instead add an option to the `watch` command in some capacity? Adding a watch feature to the build command would mean the entire production build/optimization process would be done on every change to the source, no?
> 
> Regardless, would be nice, I agree.

Makes total sense, was thinking of changing the env to development on the build command but I guess that will get messy.

Cool, should then add an option in the `watch` that will:
* disable dev server
* add watch mode to webpack
* build the ssr-bundle

Wdyt?

RyanChristian4427:
That certainly sounds great. 

Granted, I have absolutely no idea how much work that would be to add as I haven't even glanced at the source for prerendering, but it sounds like it would be a much improved experience, no doubt.

oliverstr:
@RyanChristian4427 I've got the PR ready, how can I push a new branch to the repo?

RyanChristian4427:
To this repo? Unless I'm mistaken, you can't merge a fork into a new branch on the fork's origin.

If you want to merge into a branch rather than master for some reason, you'll need one of the maintainers to add one for you.

oliverstr:
@RyanChristian4427 https://github.com/preactjs/preact-cli/pull/1264

gabrielrtakeda:
is there any progress...? this feature would be great.

RyanChristian4427:
[...]

PR #1264, opened 2020-06-29
Watch mode with no dev server

oliverstr:
Feature

No

**Summary**

https://github.com/preactjs/preact-cli/issues/1248

No

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,7 +141,10 @@ $ preact watch
 
     --src           Specify source directory  (default src)
     --cwd           A directory to use instead of $PWD  (default .)
-    --esm           Builds ES-2015 bundles for your code.  (default true)
+    --devServer     Determine if dev server should be enabled  (default true)
+    --dest          Specify output directory if dev server is disabled  (default build)
+    --ignore        Path relative to src to be ignored during watch if dev server is disabled
+	  --esm           Builds ES-2015 bundles for your code.  (default true)
     --sw            Generate and attach a Service Worker  (default false)
     --json          Generate build stats for bundle analysis
     --https         Run server with HTTPS protocol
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -141,7 +141,10 @@ $ preact watch
 
     --src           Specify source directory  (default src)
     --cwd           A directory to use instead of $PWD  (default .)
-    --esm           Builds ES-2015 bundles for your code.  (default true)
+    --devServer     Determine if dev server should be enabled  (default true)
+    --dest          Specify output directory if dev server is disabled  (default build)
+    --ignore        Path relative to src to be ignored during watch if dev server is disabled
+	  --esm           Builds ES-2015 bundles for your code.  (default true)
```
RyanChristian4427, line 147: nitpick: The slightly different indentation here is quite noticeable on the rendered file

packages/cli/lib/index.js
```diff
@@ -82,6 +82,17 @@ prog
...
+	.option('--devServer', 'Determine if dev server should be enabled', true)
+	.option(
+		'--dest',
+		'Specify output directory if dev server is disabled',
+		'build'
+	)
+	.option(
+		'--ignore',
+		'Path relative to src to be ignored during watch if dev server is disabled',
+		''
```
RyanChristian4427, line 94: This default value means that `src/` is ignored by default. I don't think this is intended?

RyanChristian4427:
The user is unable to import from style sheet without causing error if they have not wrapped their imports in `if (typeof window !== 'undefined') { ... }` which is not ideal. Don't know what other people's opinions are, maybe that's fine to have as a requirement?

packages/cli/lib/index.js
developit:
```suggestion
		null
```

PR #1308, opened 2020-07-08
(async-loader): Use reference in vnode for current node

reznord:
Use the vnode to get access the DOM node (for getting the current element).

@developit 's magic 😄 

Fixes #1293

Issue #1293, opened 2020-07-06
Cannot import async components: Undefined component passed to createElement()


hkh12:
**Do you want to request a _feature_ or report a _bug_?**
It's a bug.

**What is the current behaviour?**
I'm trying to load a component from `src/components/async` directory, and get this error:
```
Uncaught Error: Undefined component passed to createElement()

You likely forgot to export your component or might have mixed up default and named imports<#text dangerouslySetInnerHTML="[object Object]" />

  in #text
  in AsyncComponent
  in App
```

**If the current behaviour is a bug, please provide the steps to reproduce.**
```js
// src/components/async/Hello.js
export default function Hello() {
  return <h1>hello</h1>
}
```
```js
// src/index.js
import Hello from './components/async/Hello'
export default function App() {
  return <Hello />
}
```

**What is the expected behaviour?**
Load the component `Hello` as an async component.

**Please mention other relevant information.**

Please paste the results of `preact info` here.
```
Environment Info:
  System:
    OS: Linux 5.4 Ubuntu 20.04 LTS (Focal Fossa)
    CPU: (2) x64 AMD A9-9410 RADEON R5, 5 COMPUTE CORES 2C+3G
  Binaries:
    Node: 14.2.0 - /usr/local/bin/node
    Yarn: 1.7.0 - ~/.npm-global/bin/yarn
    npm: 6.14.4 - /usr/local/bin/npm
  Browsers:
    Chrome: 83.0.4103.116
    Firefox: 77.0.1
  npmPackages:
    preact: ^10.3.2 => 10.3.3 
    preact-cli: ^3.0.0-rc.16 => 3.0.0-rc.16 
  npmGlobalPackages:
    preact-cli: 3.0.0-rc.16
```

hkh12:
Update: `async!` prefix doesn't work too.

RyanChristian4427:
I assume that, while this warning shows up in your console, the component does render? As that is what I can reproduce. 

One solution could be to use Lazy & Suspense. [While marked as experimental](https://preactjs.com/guide/v10/switching-to-preact#suspense), it does seem to suppress that warning.

```
import { Suspense, lazy } from 'preact/compat';
const Hello = lazy(() => import('./components/async/Hello'));

export default function App() {
    return (
        <Suspense fallback={<div>loading...</div>}>
            <Hello />
        </Suspense>
    )
}
```

hkh12:
[...]

PR #1309, opened 2020-07-09
Fix async-loader for Preact X

developit:
I finally (FINALLY) found a solid working solution for async hydration (#1293, #1308, etc etc).

Unlike our various earlier attempts, this solution doesn't rely on DOM hacks, pointers or guessing. It crawls up the tree of VNodes, performing a Depth First Search of their children, working backwards from the position of AsyncComponent in order to find the most recently hydrated DOM element. Essentially, the first populated `vnode._dom` reference it finds is guaranteed to be the previous sibling of the parent element AsyncComponent is about to get rendered into. As long as AsyncComponent isn't being used load a component that renders a Fragment, the nextSibling of that discovered element is the exact node Preact will try to hydrate.

You can think of this algorithm as "find the last attached DOM node _before_ me in the virtual DOM tree within my nearest element parent".

In the case of an AsyncComponent wrapping a Fragment, only the first Fragment child will be hydrated. This means the remaining children will have the flicker effect, but that's pretty reasonable:

```js
// app.js
import Foo from 'async!./foo';
export default () => <div><Foo /></div>

// foo.js
export default () => <Fragment>
    <h1>Hello</h1>
    <p>world!</p>  // will be culled+reinserted during hydration
</Fragment>
```

### DIFF
--- a/packages/async-loader/async.js
+++ b/packages/async-loader/async.js
@@ -2,8 +2,28 @@ import { h, Component } from 'preact';
 
 const PENDING = {};
 
+// Given a VNode, finds its previous element sibling.
+function getPreviousSibling(vnode) {
+	const parent = vnode.__;
+	if (!parent) return;
+	let children = parent.__k;
+	if (children) {
+		if (!Array.isArray(children)) children = [children];
+		// only search previous children
+		let end = children.indexOf(vnode);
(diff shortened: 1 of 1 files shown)


prateekbh:
I tried adding a test case using CLS but it doesnt work as i hoped it would

PR #1328, opened 2020-07-16
GH-591 fix sw being loaded from CDN

### LINKED ISSUE #591: Service Worker loaded from CDN
Report a Bug

`sw.js` is loaded from a CDN if a CDN is set in the webpack `publicPath`. This is forbidden:
`Failed to register a ServiceWorker: The origin of the provided scriptURL ('https://yourcdn.net') does not match the current origin ('https://yourapp.com').`

Add the following snippet to `preact.config.js`:

    config.output.publicPath = '//youcdn.net/';

It seems this was introduced in https://github.com/developit/preact-cli/pull/323

`sw.js` must be loaded from the base domain. See https://github.com/w3c/ServiceWorker/issues/940

- Node version: v6.10.3
- npm version: 3.10.10
- Operating system: macOS
- CLI version: 2.2.1
- Browser: Chrome 67.0.3396.87

kencanak:
Bug Fix, this will close #591
- refactored entry.js to use `process.env.SW_PATH` instead of webpack public path
- introduced new cli option for setting swPath, default would be `/`

N/A, did a local test though

**Summary**
Service Worker loaded from CDN, see #591 for more details

don't think so

### DIFF
--- a/packages/cli/lib/lib/entry.js
+++ b/packages/cli/lib/lib/entry.js
@@ -1,5 +1,3 @@
-/* global __webpack_public_path__ */
-
 import * as Preact from 'preact';
 const { h, render, hydrate } = Preact;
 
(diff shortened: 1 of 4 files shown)


prateekbh:
I am still thinking if adding one more flag is the right thing to do here or can we handle it differently

packages/cli/lib/lib/entry.js
```diff
@@ -14,17 +12,17 @@ if (process.env.NODE_ENV === 'development') {
 	// only add a debug sw if webpack service worker is not requested.
 	if (process.env.ADD_SW === undefined && 'serviceWorker' in navigator) {
 		// eslint-disable-next-line no-undef
-		navigator.serviceWorker.register(__webpack_public_path__ + 'sw-debug.js');
+		navigator.serviceWorker.register(process.env.SW_PATH + 'sw-debug.js');
 	} else if (process.env.ADD_SW && 'serviceWorker' in navigator) {
 		// eslint-disable-next-line no-undef
 		navigator.serviceWorker.register(
-			__webpack_public_path__ + (process.env.ES_BUILD ? 'sw-esm.js' : 'sw.js')
+			process.env.SW_PATH + (process.env.ES_BUILD ? 'sw-esm.js' : 'sw.js')
```
prateekbh, line 19: Why cant we 
```js
// Service workers are not allowed to be loaded from external domain
navigator.serviceWorker.register(__webpack_public_path__.startsWith('https://')?'':__webpack_public_path__) + (process.env.ES_BUILD ? 'sw-esm.js' :'sw.js')
```

developit:
Can't this just always use `__webpack_public_path__`? It defaults to `"/"`.

prateekbh:
I guess people would sometimes like to build the app for a subfolder? like `/shop`?

packages/cli/lib/lib/entry.js
kencanak: I guess, I just thought that we shouldn't assume that it's going to be served from root and give some flexibility for setting the sw path.

kencanak: @prateekbh , removed the flag, and added a check for both http + https protocol.

Issue #1218, opened 2020-06-02
CSS-error when building 


smurfb:
**Do you want to request a _feature_ or report a _bug_?**
bug
**What is the current behaviour?**

C:\Users\joel\IdeaProjects\justshop-www>preact build
‼ WARN Could not find sw.js in C:\Users\joel\IdeaProjects\justshop-www\src. Using the default service worker.
 Build  [=================== ] 93% (3.7s) after chunk asset optimization
       bundle.98920.css ⏤  0 B (-3.77 kB)
    bundle.*****.esm.js ⏤  0 B (-32.9 kB)
 polyfills.*****.esm.js ⏤  0 B (-2.01 kB)
                  sw.js ⏤  0 B (-8.48 kB)
              sw-esm.js ⏤  0 B (-8.48 kB)
        bundle.67a42.js ⏤  0 B (-33.5 kB)
     polyfills.ac517.js ⏤  0 B (-2.01 kB)
   ssr-bundle.ea496.css ⏤  4.76 kB (+4.76 kB)
          ssr-bundle.js ⏤  29.3 kB (+29.3 kB)

 Build  [=================== ] 93% (12.3s) after chunk asset optimization
   ssr-bundle.ea496.css ⏤  0 B (-4.76 kB)
          ssr-bundle.js ⏤  0 B (-29.3 kB)
       bundle.fec83.css ⏤  3.77 kB (+3.77 kB)
    bundle.*****.esm.js ⏤  32.9 kB (+32.9 kB)
 polyfills.*****.esm.js ⏤  2 kB (+2 kB)
              sw-esm.js ⏤  8.44 kB (+8.44 kB)
                  sw.js ⏤  8.44 kB (+8.44 kB)
        bundle.645db.js ⏤  33.5 kB (+33.5 kB)
     polyfills.d99ed.js ⏤  2 kB (+2 kB)

Error: Could not parse CSS stylesheet
    at exports.createStylesheet (C:\Users\joel\AppData\Roaming\npm\node_modules\preact-cli\node_modules\jsdom\lib\jsdom\living\helpers\stylesheets.js:34:21)
    at HTMLStyleElementImpl._updateAStyleBlock (C:\Users\joel\AppData\Roaming\npm\node_modules\preact-cli\node_modules\jsdom\lib\jsdom\living\nodes\HTMLStyleElement-impl.js:67:5)
[...]

PR #1332, opened 2020-07-20
Configure terser to not extract comments in production mode

vivlim:
Fixes this error message I was seeing during builds, which was preventing me from creating a production build.
```
× ERROR The comment file "route-dataset/index.tsx.chunk.0d50e.esm.js.LICENSE.txt" conflicts with an existing asset, this may lead to code corruption, please use a different name
```

No. I don't think this change needs a test, but let me know if I should add one (and what it would test)

**Summary**

Sets `extractComments: false` in webpack production config.

A breaking change in terser 3.0.0 (changelog quoted below) made this message into an error, which was picked up in #1221. This message showed up as a warning in a paste from #1218.
> if the value from the extractComments.filename option conflicts with existing assets, an error will be thrown instead of a warning

Perhaps this message indicates some problem with my project, because at one point it used to work, but I couldn't figure out what was wrong. It seems like this change would be consistent with `terserOptions.output.comments = false` but I only barely grok what that setting does.

This breaks comment extraction during production builds. I have no idea if that's an expected behavior or useful to someone.

**Other information**

I didn't actually clone or build this locally, I just patched this file in my node_modules and repeated that change in my fork.

### DIFF
--- a/packages/cli/lib/lib/webpack/webpack-client-config.js
+++ b/packages/cli/lib/lib/webpack/webpack-client-config.js
@@ -229,6 +229,7 @@ function isProd(config) {
 							],
 						},
 					},
+					extractComments: false,
 					sourceMap: true,
 				}),
 				new OptimizeCssAssetsPlugin({


rinatkhaziev:
I just have stumbled upon this too, for a bundle path though:

``` The comment file "bundle.995f8.esm.js.LICENSE.txt" conflicts with an existing asset, this may lead to code corruption, please use a different name```

**Comment by dbbk** [comment]
I've just spun up a new project and production builds fail for me out of the box with this error

```
✖ ERROR The comment file "bundle.9ec82.esm.js.LICENSE.txt" conflicts with an existing asset, this may lead to code corruption, please use a different name
```

**Comment by prateekbh** [comment]
can someone confirm if this still exist with `--no-esm`?

**Comment by rinatkhaziev** [comment]
@prateekbh `--no-esm` works as expected.

**Comment by rinatkhaziev** [comment]
If y'all just want a quick workaround until the underlying issue is addressed, adding this line in `preact.config`'s `webpack` method should do:

```
config.optimization.minimizer[0].options.extractComments = false;
```

**Comment by shalin1** [comment]
ahhh thanks!  preact was broken out of the box today, first time playing with it, and was wrestling with it for the last hour!  adding --esm to the preact build script worked great for me.

**Comment by prateekbh** [comment]
does any of your have a public repo to look at?

**Comment by shalin1** [comment]
> does any of you have a public repo to look at?

Sure, https://github.com/shalin1/shalin-dot-org is using the `--no-esm` fix until something gets updates in preact.  If you clone it down and revert the latest commit you should be able to reproduce.

File: packages/cli/lib/lib/webpack/webpack-client-config.js [hunk 1]
```diff
@@ -229,6 +229,7 @@ function isProd(config) {
 							],
 						},
 					},
+					extractComments: false,
 					sourceMap: true,
```
prateekbh, line 233: also adding `comments`: false would be the right thing to do

prateekbh: oh nvm it only ends up in .map files

prateekbh:
released in 3.0.1

PR #1336, opened 2020-07-20
Avoid file locks + overwrites by using a temp file for HTML templates

### LINKED ISSUE #1331: Running two builds in parallel?
Not sure if this is a bug or if this is simply not supported. I'm trying to run a few builds in parallel (building numerous projects in a mono repo). But I get

```
✖ ERROR Error: EEXIST: file already exists, mkdir '/tmp/preact-cli'
 
```

Is there a way to set which tmp folder is used? Or is running preact-cli in parallel simply not possible?

developit:
This fixes #1331.

Note: I just made this edit quickly, need to install [tempy](https://github.com/sindresorhus/tempy) and update the package lock.


vesahyp:
This would solve our problems of developers and gitlab CI running preact-cli on the same server.

packages/cli/src/lib/webpack/render-html-plugin.js
```diff
@@ -45,7 +45,10 @@ module.exports = async function (config) {
 
 		// Unfortunately html-webpack-plugin expects a true file,
 		// so we'll create a temporary one.
-		const tmpDir = join(os.tmpdir(), 'preact-cli');
+		const tmpDir = join(
+			os.tmpdir(),
+			`preact-cli-${Math.floor(Math.random() * 100000)}`
+		);
```
rschristian, line 51: I think `tempy` is rather overkill for our needs, especially with the number of dependencies it adds: http://npm.anvaka.com/#/view/2d/tempy

A random number between 0 and 100,000 should be sufficient enough to avoid conflicts I think.

Issue #1342, opened 2020-07-21
Install access issues 


internet-individual:
I sign-in as admin and and use the command `sudo npm install -g preact-cli@rc ` and there's an error. The install is on arch with the latest versions of node and npm. Other than that, are there are other react command line tools or frameworks that implement babel (etc) like that tool?

The error is:

```
Error: EACCES: permission denied, open '/usr/lib/node_modules/preact-cli/node_modules/nodent-runtime/dist/index.js'
    at Object.openSync (fs.js:465:3)
    at Object.writeFileSync (fs.js:1416:35)
    at Module._compile (internal/modules/cjs/loader.js:1201:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1221:10)
    at Module.load (internal/modules/cjs/loader.js:1050:32)
    at Function.Module._load (internal/modules/cjs/loader.js:938:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
    at internal/main/run_main_module.js:17:47 {
  errno: -13,
  syscall: 'open',
  code: 'EACCES',
  path: '/usr/lib/node_modules/preact-cli/node_modules/nodent-runtime/dist/index.js'
}
```

prateekbh:
can you do `sudo npm install -g preact-cli`. As of yesterday we've hit stable with v3

access-apps:
The issue isn't different with that command.

RyanChristian4427:
You'll need to add the `unsafe-perm` flag.

```
sudo npm install -g --unsafe-perm preact-cli
```

Should work for you. Same as #802

developit:
Just to clarify - why would you need to install this as root? Doing so is _super_ dangerous - it's downloading megabytes of arbitrary code from the internet and executing it in a root shell.

Issue #802, opened 2019-05-19
preact-cli@3.0.0-rc.1 fails to install in a fresh node container


dionjwa:
**Do you want to request a _feature_ or report a _bug_?**
_bug_

**What is the current behaviour?**
```
docker run --rm -ti node:8.16.0-alpine /bin/sh
# npm i --global preact-cli@3.0.0-rc.1
```
Shows:
```
npm WARN deprecated sw-precache@5.2.1: Please migrate to Workbox: https://developers.google.com/web/tools/workbox/guides/migrations/migrate-from-sw
npm WARN deprecated sw-toolbox@3.6.0: Please migrate to Workbox: https://developers.google.com/web/tools/workbox/guides/migrations/migrate-from-sw
/usr/local/bin/preact -> /usr/local/lib/node_modules/preact-cli/lib/index.js

> nodent-runtime@3.2.1 install /usr/local/lib/node_modules/preact-cli/node_modules/nodent-runtime
> node build.js

fs.js:646
  return binding.open(pathModule._makeLong(path), stringToFlags(flags), mode);
                 ^

Error: EACCES: permission denied, open '/usr/local/lib/node_modules/preact-cli/node_modules/nodent-runtime/dist/index.js'
    at Object.fs.openSync (fs.js:646:18)
    at Object.fs.writeFileSync (fs.js:1299:33)
    at Object.<anonymous> (/usr/local/lib/node_modules/preact-cli/node_modules/nodent-runtime/build.js:5:4)
    at Module._compile (module.js:653:30)
    at Object.Module._extensions..js (module.js:664:10)
    at Module.load (module.js:566:32)
    at tryModuleLoad (module.js:506:12)
    at Function.Module._load (module.js:498:3)
    at Function.Module.runMain (module.js:694:10)
    at startup (bootstrap_node.js:204:16)
npm WARN preact-cli@3.0.0-rc.1 requires a peer of preact@* but none is installed. You must install peer dependencies yourself.
npm WARN preact-cli@3.0.0-rc.1 requires a peer of preact-render-to-string@* but none is installed. You must install peer dependencies yourself.
npm WARN @preact/async-loader@3.0.0-rc.0 requires a peer of preact@^8.1.0 but none is installed. You must install peer dependencies yourself.
npm WARN react-hot-loader@4.8.6 requires a peer of react@^15.0.0 || ^16.0.0 but none is installed. You must install peer dependencies yourself.
[...]

PR #1343, opened 2020-07-21
Docs(ReadMe): Adding note for installing CLI as root 

RyanChristian4427:

Adding documentation to cover (what might become) a common issue.

N/A

**Summary**

Now that the default install for the CLI is v3.0.0, more people will probably run into the same problem that is presented in #802 and #1342. While running NPM as root, the `unsafe-perm` flag will be set to `false` by default which causes an issue with (at least) `nodent-runtime`. [NPM docs](https://docs.npmjs.com/misc/config#unsafe-perm), [nodent-runtime](https://github.com/MatAtBread/nodent-runtime/issues/5). The simple fix is to add the `unsafe-perm` flag.

As running global installations as root is a pretty common thing to do, having documentation to catch the people who run into this issue is likely worthwhile. 

No breaking changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,14 @@
 $ npm install -g preact-cli
 ```
 
+> **Note**:  
+> If you are installing as root, you will need to add the `unsafe-perm` flag to your install command:
+>
+> ```sh
+> $ sudo npm install -g --unsafe-perm preact-cli
+> ```
+>
+
 ### Usage
 
 ```sh


developit:
I'm (very) hesitant to include anything about running as root in the readme. 99.99% of the time it's danger without justification.

RyanChristian4427:
That's certainly reasonable.

I only mention it as it is the default on non-Windows systems (assuming the dev does a simple Node/NPM install without a Node version manager) to need root for global installs. It's entirely anecdotal and without any numbers backing it, but I feel like I've seen rather large amount of libs/tools telling users to just use `sudo`. That sort of advice in addition to it "working" (*with potential security risks) has led me to believe the majority probably do just use `sudo` rather than a version manager or moving NPM's default directory. 

Could totally be wrong. Also might not be worth showing the easy way around EACCES issues and instead having them find the NPM documentation showing the proper way around instead.

teodragovic:
How about updating docs to use `npx preact create`?

RyanChristian4427:
Certainly could do.

CRA has been telling users to use npx over global installs for a while now, to ensure they're always up to date. CRA as a global dependency fills the same area that the Preact-CLI does, so maybe it's worth doing the same here?

Npx certainly is slower but that might be a better suggestion going forward. A bit slower of initializing a new project every now and then probably won't be noticeable for most users.

ForsakenHarmony:
> the majority probably do just use sudo rather than a version manager or moving NPM's default directory.

Honestly that's on npm, and they should really change the default directory

RyanChristian4427:
> 
> Honestly that's on npm, and they should really change the default directory

Oh definitely. However, until they do, people will probably keep using `sudo` for their global installs, which comes with some issues for build scripts that the Preact CLI employs (at least for now). 

This PR is just here in case you do want to catch the few people who will run into issues, as undoubtedly some percentage will. I could amend this to discourage `sudo` use, or perhaps this should just be closed without any additional documentation?

Not my repository by any means, so let me know if you'd like for me to change something or if this isn't what the team wants to encourage.

RyanChristian4427:
Closing as better alternatives have been pointed out, so I doubt this is going anywhere or worth keeping.

PR #1347, opened 2020-07-25
Remove fast-async

### LINKED ISSUE #1591: Cloud Firestore v9 build error with preact-cli
CLI throws error when `dev` or `build` script are used with the new Firebase SDK v9.

**Error output:**
```
$ npm run build

> webapp@0.0.0 build
> preact build

 Build  [==                  ] 12% (1.3s) building(13200:6) switch-case fall-through not supported - added break. See https://github.com/MatAtBread/nodent#differences-from-the-es7-specification
[BABEL] Note: The code generator has deoptimised the styling of ../node_modules/@firebase/firestore/dist/exp/index.browser.esm2017.js as it exceeds the max of 500KB.
✖ ERROR ../node_modules/@firebase/firestore/dist/exp/index.browser.esm2017.js 5055:431
Module parse failed: Unsyntactic break (5055:431)
File was processed with these loaders:
 * ../node_modules/babel-loader/lib/index.js
You may need an additional loader to handle the result of these loaders.
(Source code omitted for this binary file)
 @ ../node_modules/firebase/firestore/dist/index.esm.js 1:0-36 1:0-36
 @ ./components/app.tsx
 @ ./index.ts
```

Read more about this bug on SO: https://stackoverflow.com/questions/67350610/cloud-firestore-build-error-with-preact-cli.
Even though I'm not the creator of this SO question I see the exact same error.

**Steps to Reproduce**

1. Create a blank project with `preact-cli`
2. Follow the [instructions](https://firebase.google.com/docs/web/setup?sdk_version=v9#add-sdks-initialize) on how to add Firebase v9 to the project
3. Run `dev` or `build` scripts

Preact CLI should work smoothly with new version of Firebase especially if it aims to reduce bundle size.

Make sure to read comments in the mentioned SO question, especially the one from `DrYap`.

```
Environment Info:
  System:
    OS: macOS 11.4
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
  Binaries:
    Node: 16.6.2 - /var/folders/26/tlzqz_2d4cx3wr1f8w4ns_ww0000gp/T/yarn--1629174442990-0.9348698159252729/node
    Yarn: 1.22.10 - /var/folders/26/tlzqz_2d4cx3wr1f8w4ns_ww0000gp/T/yarn--1629174442990-0.9348698159252729/yarn
    npm: 7.20.3 - /usr/local/bin/npm
  Browsers:
    Chrome: 92.0.4515.131
    Safari: 14.1.1
  npmPackages:
    preact: ^10.3.1 => 10.5.14 
    preact-cli: ^3.0.0 => 3.2.2 
    preact-router: ^3.2.1 => 3.2.1 
```

teodragovic:

Remove `fast-async` from dependency list in favor of using `transform-async-to-generator`

Fixes #1591

No

**Summary**

Going off [this comment](https://github.com/preactjs/preact-cli/pull/1343#issuecomment-663738357) by @developit I jumped on it as a chance to get rid of moderate security vulnerability after `preact-cli` install. 

I don't think so?


packages/cli/lib/lib/webpack/webpack-client-config.js
```diff
@@ -138,17 +138,7 @@ function getBabelEsmPlugin(config) {
 					: '[name].esm.js',
 				chunkFilename: '[name].chunk.[chunkhash:5].esm.js',
 				excludedPlugins: ['BabelEsmPlugin', 'InjectManifest'],
-				beforeStartExecution: (plugins, newConfig) => {
-					const babelPlugins = newConfig.plugins;
-					newConfig.plugins = babelPlugins.filter((plugin) => {
```
prateekbh, line 143: this makes sure `'fast-async'` are not added, why remove this?

packages/cli/lib/lib/babel-config.js
```diff
@@ -11,11 +11,12 @@ module.exports = function (env, options = {}) {
 					targets: {
 						browsers: options.browsers,
 					},
-					exclude: ['transform-regenerator', 'transform-async-to-generator'],
+					exclude: ['transform-regenerator'],
 				},
 			],
 		],
 		plugins: [
+			require.resolve('babel-plugin-transform-async-to-promises'),
```
prateekbh, line 19: in modern mode we don't need this. All modern browsers support async/await.  Wouldn't adding this here add it for both modern and legacy mode?

prateekbh (dismissed):
some questions

packages/cli/lib/lib/webpack/webpack-client-config.js
teodragovic: I did a serach for `fast-async` in code and figured if I'm removing it as dependency any related code would be safe to remove also. Unles goal was to also prevent users from adding it via custom `babelrc`?

packages/cli/lib/lib/babel-config.js
teodragovic: True. Removed.

teodragovic: Hm, now the tests are failing again. I'll admit this task proved more complex than I thought so I would welcome some guidance around it.

developit:
One thing I'd like to consider before we jump to merging this: in the case where `esm` output is _not_ enabled, outputting regenerator is a bad default. Perhaps we could have that case use `babel-plugin-transform-async-to-promises`?

teodragovic:
> outputting regenerator is a bad default.

Due to amount of code transpiled or is more to it? My (very shallow) understanding is that babel would adjust output based on browserlist config (either default or provided by user) so if deeper browser support is needed (ie. no esm) it would transpile all the way down to promises. I guess then problem might be disabling esm but not updating browserlist?

I'm also interested to learn how https://github.com/preactjs/preact-cli/pull/1116 plays into this?

peterbe:
@teodragovic Any chance you can take a look at those merge conflicts with the main branch and give this PR some fresh air? 
Also, I thing you can add `Fixes #1591` into the original PR comment.

ForsakenHarmony (approved):

PR #1355, opened 2020-07-30
Update install instructions to use npx

teodragovic:

This change recommends using npx to avoid global installs of `preact-cli`

**Summary**

Related discussions: https://github.com/preactjs/preact-cli/pull/1343 and https://github.com/preactjs/preact-cli/issues/1342

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ $ npx preact-cli create <template-name> <project-name>
 Example:
 
 ```sh
-$ npx preact create default my-project
+$ npx preact-cli create default my-project
 ```
 
 The above command pulls the template from [preactjs-templates/default], prompts for some information, and generates the project at `./my-project/`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -43,24 +43,20 @@
...
-```sh
-$ npm install -g preact-cli
-```
+> **Important**: [Node.js](https://nodejs.org/en/) > V8.x and npm 5.2+ is a minimum requirement.
 
 ### Usage
 
 ```sh
-$ preact create <template-name> <project-name>
+$ npx preact create <template-name> <project-name>
```
prateekbh, line 53: should this be preact or preact-cli?

RyanChristian4427:
```suggestion
$ npx preact-cli create <template-name> <project-name>
```

prateekbh:
The command for create would now be different from build/etc. any thoughts on that @developit ?

developit:
I'm totally fine with recommending npx here, or if we wanted to show both options. People routinely file issues relating to `npm i -g`. TBH the create command could be its own module, it doesn't use anything from the rest of the CLI and would be tiny.

PR #1382, opened 2020-08-20
fix(cli): add missing resolve-from dependency to package.json

NicolaiSchmid:
Bugfix

**Summary**
The cli package makes a request to `resolve-from` which isn't listed in its dependencies and causes issues when using with yarn pnp.
https://github.com/preactjs/preact-cli/blob/master/packages/cli/lib/lib/webpack/utils.js

No breaking change


ForsakenHarmony (approved):

ForsakenHarmony (approved):

ForsakenHarmony:
Thanks for the PR @NicolaiSchmid

PR #1389, opened 2020-08-31
remove installation option

p-adams:
Since the usage option explains the installation steps, and the installation option doesn't link anything in the `README`, it seems like it can be removed.


ForsakenHarmony (approved):

ForsakenHarmony:
Thanks for the pr @p-adams

PR #1395, opened 2020-09-01
(fix) - HMR and Typescript

JoviDeCroock:
On windows the exclude of the Babel.overrides doesn't work making transformations not apply correctly. We had multiple copies of the devServer present and some of them hot others not which make WDS confused.

I need to accurately test this PR a bit more, tested on Windows and it works now need to do it on a UNIX machine.

### DIFF
--- a/packages/cli/lib/lib/babel-config.js
+++ b/packages/cli/lib/lib/babel-config.js
@@ -36,7 +36,7 @@ module.exports = function (env, options = {}) {
 		overrides: [
 			// Transforms to apply only to first-party code:
 			{
-				exclude: /(^|\/|\\)node_modules(\/|\/)/,
+				exclude: /node_modules/,
 				presets: [
 					[require.resolve('@babel/preset-typescript'), { jsxPragma: 'h' }],
 				],
--- a/packages/cli/lib/lib/babel-config.js
+++ b/packages/cli/lib/lib/babel-config.js
@@ -36,7 +36,7 @@ module.exports = function (env, options = {}) {
 		overrides: [
 			// Transforms to apply only to first-party code:
 			{
-				exclude: /node_modules/,
+				exclude: '**/node_modules/**',
 				presets: [
 					[require.resolve('@babel/preset-typescript'), { jsxPragma: 'h' }],
 				],
--- a/packages/cli/lib/lib/webpack/run-webpack.js
+++ b/packages/cli/lib/lib/webpack/run-webpack.js
@@ -73,7 +73,6 @@ async function devBuild(env) {
 
 		let c = Object.assign({}, config.devServer, {
 			stats: { colors: true },
-			hot: !env.refresh,
 		});
 
 		let server = new DevServer(compiler, c);
--- a/packages/cli/lib/lib/webpack/webpack-client-config.js
+++ b/packages/cli/lib/lib/webpack/webpack-client-config.js
@@ -25,7 +25,7 @@ const cleanFilename = (name) =>
 	);
 
 async function clientConfig(env) {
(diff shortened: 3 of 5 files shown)


packages/cli/lib/lib/webpack/webpack-client-config.js
```diff
@@ -36,14 +36,6 @@ async function clientConfig(env) {
 		polyfills: resolve(__dirname, './polyfills'),
 	};
 
-	if (!isProd) {
-		entry.bundle = [
```
JoviDeCroock, line 40: Removed the unneeded devServers

packages/cli/lib/lib/webpack/run-webpack.js
```diff
@@ -73,7 +73,6 @@ async function devBuild(env) {
 
 		let c = Object.assign({}, config.devServer, {
 			stats: { colors: true },
-			hot: !env.refresh,
```
JoviDeCroock, line 76: Enabled the hot-server which we use as a base

packages/cli/lib/lib/babel-config.js
```diff
@@ -36,7 +36,7 @@ module.exports = function (env, options = {}) {
 		overrides: [
 			// Transforms to apply only to first-party code:
 			{
-				exclude: /(^|\/|\\)node_modules(\/|\/)/,
+				exclude: /node_modules/,
```
JoviDeCroock, line 39: Windows compatible exclude

ForsakenHarmony: I think the second `\/` at the end of the old regex was maybe supposed to be a `\\`, but this should be fine as well

packages/cli/lib/lib/webpack/webpack-client-config.js
ForsakenHarmony: Does this automatically get added by the plugin / config option?

packages/cli/lib/lib/babel-config.js
developit: lol yeah I thought that was odd. This should work:

```suggestion
				exclude: /(^|\/|\\)node_modules(\/|\\)/,
```

developit: Another option would be to use the match pattern:

```suggestion
				exclude: '**/node_modules/**',
```

packages/cli/lib/lib/webpack/webpack-client-config.js
JoviDeCroock: This is part of the `DevServer.hot` option

packages/cli/lib/lib/webpack/webpack-client-config.js
```diff
@@ -25,7 +25,7 @@ const cleanFilename = (name) =>
 	);
 
 async function clientConfig(env) {
-	const { isProd, source, src, refresh, cwd /*, port? */ } = env;
+	const { isProd, source, src, cwd /*, port? */ } = env;
```
marvinhagemeister, line 28: We should probably remove any mentions of the `--refresh` flag in the Readme too👍

marvinhagemeister (approved):
This is fantastic! Outstanding debugging skills🙌

packages/cli/lib/lib/webpack/webpack-client-config.js
JoviDeCroock: it still works, just removed from the `clientConfig` and now exclusively present in the `devConfig`

reznord (approved):

Inviz:
Thank you for giving this issue your attention!

Issue #1105, opened 2020-04-01
Error: Cannot find module 'mkdirp' when using preact-cli with Volta


t21u:
**Do you want to request a _feature_ or report a _bug_?**
Bug

**What is the current behaviour?**
I use [volta](https://volta.sh) to manage my stack. I tried to install preact-cli using volta but I got this error:

```
internal/modules/cjs/loader.js:985
  throw err;
  ^

Error: Cannot find module 'mkdirp'
Require stack:
- /opt/volta/tools/image/packages/preact-cli/3.0.0-rc.10/lib/commands/create.js
- /opt/volta/tools/image/packages/preact-cli/3.0.0-rc.10/lib/commands/index.js
- /opt/volta/tools/image/packages/preact-cli/3.0.0-rc.10/lib/index.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:982:15)
    at Function.Module._load (internal/modules/cjs/loader.js:864:27)
    at Module.require (internal/modules/cjs/loader.js:1044:19)
    at require (internal/modules/cjs/helpers.js:77:18)
    at Object.<anonymous> (/opt/volta/tools/image/packages/preact-cli/3.0.0-rc.10/lib/commands/create.js:6:16)
    at Module._compile (internal/modules/cjs/loader.js:1158:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1178:10)
    at Module.load (internal/modules/cjs/loader.js:1002:32)
    at Function.Module._load (internal/modules/cjs/loader.js:901:14)
    at Module.require (internal/modules/cjs/loader.js:1044:19)
    at require (internal/modules/cjs/helpers.js:77:18)
    at Object.<anonymous> (/opt/volta/tools/image/packages/preact-cli/3.0.0-rc.10/lib/commands/index.js:2:18)
    at Module._compile (internal/modules/cjs/loader.js:1158:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1178:10)
    at Module.load (internal/modules/cjs/loader.js:1002:32)
    at Function.Module._load (internal/modules/cjs/loader.js:901:14) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/opt/volta/tools/image/packages/preact-cli/3.0.0-rc.10/lib/commands/create.js',
    '/opt/volta/tools/image/packages/preact-cli/3.0.0-rc.10/lib/commands/index.js',
[...]

Issue #1046, opened 2020-03-11
Incompatible with Yarn 2 / berry


DerDrodt:
**Do you want to request a _feature_ or report a _bug_?**

Feature

**What is the current behaviour?**

When running the CLI with Yarn 2, because of Plug'n'Play, we get an error:

> ERROR No `node_modules` found! Please run `npm install` before continuing.

**If the current behaviour is a bug, please provide the steps to reproduce.**

[Install Yarn 2](https://yarnpkg.com/getting-started/install) and run `preact build`

**What is the expected behaviour?**

It should be able to build with Yarn 2

**If this is a feature request, what is motivation or use case for changing the behaviour?**

Developers will want to use the new version of Yarn.

**Please mention other relevant information.**

As far as I can tell, the problem is that the loaders in the Webpack configs are not declared using `require.resolve`. Doing so would allow us to use the [pnp-Webpack-plugin](https://github.com/arcanis/pnp-webpack-plugin)

Please paste the results of `preact info` here.
version 2.2.1

RyanChristian4427:
I'm also having a similar issue with yarn 2, though my error message looks slightly different.

```
/home/runner/work/portfolio/portfolio/.pnp.js:23658
    throw firstError;
    ^

Error: preact-cli tried to access resolve-from, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.

Required package: resolve-from (via "resolve-from")
Required by: preact-cli@virtual:b13403f544ae60ed18db7322dbe40ce5e5b7a27527d99666902b0cbfb04ef45f841e7aa0c277aa82316facbf58907381a40ff63acfc131de85da500a609d224e#npm:3.0.0-rc.10 (via /home/runner/work/portfolio/portfolio/.yarn/$$virtual/preact-cli-virtual-3d06aac47b/0/cache/preact-cli-npm-3.0.0-rc.10-d33f292597-2.zip/node_modules/preact-cli/lib/lib/webpack/)
```

DerDrodt:
> I'm also having a similar issue with yarn 2, though my error message looks slightly different.

That can easily be fixed (although preact-cli should fix it themselves)

Just add a .yarnrc.yml with the following contents: 
```yml
packageExtensions:
    "preact-cli@*":
        dependencies:
            "mkdirp": "*"
```

Side-note: to make preact-cli work right now, also add:
```yml
nodeLinker: node-modules
```

But that's not ideal

RyanChristian4427:
[...]

Issue #1043, opened 2020-03-10
Lerna, Yarn & Workspaces path resolution wrong for `preact` / prerender.js


tomsiwik:
**Do you want to request a _feature_ or report a _bug_?**
_bug_

**What is the current behaviour?**
Because of #803, which was a fix for #795 mono-repo configurations no longer resolves `preact` correctly and returns:

`✖ ERROR Template execution failed: Error: Cannot find module './node_modules/preact'`

This is happening when using `lerna`, `yarn` and a `workspaces` configuration.

**If the current behaviour is a bug, please provide the steps to reproduce.**

Please use the following global packages and pkgmgr to reproduce:
- `yarn`
- `preact-cli`
- `lerna`

Easy shell script:
https://gist.github.com/tomsiwik/901c1acb122c9bb18d513bb348b201ea

**What is the expected behaviour?**

`preact` and `preact-render-to-string` should use common `node_modules` path either with `require('preact')` or relative to the module beeing used `${__dirname}/../../../../preact` (fragile).
One other solution is regexing for `__dirname`: `const moduleRoot = __dirname.match(/.+(?<=.*node_modules)/gm)[0]`

**If this is a feature request, what is motivation or use case for changing the behaviour?**

Workspaced lerna breaks now what used to work pre-fix.
See change: https://github.com/preactjs/preact-cli/pull/803/files#diff-dc9a93cd49896d6e2f0463ca332bf943L28

**Please mention other relevant information.**

Please paste the results of `preact info` here.
`preact-cli 2.2.1`

alexjsdev:
I just moved a preact app into my lerna monorepo that's using yarn workspaces and I'm encountering the same exact issue. Is there any workaround or progress on this?

KeithMcKiernan:
Running into the same issue as this when using preact-cli inside a mono-repo. Is there any progress on this issue?

tomsiwik:
The only workaround for me was disabling yarn workspaces for lerna.

kidqueb:
Not ideal, but after install I just copy `preact` and `preact-render-to-string` to the packages `node_modules`. I run a top level `prepare` command (I actually use ultra-runner) which each package handles in their own specific package.json.

If you have everything in `./packages` it'd be something like: 

```
...
"scripts": {
    "prepare": "cp -R ../node_modules/preact ./node_modules && cp -R ../node_modules/preact-render-to-string ./node_modules",
}
...
```

JohnnyCrazy:
[...]

PR #1418, opened 2020-09-13
feat(cli): add monorepo and Yarn PnP support

merceyz:

bugfix, feature

Added an e2e test that will test `preact-cli` for undeclared dependencies and incorrect assumptions about the `node_modules` layout using Yarn PnP's strict dependency checks.

**Summary**

`preact-cli` doesn't work with Yarn PnP and monorepos using `node_modules` due to its incorrect assumptions about the `node_modules` layout, various undeclared dependencies, and not resolving dependencies it provides.

Fixes https://github.com/preactjs/preact-cli/issues/1046
Fixes https://github.com/preactjs/preact-cli/issues/1105
Fixes https://github.com/preactjs/preact-cli/issues/1043

No

**Other information**

N/A

### DIFF
--- a/packages/cli/lib/commands/build.js
+++ b/packages/cli/lib/commands/build.js
@@ -1,7 +1,6 @@
 const rimraf = require('rimraf');
 const { resolve } = require('path');
 const { promisify } = require('util');
-const { isDir, error } = require('../util');
 const runWebpack = require('../lib/webpack/run-webpack');
 const { validateArgs } = require('./validate-args');
 
@@ -94,15 +93,6 @@ async function command(src, argv) {
 	argv.production = toBool(argv.production);
 
 	let cwd = resolve(argv.cwd);
-	let modules = resolve(cwd, 'node_modules');
-
-	if (!isDir(modules)) {
-		return error(
-			'No `node_modules` found! Please run `npm install` before continuing.',
-			1
-		);
-	}
-
 	if (argv.clean === void 0) {
 		let dest = resolve(cwd, argv.dest);
 		await promisify(rimraf)(dest);
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -130,8 +130,6 @@
 		"prompts": "^2.2.1",
 		"raw-loader": "^4.0.0",
 		"react-refresh": "0.8.3",
-		"require-relative": "^0.8.7",
-		"resolve-from": "^5.0.0",
 		"rimraf": "^3.0.2",
 		"sade": "^1.4.1",
 		"size-plugin": "^2.0.1",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -123,6 +123,7 @@
 		"native-url": "0.3.4",
 		"optimize-css-assets-webpack-plugin": "^5.0.1",
 		"ora": "^4.0.3",
+		"pnp-webpack-plugin": "^1.6.4",
 		"postcss-load-config": "^2.1.0",
 		"postcss-loader": "^3.0.0",
 		"progress-bar-webpack-plugin": "^2.1.0",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -47,6 +47,7 @@
 	"devDependencies": {
 		"html-looks-like": "^1.0.2",
 		"jest": "^26.0.1",
+		"less-loader": "^7.0.1",
 		"ncp": "^2.0.0",
 		"node-sass": "^4.12.0",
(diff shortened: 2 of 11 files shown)


prateekbh:
This is awesome. Thanks @merceyz .
Can you add a test case where one of the pages uses any `hook`? Thats where i guess this broke earlierr

merceyz:
> Can you add a test case where one of the pages uses any hook? Thats where i guess this broke earlierr

In normal repos the behaviour should be the same, and hopefully already covered by tests

packages/cli/package.json
```diff
@@ -88,6 +103,7 @@
 		"compression-webpack-plugin": "^4.0.0",
 		"console-clear": "^1.0.0",
 		"copy-webpack-plugin": "^5.0.4",
+		"create-require": "^1.0.2",
```
rschristian, line 106: I'm not sure we need this. v10 is about to be EOL and `createRequire` is >= 12.2. I'm quite tempted to say that's fine?

merceyz: The `package.json` claims to support `node >= 8` https://github.com/preactjs/preact-cli/blob/fea6647982b639044130bbcc15a85ff8878a7388/packages/cli/package.json#L28-L30
As soon as it's over v8.9 I can just use `require.resolve`

rschristian: Ah, wasn't entirely sure what you were doing with it, just looked up the dep.

Anyways, you can ignore that, that's outdated and really needs to be updated. https://github.com/preactjs/preact-cli/pull/1493#issuecomment-744227337

Though that comment does say v12.0, so I take back what I said about the 12.2. I thought it was just 12.x

packages/cli/lib/commands/build.js
```diff
@@ -94,15 +93,6 @@ async function command(src, argv) {
...
 	let cwd = resolve(argv.cwd);
-	let modules = resolve(cwd, 'node_modules');
-
-	if (!isDir(modules)) {
-		return error(
-			'No `node_modules` found! Please run `npm install` before continuing.',
-			1
-		);
-	}
-
```
merceyz, line 105: If you're in a monorepo and the hoisting is good this folder wont exist so checking if the project is installed like this is not correct

packages/cli/package.json
merceyz: Alright, `create-require` has been removed

ljharb:
Note that adding peer deps is a breaking change, even when optional, for clients that don’t support optional peer deps - which includes most of npm 6, and some versions of yarn.

merceyz:
> Note that adding peer deps is a breaking change, even when optional, for clients that don’t support optional peer deps - which includes most of npm 6, and some versions of yarn.

@ljharb To prevent old package managers from freaking out while letting newer package mangers know about them Yarn and pnpm (unsure about npm) supports just setting `peerDependenciesMeta.optional`, so the entries in `peerDependencies` can be removed if that's what the Preact team wants

ljharb:
@merceyz yes, i'm saying that support for `peerDependenciesMeta` itself isn't in older versions of yarn, pnpm, or npm, which makes it a breaking change, because prior to `peerDependenciesMeta.optional`, all peer deps are and have always been required.

You can mitigate this by adding engines entries for yarn and npm (and pnpm, i suppose) - but doing so is also a breaking change.

merceyz:
> @merceyz yes, i'm saying that support for peerDependenciesMeta itself isn't in older versions of yarn, pnpm, or npm, which makes it a breaking change, because prior to peerDependenciesMeta.optional, all peer deps are and have always been required.

I know, that's why I said only specifying that field is the solution to that problem

> [...] just setting peerDependenciesMeta.optional, so the entries in peerDependencies can be removed if that's what the Preact team wants

ljharb:
I’m under the impression that peer deps are invalid if they are only listed in peerDependenciesMeta - Meta is a modifier, not a declaration.

merceyz:
Yarn and pnpm (again, unsure about npm) treats them as an implicit peer dependency to allow projects to declare their dependencies without messing with older package managers

ForsakenHarmony (approved):

osdiab:
I can really use this at the moment - any idea when the next `preact-cli` release to `npm` might be released? thanks!

rschristian:
This breaks the [`getLoadersByName` helper](https://github.com/preactjs/preact-cli/blob/master/docs/webpack-helpers.md#getloadersbyname) as loaders will now have their full path in the file system as the name.

PR #1423, opened 2020-09-14
fix: Enabling sideEffects on style loaders

### LINKED ISSUE #1411: CLI takes "sideEffects" field of package into account while building application

This is a bug

Currently the CSS loader does not override the `"sideEffects"` field of a package. This is an issue as CSS imports in the form of `import './style.css';` have side effects by their very nature. They should not be affected as that style of import should either causes a side effect or just doesn't work.

Set the `"sideEffects"` field to `false` in a package file, import non-moduled CSS (`import './style.css';`), then run a production build. The CLI will ignore that import entirely.

The CSS should be included.  

```
Environment Info:
  System:
    OS: Linux 5.7 Manjaro Linux
    CPU: (4) x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
  Binaries:
    Node: 14.9.0 - /tmp/yarn--1599859321698-0.6191202931910675/node
    Yarn: 1.22.5 - /tmp/yarn--1599859321698-0.6191202931910675/yarn
    npm: 6.14.8 - ~/.npm-global/bin/npm
  Browsers:
    Firefox: 80.0.1
  npmPackages:
    preact: ^10.4.7 => 10.4.7 
    preact-cli: ^3.0.1 => 3.0.1 
    preact-render-to-string: ^5.1.9 => 5.1.
```

rschristian:

Bugfix / feature

Yes. Added a test to ensure that non-module CSS imported within a package that has `"sideEffects": false` would result in CSS being inlined in the head of the document. If the change to the style loader is reverted, the CSS import is ignored (skipped? not sure what it technically does) and the CSS will not end up being inlined.

**Summary**

Fixes #1411

I noticed this issue while building a component library. This library uses the Preact CLI to (mainly) run a development environment and Microbundle is used to the build/package the library itself. This is all well and fine. 

The issue that this PR fixes arose when I had decided that my dev environment looked good enough to toss up as demo of the library. The `sideEffects` field stops the Preact CLI from importing non-module CSS when set to `false` in a production build. This should not happen as `import 'x.css';` will always have a side effect. It's the very nature of that type of import.

Reading through various threads led me to CRA's solution, which was to [enable sideEffects for their style loaders](https://github.com/facebook/create-react-app/blob/65d8eb23f28475716e7e9497215cd574c1bf0b6c/packages/react-scripts/config/webpack.config.js#L502). Doing something similar with our setup seems to work just as well.

@developit had suggested something along the lines of the following which also would work with slight modification:

```
// in webpack.config.base.js
module: {
  rules: [
    {
      test: /\.m?[jt]sx?$/,
      exclude: nodeModules,
      sideEffects: true
    }
  ]
}
```

Changing this to target CSS works just as well as what I did in this PR. I figured it's simpler to just add a field to the existing loader rather than add a new one. There might be a distinction between the two methods, but I'm not sure what that would be.

Shouldn't be breaking.

### DIFF
--- a/packages/cli/lib/lib/webpack/webpack-base-config.js
+++ b/packages/cli/lib/lib/webpack/webpack-base-config.js
@@ -247,7 +247,6 @@ module.exports = function (env) {
 							},
 						},
 					],
-					sideEffects: true,
 				},
 				{
 					// External / `node_module` styles
--- a/packages/cli/lib/lib/webpack/webpack-base-config.js
+++ b/packages/cli/lib/lib/webpack/webpack-base-config.js
@@ -269,6 +269,10 @@ module.exports = function (env) {
 							},
 						},
 					],
+					// Don't consider CSS imports dead code even if the
+					// containing package claims to have no side effects.
+					// Remove this when webpack adds a warning or an error for this.
+					// See https://github.com/webpack/webpack/issues/6571
 					sideEffects: true,
 				},
 				{
--- a/packages/cli/lib/lib/webpack/webpack-base-config.js
+++ b/packages/cli/lib/lib/webpack/webpack-base-config.js
@@ -247,6 +247,7 @@ module.exports = function (env) {
 							},
 						},
 					],
+					sideEffects: true,
 				},
 				{
 					// External / `node_module` styles
@@ -269,6 +270,7 @@ module.exports = function (env) {
 							},
 						},
 					],
+					sideEffects: true,
 				},
 				{
 					test: /\.(xml|html|txt|md)$/,
--- a/.changeset/gold-pumas-invite.md
+++ b/.changeset/gold-pumas-invite.md
@@ -1,0 +1,5 @@
+---
+'preact-cli': patch
+---
+
+Fixes bug with style loader that would strip non-module CSS files if 'sideEffects' was set to false for the package.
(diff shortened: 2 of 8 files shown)


packages/cli/lib/lib/webpack/webpack-base-config.js
```diff
@@ -269,6 +269,11 @@ module.exports = function (env) {
 							},
 						},
 					],
+					// Don't consider CSS imports dead code even if the
+					// containing package claims to have no side effects.
+					// Remove this when webpack adds a warning or an error for this.
+					// See https://github.com/webpack/webpack/issues/6571
```
RyanChristian4427, line 275: [Comment taken right from CRA](https://github.com/facebook/create-react-app/blob/65d8eb23f28475716e7e9497215cd574c1bf0b6c/packages/react-scripts/config/webpack.config.js#L498), I felt that having an explanation in the source is useful.

RyanChristian4427:
Falling victim to the flaky `apple-touch-icon` test. Unrelated failing.

RyanChristian4427:
Something that might be awkward with this setup is the disabling of CSS modules in `routes/` and `components/`. If someone were to disable modules then they'd need to know to add `sideEffects: true` else they run into issues too. Not sure if that's something we need to handle or just keep in mind in case anyone runs into issues.

ForsakenHarmony (approved):

ForsakenHarmony:
bors retry

PR #1427, opened 2020-09-16
Feature: Watch mode without dev server 

### LINKED ISSUE #1248: Watch mode without webpack's dev server

rschristian:

Feature

Yes. I added a test under `watch.test.js` that will run the file watcher and check to see if the built output is updated after a change occurs to the source. I figure this approach is simpler than the alternatives and still ensures the functionality works. So long as the built output changes the feature should be fine. I don't think we need a browser test. 

**Summary**

This builds upon #1264 which leads back to & would close #1248.

Currently, the experience for developing SSR apps using the Preact CLI isn't great. To produce a bundle that can be used a developer needs to run the full production build and then rerun it on every change. This is slow and un-automated.

A solution, as pointed out by @oliverstr, is to provide an option to disable the dev server in `watch` mode. Any changes to the source would have the CLI rebuild and developers can consume the build in any way that they please.

@oliverstr's PR for implementing this feature (#1264) has been untouched since June 29th so I figure at this point it is fair to step in and try to complete. His PR had a few issues, besides being behind the current head, namely, it would not clear the `build/` directory, leading to a large build up of temporary files, it ignored changes to `src/*` by default, and it could not handle CSS imports without wrapping them in a window check (`if (typeof window !== "undefined") { ... }`). I have fixed all three of these issues and a few more that popped up along the way to finish off his feature. 

You can use the following repo to try this out, though you'll need to bring your own CLI copy with linking or relative paths. Simple server set up with Polka and Nodemon. Simply run `yarn serve:dev` in one terminal window and `yarn serve` in another. `serve:dev` starts the CLI watch mode, `serve` starts Polka with Nodemon. Nothing fancy.

[ryanchristian4427/preact-cli-watch-without-devserver](https://github.com/RyanChristian4427/preact-cli-watch-without-devserver)

No breaking changes. This feature is behind a flag.

### DIFF
--- a/packages/cli/tests/watch.test.js
+++ b/packages/cli/tests/watch.test.js
@@ -2,6 +2,7 @@ const fs = require('../lib/fs');
 const { resolve } = require('path');
 const startChrome = require('./lib/chrome');
 const { create, watch } = require('./lib/cli');
+const { sleep } = require('./lib/utils');
 
 const { loadPage, waitUntilExpression } = startChrome;
 let chrome, server;
@@ -32,5 +33,29 @@ describe('preact', () => {
 		);
 
 		server.close();
+	});
+
+	it('should hot reload and prerender without a development server', async () => {
+		let dir = await create('default');
+		await watch(dir, undefined, undefined, false);
+
+		const initialContent = await fs.readFile(
+			resolve(dir, './build/bundle.js'),
+			'utf8'
+		);
+
+		let header = resolve(dir, './src/components/header/index.js');
+		let original = await fs.readFile(header, 'utf8');
+		let update = original.replace('<h1>Preact App</h1>', '<h1>Test App</h1>');
+		await fs.writeFile(header, update);
+
+		await sleep(1500); // wait for a rebuild
+		const updatedContent = await fs.readFile(
+			resolve(dir, './build/bundle.js'),
+			'utf8'
+		);
+
+		expect(updatedContent).not.toEqual(initialContent);
+		expect(updatedContent).not.toContain('Preact App');
 	});
 });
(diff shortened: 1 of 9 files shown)


RyanChristian4427:
A few changes are purely code-style based that my IDE caught. I figure I won't revert them for now as they do look to be correct changes, but I can undo this if it makes the PR too messy.

RyanChristian4427:
I've tried a few times now to get the `apple-touch-icon` to pass but it just doesn't seem to want to today.

packages/cli/tests/watch.test.js
```diff
@@ -33,4 +34,28 @@ describe('preact', () => {
...
+
+		await sleep(1500); // wait for a rebuild
+		const updatedContent = await fs.readFile(
+			resolve(dir, './build/bundle.js'),
+			'utf8'
+		);
+
+		expect(updatedContent).not.toEqual(initialContent);
+		expect(updatedContent).not.toContain('Preact App');
+	});
```
RyanChristian4427, line 60: I'm not 100% on this test, if there's a better way to go about this certainly let me know. Without starting and stopping the Polka server that's there I believe this is the best option? Just does a simple comparison/look through the bundle output. 

Very possible there's a better way to do this though.

rschristian:
I'd need to come back to look at this again.

PR #1435, opened 2020-09-21
Fix: Include ReadMe when publishing package

rschristian:

Bugfix when publishing

No

**Summary**

Currently, [the ReadMe is not included with the package when publishing it](https://www.npmjs.com/package/preact-cli/v/3.0.2). This can be quite inconvenient as there's no documentation to go with the package.

It seems [the last time the readme was included was back with `3.0.0-next.1`](https://www.npmjs.com/package/preact-cli/v/3.0.0-next.1). While there was an attempt to fix this with #806, it doesn't seem to have worked. I'm going to guess that this is because publishing happens from with `packages/cli` rather than the root.

This moves the copy & clean up commands into the cli package itself.

No

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -9,9 +9,7 @@
 	},
 	"scripts": {
 		"pretest": "rimraf ./tests/output",
-		"test": "jest --forceExit",
-		"prepack": "ncp ../../README.md README.md",
-		"postpack": "rimraf README.md"
+		"test": "jest --forceExit"
 	},
 	"jest": {
 		"setupFilesAfterEnv": [
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -9,7 +9,9 @@
 	},
 	"scripts": {
 		"pretest": "rimraf ./tests/output",
-		"test": "jest --forceExit"
+		"test": "jest --forceExit",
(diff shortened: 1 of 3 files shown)


developit (approved):

developit:
🦙!

packages/cli/package.json
```diff
@@ -9,7 +9,9 @@
 	},
 	"scripts": {
 		"pretest": "rimraf ./tests/output",
-		"test": "jest --forceExit"
+		"test": "jest --forceExit",
+		"prepack": "ncp ../../README.md README.md",
```
JoviDeCroock, line 13: I don’t think this would work, we use yarn release to publish which triggers the changeset publish. I’d make it part of that command since changesets seems to circumvent package lifecycles. An example can be found [here](https://github.com/JoviDeCroock/prefresh/blob/main/package.json#L8)

JoviDeCroock:
This Pr will also need a changeset, you can do so by running yarn changeset at the root

packages/cli/package.json
rschristian: Ah, gotcha, thanks. Should be fixed now.

ForsakenHarmony (approved):

PR #1440, opened 2020-09-29
Refactor: Move CLI `create` into its own package

rschristian:

Refactor

No new tests, but old ones were updated/fixed

**Summary**

Currently the CLI performs two, mostly separate functions: creating new projects and running existing ones. With this combined functionality, the package is heavier than need be in both applications of the package. Especially now that `npx` is the suggested way to use the CLI to bootstrap new projects, having a lighter tool is very beneficial.

This PR would separate out the `create` functionality of the CLI into what I've temporarily called `create-preact-app`. I do believe that name is taken, so will likely need a different name.

This was mentioned in a comment on the change to use `npx` as a potential avenue for change in the future: https://github.com/preactjs/preact-cli/pull/1355#issuecomment-668680852

I don't think this counts as breaking?


packages/cli/tests/lib/cli.js
```diff
@@ -1,6 +1,7 @@
 const { join } = require('path');
 const { existsSync, unlinkSync, symlinkSync } = require('fs');
 const cmd = require('../../lib/commands');
+const create = require('../../../create-preact-app/lib/commands').create;
```
rschristian, line 4: This is a bit messy, not sure if there's a better way

packages/create-preact-app/lib/index.js
```diff
@@ -0,0 +1,47 @@
...
+notifier({ pkg }).notify();
+
+process.on('unhandledRejection', (err) => {
+	error(err.stack || err.message);
+});
+
+let prog = sade('create-preact-app').version(pkg.version);
+
+prog
+	.command('create [template] [dest]', '', { default: true })
```
rschristian, line 33: I made this the default so it can be ran as `npx create-preact-app default my-application`. I figure that's better than the alternative `npx create-preact-app create default my-application`.

packages/create-preact-app/lib/symbols.js
```diff
@@ -0,0 +1,15 @@
...
+	error: red('✖'),
+};
+
+const win = {
+	info: blue('i'),
+	warning: yellow('‼'),
+	error: red('×'),
+};
+
+module.exports = process.platform === 'win32' ? win : main;
```
rschristian, line 15: Items like this (that are duplicated) might need suggestions from someone who's more experienced with mono-repos. Not sure if there's a better solution without introducing another package that holds some of this common stuff?

packages/create-preact-app/tests/create.test.js
```diff
@@ -37,10 +37,4 @@ describe('preact create', () => {
 
 		expect(template.includes('apple-touch-icon')).toEqual(true);
 	});
-
-	// it('should fail given an invalid name', async () => {
-	// 	const exit = jest.spyOn(process, 'exit');
-	// 	await create('default', '*()@!#!$-invalid-name');
-	// 	expect(exit).toHaveBeenCalledWith(1);
-	// });
```
rschristian, line 45: Removed as this has been commented out and unused for 3 years: https://github.com/preactjs/preact-cli/commit/fef996e7ee55c58969ba52557d33e8e5bde81f99

rschristian (commented):
There's also a fair few stylistic changes done automatically by my IDE. I figure I'll keep them in, they look to be right. I can remove them if they make this too busy.

ForsakenHarmony:
Personally preferred having it in the same package, it was separate enough to me, but I'm fine with it if other people agree

rschristian:
It certainly adds some complexity to the repo and probably some slight duplication of code too.

Just an idea for a lighter (and therefore faster with `npx`) `create`.

rschristian:
Closing this as I'd like to redo it.

PR #1441, opened 2020-09-30
Adding blank page for fallback

prateekbh:

Creates a non SSRed html file for caching in service worker. All failed requests would go to this page and the app would render from here

Yes

**Summary**
Currently we pre-cache all the index.html files of the application. Given the size of the app, these can be many. In order to avoid this we will now create a non ssred version of homepage and upon error will use that to bootstrap the app.

No


developit:
@prateekbh this is great.  One thing I was thinking: it would be amazing if we could check for the fallback URL (or something like `200.html`?) in prerender-urls. If the developer has listed it there, we could allow prerendering of that URL rather than using the fully blank page. That way URL misses would still get an App She'll to boot up from.

prateekbh:
Yes that sounds great! I'll make those changes

prateekbh:
@developit done 👍

packages/cli/lib/lib/webpack/render-html-plugin.js
```diff
@@ -154,7 +171,11 @@ class PrerenderDataExtractPlugin {
 		this.data_ = JSON.stringify(cliData.preRenderData || {});
 	}
 	apply(compiler) {
-		compiler.hooks.emit.tap('PrerenderDataExtractPlugin', compilation => {
+		compiler.hooks.emit.tap('PrerenderDataExtractPlugin', (compilation) => {
+			if (this.location_ === `${PREACT_FALLBACK_URL}/`) {
```
developit, line 175: Wondering if we still need this with the custom 200 option.

prateekbh: Yes, this makes sure that the fallback page has no pre-render data. 
Thus in this case the pre-render data will always rely on the current URL .

packages/cli/lib/lib/webpack/render-html-plugin.js
```diff
@@ -137,12 +145,21 @@ module.exports = async function(config) {
 			);
 		}
 	}
+	/**
+	 * We cache a non SSRed page in service worker so that there is
+	 * no flash of content when user lands on routes other than `/`.
+	 * And we dont have to cache every single html file.
+	 * Go easy on network usage of clients.
+	 */
+	!pages.map((page) => page.url).includes(PREACT_FALLBACK_URL) &&
```
ForsakenHarmony, line 154:
```suggestion
	!pages.find((page) => page.url === PREACT_FALLBACK_URL) &&
```

packages/cli/tests/service-worker.test.js
```diff
@@ -64,4 +84,25 @@ describe('preact service worker tests', () => {
 		expect(initialContent).not.toEqual(refreshedContent);
 		expect(refreshedContent.includes(NEW_TITLE)).toEqual(true);
 	});
+	it('should respond with 200.html when offline', async () => {
```
ForsakenHarmony, line 87:
```suggestion

	it('should respond with 200.html when offline', async () => {
```

ForsakenHarmony (approved):

developit (approved):

PR #1467, opened 2020-10-31
Allow only valid options for commands

prateekbh:

This now throws error if the user passes invalid argument around `build`, `create` or `watch`.

Yes

**Summary**

Sample
```
$ mypreact watch --refresher
✖ ERROR Invalid argument refresher passed to watch. Please refer to 'preact watch --help' for full list of options.

$ mypreact build --no-service-worker
✖ ERROR Invalid argument service-worker passed to build. Please refer to 'preact build --help' for full list of options.
```

No.

### DIFF
--- a/packages/cli/lib/commands/build.js
+++ b/packages/cli/lib/commands/build.js
@@ -3,10 +3,12 @@ const { resolve } = require('path');
 const { promisify } = require('util');
 const { isDir, error } = require('../util');
 const runWebpack = require('../lib/webpack/run-webpack');
+const { validateArgs } = require('./validate-args');
 
 const toBool = val => val === void 0 || (val === 'false' ? false : val);
 
-module.exports = async function (src, argv) {
+async function command(src, argv) {
+	validateArgs(argv, options, 'build');
 	argv.src = src || argv.src;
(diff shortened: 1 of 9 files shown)


packages/cli/lib/commands/build.js
```diff
@@ -32,4 +34,83 @@ module.exports = async function (src, argv) {
...
+		name: '--preload',
+		description: 'Adds preload tags to the document its assets',
+		default: false,
+	},
+	{
+		name: '--analyze',
+		description: 'Launch interactive Analyzer to inspect production bundle(s)',
+	},
+	{
+		name: '--prerenderUrls',
```
developit, line 88: We allow `--prerender-urls` too right?

ForsakenHarmony: I don't know if camel case is really something a cli should use, but seems like that's what we had before and only that

prateekbh: seems like we don't allow `--prerender-urls` as of now. So technically changing this would be a breaking change.

ForsakenHarmony (approved):

PR #1493, opened 2020-12-09
Fix Node 10 support

gaetanmaisse:

Bugfix: replace use of Array.prototype.flat() to ensure Node 10 compatibility

No

**Summary**

As stated in the Readme and in the `engine` property of the `package.json`, `preact-cli` 3.0.x should be compatible with Node 10. However, https://github.com/preactjs/preact-cli/pull/1467/files#diff-490a9490c1fcd253d926ffeaea6b185b15522512e53f6a9f36493122fcae9d2eR4 introduced the use of `Array.prototype.flat()` (in `validate-args.js` file) which is available on Node 11+.

To work around that and ensure Node 10 compatibility I replaced it with a good old `reduce` + `concat`.

---

Here is the current issue when running `npx preact-cli create preactjs-templates/default preact-latest` with Node 10:
```sh
✖ ERROR TypeError: options.map(...).flat is not a function
    at validateArgs (/Users/gmaisse/.npm/_npx/13643/lib/node_modules/preact-cli/lib/commands/validate-args.js:4:72)
    at command (/Users/gmaisse/.npm/_npx/13643/lib/node_modules/preact-cli/lib/commands/create.js:210:2)
    at Sade.parse (/Users/gmaisse/.npm/_npx/13643/lib/node_modules/preact-cli/node_modules/sade/lib/index.js:189:56)
    at Object.<anonymous> (/Users/gmaisse/.npm/_npx/13643/lib/node_modules/preact-cli/lib/index.js:83:6)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:831:12)
```

No, it fixes one.

### DIFF
--- a/packages/cli/lib/commands/validate-args.js
+++ b/packages/cli/lib/commands/validate-args.js
@@ -1,7 +1,7 @@
 const { error } = require('../util');
 
 function validateArgs(argv, options, command) {
-	let normalizedOptions = options.map(option => option.name.split(',')).flat(1);
+	let normalizedOptions = options.map(option => option.name.split(',')).reduce((acc, val) => acc.concat(val), []);
 	normalizedOptions = normalizedOptions.map(option => {
 		option = option.trim();
 		if (option.startsWith('--')) {


yannbf:
Hey @prateekbh do you think you could maybe check this out? Thanks!

rschristian:
@gaetanmaisse Nice catch.

Do you mind adding 10.x to the CI version matrix? It would help prevent regressions.

gaetanmaisse:
Node 10's end of life is 2021-04-30 so it feels right to not spend a lot of time to ensure compatibility.
My opinion would be to merge this fix to ensure 3.x is still compatible with Node 10 and in the next few months stop the support of Node 10, update readmes, `engine` in `package.json` and publish a 4.x. But I really don't know what is the process in the `preact` ecosystem. 

What's your opinion on that?

Note: I discovered the problem with Storybook E2E tests as we are bootstrapping a fresh preact project and all our E2Es are currently running on Node 10 because the current version of Storybook is targeting Node 10+.

prateekbh:
I mean this PR itself is not a huge deal, but I do remember that it was a conscious call to keep the support to 12.0.

I would not want to add `10.x` to the matrix as that would mean that everyone needs to comply with that.

gaetanmaisse:
Not sure to understand what I need to do then, revert the matrix change but keep my fix? or closed this PR, update the [readme](https://github.com/preactjs/preact-cli/blob/master/README.md#requirements) + [package.json](https://github.com/preactjs/preact-cli/blob/master/packages/cli/package.json#L29) and ask for a major release?

rschristian:
Revert the matrix, yeah. I wasn't privy to the conversation about support to 12.0 and I only caught how close we were to it being EOL after I sent the message originally. 

Engine stuff will certainly need to be updated eventually.

gaetanmaisse:
Ok, I just forced push to clean the branch and only get the fix commit ;)

prateekbh (approved):

PR #1501, opened 2020-12-14
[WIP] Experimental prerenderer

### LINKED ISSUE #1482: RFC: Experimental Fast renderer

RFC

Currently pre-rendering is handled by `HTMLWebpackPlugin` and `webpack`. This is quite memory expensive and works on main thread. On a few random tests the `preact build` command simply crashes when asked to create a build with 500-1000 pages and some sizable amount of pre-render data.

Even in the scenario where it does not crash it takes quite a long time to finish the build, sometimes even 30 minutes or more. See #1463 .

Pre-rendering should not be this memory expansive and should finish much much early on. See gatsby/next and others for inspiration.

We can probably move pre-rendering out of the webpack toolchain and use nodejs worker threads to pre-render the pages in parallel. The new rendering will have a set API for the `ejs` templates which can be guarded with SEMVER. All pages can pre-render in their own worker threads and the files can be written at the end.

We can start with an experimental option in V3 and land it as stable in V4.

e.g. 
`preact build --experimental-fast-render`.

The API for the `ejs` templates would be as follows.

```js
{
  options: {
    webpack: {
      public_path: '...',
      assets: [...], //list of all assets generated by webpack
    },
    data: { 
      url: '...',
      title: '...'.
      CLI_DATA: {}, // all the pre-render data for this route
    }
  }
}
```

We would also morph this data to slightly match the `HTMLWebpackPlugin`'s existing schema so that if anybody is trying to use this option without much change in `template.html` then they would be able to.

prateekbh:
Fixes #1482 

Feature request.

Yes

**Summary**

Implements an experimental pre-renderer with a set of API for the data.

No

**Other information**

There is some effort to keep the current templates non-breaking but the API will eventually need to be updated and can onlhy land as stable in v4.0

**API**
The `template.html` will have access to the following API
```js
{
  options: {
    url,
    manifest,
    ssr: () => {},
    CLI_DATA: { url, ...routeData },
    webpack: {publicPath: '', asset: ['all generated assets']},
    
  }
}
```

### DIFF
--- a/packages/cli/lib/lib/fast-render/pool.js
+++ b/packages/cli/lib/lib/fast-render/pool.js
@@ -1,0 +1,25 @@
+const { cpus } = require('os');
+const resolved = Promise.resolve();
+module.exports = async function pool(
+	items,
+	iteratorFn,
+	concurrency = cpus().length
+) {
+	const itemsLength = items.length;
+	const returnable = [];
+	const executing = [];
+	for (const item of items) {
+		const promise = resolved.then(() => iteratorFn(item, items));
+		returnable.push(promise);
+		if (concurrency <= itemsLength) {
+			const execute = promise.then(() =>
+				executing.splice(executing.indexOf(execute), 1)
+			);
+			executing.push(execute);
+			if (executing.length >= concurrency) {
+				await Promise.race(executing);
+			}
+		}
+	}
+	return Promise.all(returnable);
+};
--- a/packages/cli/lib/lib/fast-render/renderer.js
+++ b/packages/cli/lib/lib/fast-render/renderer.js
@@ -21,7 +21,13 @@ async function render(src, dest, cwd, webpack, data) {
 		url,
 		manifest,
 		ssr: () => {
-			return prerender({ cwd, dest, src }, data);
+			const params = {
+				...data,
+				CLI_DATA: {
+					preRenderData: data,
+				},
+			};
+			return prerender({ cwd, dest, src }, params);
 		},
 		CLI_DATA: { url, ...routeData },
 		webpack,
--- a/packages/cli/lib/lib/fast-render/renderer.js
+++ b/packages/cli/lib/lib/fast-render/renderer.js
@@ -1,0 +1,60 @@
+const { isMainThread, parentPort, workerData } = require('worker_threads');
+const { readFile } = require('../../fs');
+const { join } = require('path');
+const prerender = require('../webpack/prerender');
+
+const ejs = require('ejs');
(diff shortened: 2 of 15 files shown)


packages/cli/lib/lib/fast-render/pool.js
```diff
@@ -0,0 +1,25 @@
+const { cpus } = require('os');
```
prateekbh, line 1: put copyright here

packages/cli/lib/lib/fast-render/renderer.js
```diff
@@ -0,0 +1,72 @@
...
+				CLI_DATA: {
+					preRenderData: data,
+				},
+			};
+			return prerender({ cwd, dest, src }, params);
+		},
+		CLI_DATA: { url, ...routeData },
+		webpack,
+	};
+	const htmlWebpackPlugin = {
```
prateekbh, line 35: explain this

psabharwal123:
@prateekbh can we also upgrade to html-webpack-plugin 4? https://dev.to/jantimon/html-webpack-plugin-4-has-been-released-125d

who can help review this change? would really like to try it out, our build with 100 pages takes a long time currently

psabharwal123:
@prateekbh why was this closed? What is the solution for builds that take a very long time?

rschristian:
@psabharwal123 Your best bet would be to move over to [WMR](https://github.com/preactjs/wmr) likely. We don't have infinite resources to develop everything in parallel, and WMR has a lot more momentum as well as advantages.

psabharwal123:
Is there a guide to migrating from preact cli to WMR? Where can I see the advantages of it over preact-cli so I can convince our team to look into this?

rschristian:
There is no guide at the moment, nor is there a list of advantages. WMR wasn't necessarily meant to be a successor to Preact-CLI, though that's now where more effort is being spent these days. Because of that no one's really taken time to try to sell WMR as an upgraded CLI, though maybe that's something that should be done. I'll see how people feel about that.

The upgrade might be quite a bit of work, as WMR is not Webpack-based, but Rollup-based. Any plugins you've added in your `preact.config.js` will need to find Rollup equivalents, or you could of course write your own.

LionsAd:
@psabharwal123 

Unfortunately the experimental renderer never did make it in, but if someone still needs a fast prerender.js script:

https://github.com/preactjs/preact-cli/issues/1684#issuecomment-1094349753

psabharwal123:
Thanks! I ll look into it

PR #1502, opened 2020-12-18
fix: Corrects mkdirp's status as dependency

rschristian:

Minor potential bugfix

No

**Summary**

The package `mkdirp` was listed as a `devDependency` even though [it was required during general, production use](https://github.com/preactjs/preact-cli/blob/58fe77789a119e3013c52b7ef75c35de9384f007/packages/cli/lib/commands/create.js#L6).

I don't believe this has ever been a problem due to it being a dependency of a fair number of other dependencies, but if those were to ever change then this would break.

No


rschristian:
Closing in favor of #1504

PR #1527, opened 2021-02-04
fix: do not prompt on supplying template as an argument.

### LINKED ISSUE #1523: Ambiguities with create command usage
This proposal aims at changing the current behavior of the `create` command.

As specified in the [docs](https://github.com/preactjs/preact-cli#preact-create), the `create` command expects the `template-name` and `project-name` (required arguments). However, the current implementation treats them as optional arguments.
https://github.com/preactjs/preact-cli/blob/4acc8c7768936cd128998971e0e6d7425cd5c204/packages/cli/lib/index.js#L44

```sh
$ preact create default
```

https://github.com/preactjs/preact-cli/blob/4acc8c7768936cd128998971e0e6d7425cd5c204/packages/cli/lib/commands/create.js#L209-L214

The current implementation checks for a `--template` flag, which doesn't exist, The prompt to choose a `preactjs-template` show up even though it is specified.

https://github.com/preactjs/preact-cli/blob/4acc8c7768936cd128998971e0e6d7425cd5c204/packages/cli/lib/commands/create.js#L81-L87

```sh
preact create [project-name] [options]
```
- It would be better to add a `--template` flag.
- The `--name` flag is not required.

N/A

jamesgeorge007:
patch

Nope

**Summary**
Supplying the `preactjs-template` as an argument doesn't show the respective prompt.

#1523 

Nope

**Other information**
N/A

### DIFF
--- a/packages/cli/lib/commands/create.js
+++ b/packages/cli/lib/commands/create.js
@@ -78,13 +78,13 @@ const options = [
 ];
 
 // Formulate Questions if `create` args are missing
-function requestParams(argv, templates) {
+function requestParams(repo, dest, argv, templates) {
 	const cwd = resolve(argv.cwd);
 
 	return [
 		// Required data
 		{
-			type: argv.template ? null : 'select',
+			type: repo ? null : 'select',
 			name: 'template',
 			message: 'Pick a template',
 			choices: templates,
@@ -96,12 +96,12 @@ function requestParams(argv, templates) {
 			message: 'Remote template to clone (user/repo#tag)',
 		},
 		{
-			type: argv.dest ? null : 'text',
+			type: dest ? null : 'text',
 			name: 'dest',
 			message: 'Directory to create the app',
 		},
 		{
-			type: prev => (!dirExists(cwd, prev || argv.dest) ? null : 'confirm'),
+			type: prev => (!dirExists(cwd, prev || dest) ? null : 'confirm'),
 			name: 'force',
 			message: 'The destination directory exists. Overwrite?',
 			initial: false,
@@ -211,7 +211,7 @@ async function command(repo, dest, argv) {
 	// Prompt if incomplete data
 	if (!repo || !dest) {
 		const templates = await fetchTemplates();
-		const questions = requestParams(argv, templates);
+		const questions = requestParams(repo, dest, argv, templates);
 		const onCancel = () => {
 			info('Aborting execution');
 			process.exit();
--- a/.changeset/tame-ravens-own.md
+++ b/.changeset/tame-ravens-own.md
@@ -1,0 +1,5 @@
+---
+'preact-cli': patch
+---
+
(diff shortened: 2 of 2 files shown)


rschristian:
Definitely don't think we want to deprecate anything here. This would also be a major, not a minor, as all deprecations are.

We just need to clean up the ambiguity/consistency between the docs and actual commands while also fixing that issue with the flag not existing. I don't think there's any need to break the existing behavior like this.

packages/cli/lib/commands/create.js
```diff
@@ -78,11 +78,16 @@ const options = [
...
 
 	return [
 		// Required data
+		{
+			type: name ? null : 'text',
+			name: 'name',
+			message: 'The name of your application',
+		},
 		{
 			type: argv.template ? null : 'select',
```
rschristian, line 87: You could just swap this flag for `repo` from `create()` and the issue would be resolved.

jamesgeorge007: The intended way to supply additional arguments to a subcommand would be along with the flags that denote their purpose.
```diff
- preact create [template-name] [project-name]
+ preact create [name] --template [template-name]
```

```sh
preact create [name] [options]
```

It would be better to have a `--template` flag in that regard. Nevertheless, I would be happy to address the current issue if the proposed change is not intentional.

rschristian: Sure, but I don't think the template is an "additional" argument. It's required; we just also have a way to catch users if they forget it. 

Regardless, a major, as this currently is, will likely be a long ways away. Better to just fix up the behavior now.

packages/cli/lib/commands/create.js
```diff
@@ -78,13 +78,13 @@ const options = [
 ];
 
 // Formulate Questions if `create` args are missing
-function requestParams(argv, templates) {
+function requestParams(repo, dest, argv, templates) {
```
rschristian, line 81: Is there a reason to drag in `dest` too?

rschristian:
I do agree with your original work and I think it's something we should look into, getting all of the commands a bit more consistent. It was just that a breaking change is so far away from being merge-able in all likelihood that it wouldn't solve the immediate issue of the flag not existing.

I might end up copying your original commit over to a more long term PR where the API of the CLI can be more properly considered.

packages/cli/lib/commands/create.js
jamesgeorge007: https://github.com/preactjs/preact-cli/blob/0ccce89fc2ed6e9367d5889f9c16f2f3cd89f988/packages/cli/lib/commands/create.js#L99

There isn't a `--dest` flag.

rschristian: Whoops, totally right. Forgot we used 'name' instead.

rschristian (approved):
LGTM

Issue #1531, opened 2021-02-06
Bug: Failing GET request upon fresh project creation


merlinpatt:
**Do you want to request a _feature_ or report a _bug_?**
Bug

**What is the current behaviour?**

Upon running `npx preact-cli create` and `npm run dev`, I see a failing GET request in the console, a network error response, and a "site cannot be installed" message.

```
⚛️Preact CLI development tip: A GET request just failed. This might fail for your users as well due to a network error. It may be worth adding runtimeCaching to your Service Worker.

The FetchEvent for "http://localhost:8080/" resulted in a network error response: an object that was not a Response was passed to respondWith().
Promise.then (async)
(anonymous) @ sw-debug.js:3

localhost/:1 Site cannot be installed: Page does not work offline. The page will not be regarded as installable after Chrome 93, stable release August 2021.
```

**If the current behaviour is a bug, please provide the steps to reproduce.**

1. Run `npx preact-cli create`
2. Run `npm run dev`
3. Open `http://localhost:8080/`
4. Open Chrome DevTools Console
5. See failing GET request, network error response, and "site cannot be installed" message

**What is the expected behaviour?**

There should be no failing GET request with a fresh install.

**Please mention other relevant information.**

This is the info from the Network tab of the failed request

```
General
Request URL: http://localhost:8080/
Referrer Policy: unsafe-url

Request Headers
Provisional headers are shown
Referer
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 11_2_0) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/89.0.4389.40 Safari/537.36
```

<img width="794" alt="Screen Shot 2021-02-05 at 21 58 00" src="https://user-images.githubusercontent.com/421960/107107161-51b2ef00-67fd-11eb-8b47-0c66e5e3d16c.png">

This is the file the network error response points to.

```
self.addEventListener('fetch', function (event) {
	var isPostRequest = event.request.method === 'POST';
	event.respondWith(
		fetch(event.request).catch(function (err) {
			if (err instanceof TypeError) {
				if (isPostRequest) {
					// eslint-disable-next-line
					console.log(
						'⚛️Preact CLI development tip: A POST request just failed. This might fail for your users as well due to a network error. It may be worth exploring the backgroundSync API.'
					);
				} else {
					// eslint-disable-next-line
					console.log(
[...]

PR #1534, opened 2021-02-07
fix: Correcting SW flag in dev

rschristian:

Bug fix

No

**Summary**

Running `yarn dev --sw` would create an error that `process` was not defined. This is because there was no `DefinePlugin` entry for `process.env.ES_BUILD` on dev. This was found here https://github.com/preactjs/preact-cli/issues/1531#issuecomment-774520544

Upon adding that, the service worker would attempt to be installed but run into errors upon registering. The solution for that seems to be adding the `InjectManifest` as a plugin, though this isn't ideal and throws a workbox warning (https://github.com/GoogleChrome/workbox/issues/1790). Kinda seems like we just shouldn't support the non-debug SW during dev? 

Additionally, this still fails if ran with the `--esm` flag, as no `sw-esm.js` is generated in dev. But that's probably a separate issue. 

No

### DIFF
--- a/packages/cli/lib/lib/webpack/webpack-client-config.js
+++ b/packages/cli/lib/lib/webpack/webpack-client-config.js
@@ -37,15 +37,45 @@ async function clientConfig(env) {
 		polyfills: resolve(__dirname, './polyfills'),
 	};
 
-	let swPath;
+	let swInjectManifest = [];
 	if (env.sw) {
-		swPath = join(__dirname, '..', '..', '..', 'sw', 'sw.js');
+		let swPath = join(__dirname, '..', '..', '..', 'sw', 'sw.js');
 		const userSwPath = join(src, 'sw.js');
 		if (existsSync(userSwPath)) {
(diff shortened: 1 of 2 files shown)


packages/cli/lib/lib/webpack/webpack-client-config.js
```diff
@@ -103,6 +117,38 @@ async function clientConfig(env) {
 					},
 				].filter(Boolean)
 			),
+			...(env.sw
```
ForsakenHarmony, line 120: this makes things a lot less readable

rschristian: The nested ternary? 

Yeah, sorry, can't remember why I did it like that. I'll try to fix that up tomorrow (3am here)

rschristian: Pushed an update, hopefully that's more readable.

ForsakenHarmony (approved):

PR #1564, opened 2021-05-28
Introduce `devServerRunning` hook

tooxie:

A build-related feature.

Yes.

**Summary**

Custom hooks are useful for other plugins to tap into certain points of the build process. In this case a `devServerRunning` hook is introduced which would allow a 3rd party plugin to, for example, display a QR code with the address the app is running on once the dev server started:

Usage:
```ts
import hooks from 'preact-cli/lib/lib/webpack/hooks';

hooks.devServerRunning.tap('QrCodePlugin', () => {
  printQrCode();
});
```

No.

**Other information**

```
Environment Info:
  System:
    OS: macOS 11.3.1
    CPU: (8) x64 Intel(R) Core(TM) i5-1038NG7 CPU @ 2.00GHz
  Binaries:
    Node: 12.13.0 - /usr/local/bin/node
    Yarn: 1.22.10 - /usr/local/bin/yarn
    npm: 6.12.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 91.0.4472.77
    Firefox: 88.0.1
    Safari: 14.1
  npmPackages:
    preact: ^10.3.1 => 10.5.13 
    preact-cli: ^3.0.0 => 3.0.5 
    preact-router: ^3.2.1 => 3.2.1 
  npmGlobalPackages:
    preact-cli: 3.0.5
```

### DIFF
--- a/packages/cli/lib/lib/webpack/hooks.js
+++ b/packages/cli/lib/lib/webpack/hooks.js
@@ -1,0 +1,5 @@
+const SyncHook = require('tapable').SyncHook;
+
+module.exports = {
+	devServerRunning: new SyncHook(),
+};
--- a/packages/cli/lib/lib/webpack/run-webpack.js
+++ b/packages/cli/lib/lib/webpack/run-webpack.js
@@ -10,6 +10,7 @@ const clientConfig = require('./webpack-client-config');
 const serverConfig = require('./webpack-server-config');
 const transformConfig = require('./transform-config');
 const { error, isDir, warn } = require('../../util');
+const { devServerRunning } = require('./hooks');
 
 async function devBuild(env) {
 	let userPort = parseInt(process.env.PORT || env.port, 10) || 8080;
@@ -68,6 +69,7 @@ async function devBuild(env) {
 			}
 
 			showStats(stats, false);
+			devServerRunning.call();
 		});
 
 		compiler.hooks.failed.tap('CliDevPlugin', rej);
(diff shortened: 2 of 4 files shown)


ForsakenHarmony:
Do any other projects have hooks like this in place?

tooxie:
Hi @ForsakenHarmony, thanks for taking a look into this PR. I have not looked at other projects, and honestly I am not sure which projects you are referring to, sorry. I would gladly check if you point me in the right direction.

Let me give you some context as to where this PR is coming from. I am using `preact-cli` to develop a PWA which for the moment is mobile only. It's not uncommon for me to change networks (home, office, public wifi, friend's, etc) so it's orders of magnitude more convenient to have a QR code I can scan every time I fire up the server instead of checking the IP the server is running on and type it into the phone. And by using [qrcode-terminal](https://www.npmjs.com/package/qrcode-terminal) the generation is extremely lightweight.

At first I thought about introducing the QR code generation into `preact-cli` but it felt that we could reuse some of the webpack QR generation libraries by introducing a hook, instead of baking the code in, and that tapping into the build process could be useful for some future use case I can't foresee. That's why I went down the hook route.

So that's the _rationale_ for this PR. Do you see a place for this feature in the project, either as a hook or generating the QR code directly in `preact-cli`? Or do you think this is out of scope?

Again, thanks a lot for taking the time to review my PR, and for all the great work you are all putting into this project :)

tooxie:
Hi @ForsakenHarmony, any comments on this? I would appreciate a response to know whether I should work on it further or just close it.

ForsakenHarmony:
@tooxie sorry for the late reply

I meant other clis like next.js / vue cli / create react app

Do they have a similar hook or another way to achieve this?


Also, I have to say `preact-cli` isn't really very well maintained right now as you might have noticed (partially my fault)
If you can switch to `wmr` that might be a good option for you

PR #1465, opened 2020-10-31
chore: update dependencies

ForsakenHarmony:

PR #1567, opened 2021-05-28
fix: Reverts fork-ts-checker version bump

rschristian:

Bugfix / partial revert of #1465

No

**Summary**

`fork-ts-checker-webpack-plugin` was broken as of v`3.2.0`, as it was upgraded a couple major versions without updating its config.

Unfortunately versions >= 5 seem to entirely lock our build process. I spent an hour trying to hunt down the reason but no luck. Hangs with no errors or logs to check. 

Reverting at least allows TS users to have a functional build tool.

No

### DIFF
--- a/.changeset/chilled-queens-live.md
+++ b/.changeset/chilled-queens-live.md
@@ -1,0 +1,5 @@
+---
+'preact-cli': patch
+---
+
+Reverts update to fork-ts-checker-webpack-plugin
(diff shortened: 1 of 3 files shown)


rschristian:
Seems like the pnpm ci is broken in general, nothing to do with this PR

ForsakenHarmony:
hmm, that's weird, it was passing on master

ForsakenHarmony (approved):

rschristian:
We don't have any TS tests and [the plugin only is enabled if there's a `tsconfig.json`](https://github.com/preactjs/preact-cli/blob/d6bde233e86fa6106ffd356c76a612109f530776/packages/cli/lib/lib/webpack/webpack-base-config.js#L341-L350)

PR #1568, opened 2021-05-28
fix: Un-breaks getLoadersByName helper

### LINKED ISSUE #1565: getLoadersByName helper always returns an empty array
When using the WebpackConfigHelper [getLoadersByName](https://github.com/preactjs/preact-cli/blob/master/docs/webpack-helpers.md#getloadersbyname) helper it always returns an empty array.

**Steps to Reproduce**
A basic reproduction would be to create a custom  `preact.config.js` and log out a helper you'd like to play with.
```
    const urlLoader = helpers.getLoadersByName(config, 'url-loader')
    console.log(urlLoader)
}
```

the loader requested to be returned.

This looks like it was introduced in preact-cli 3.1.0 (and is also present in 3.2.0).

A simple workaround is to use the `helpers.getLoaders(config)` helper, to get all the loaders, and manually try and find the loader you're after.

rschristian:

bugfix

No

**Summary**

Fixes #1565 

#1418 broke the helper and it wasn't until after it was merged that I noticed this

No

### DIFF
--- a/packages/cli/lib/lib/webpack/transform-config.js
+++ b/packages/cli/lib/lib/webpack/transform-config.js
@@ -235,7 +235,9 @@ class WebpackConfigHelpers {
 			)
 			.reduce((arr, loaders) => arr.concat(loaders), [])
 			.filter(
-				({ loader }) => loader === name || (loader && loader.loader === name)
+				({ loader }) =>
+					(typeof loader === 'string' && loader.includes(name)) ||
+					(typeof loader.loader === 'string' && loader.loader.includes(name))
 			);
 	}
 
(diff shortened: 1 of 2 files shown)


packages/cli/lib/lib/webpack/transform-config.js
```diff
@@ -235,7 +235,9 @@ class WebpackConfigHelpers {
 			)
 			.reduce((arr, loaders) => arr.concat(loaders), [])
 			.filter(
-				({ loader }) => loader === name || (loader && loader.loader === name)
+				({ loader }) =>
+					(typeof loader === 'string' && loader.includes(name)) ||
+					(typeof loader.loader === 'string' && loader.loader.includes(name))
```
rschristian, line 240: There very well could be a prettier way to express this but it's Friday and my brain is fried

ForsakenHarmony (approved):

PR #1570, opened 2021-05-29
ci: Attempting to fix Yarn 2 PNP test

rschristian:

CI fix

Existing cover this

**Summary**

https://github.com/yarnpkg/berry/issues/2935

Honestly Yarn 2 seems to be so broken in the ecosystem at large & unused I don't think it's worth the time testing against.

No


ForsakenHarmony (approved):

merceyz:
> Honestly Yarn 2 seems to be so broken in the ecosystem at large & unused I don't think it's worth the time testing against.

Ouch, sorry this issue made you come to that conclusion.

To give some context on what's going on here:
TypeScript has it's own resolution algorithm so patching `require.resolve`, which is usually enough for most tools, wasn't enough to give it PnP support. We opened https://github.com/microsoft/TypeScript/pull/35206 but progress on getting it merged was too slow for us so we resorted to applying the PR as a patch automatically when installing TypeScript, sadly the patch doesn't apply cleanly to all versions of TypeScript so we need to update it every now and then which is what happened here https://github.com/yarnpkg/berry/pull/2889.

rschristian:
Oh it's most certainly not just this issue, but the myriad of issues/discussions that have been raised (and raised again and again) about Yarn's future. 

I'm not against merging in any PR's that help compat like your own #1418, or even having code that does nothing but support Yarn 2, but I do disagree with spending any time ensuring compat ourselves. It's just not an effective use of time for a niche tool which breaks compat with a ton of the ecosystem.

merceyz:
We've fixed it on our side so this can be closed as the next CI run will pick up the fix on its own

PR #1574, opened 2021-05-30
Chore/remove safari nomodule polyfill

rschristian:

Chore, removing unnecessary code

No

**Summary**

Was reminded of Safari's `nomodule` polyfill that we have. Pretty unnecessary these days, [CanIUse lists 0.18% globally still affected](https://caniuse.com/es6-module) which I think is fine to remove automatic support for.

Users certainly could add the polyfill back in if they need it.

No


.changeset/lazy-fireants-think.md
```diff
@@ -0,0 +1,5 @@
+---
+'preact-cli': patch
```
developit, line 2: This seems possibly worth releasing as a minor.

developit (approved):
I think the time has come to remove this hack.

ForsakenHarmony (approved):

PR #1576, opened 2021-06-03
Remove typo'd package from package.json

feross:

`browserlist` does nothing but immediately throw. It should be removed from `package.json`.

No

**Summary**

No

**Other information**


developit (approved):
Good catch!

PR #1580, opened 2021-06-09
Generate ES2017 in development

developit:
This avoids the issue where Preact CLI generates broken loose-mode output in development, but generates the correct output in prod. With this patch landed, the code generated during development should use the same syntax as production. As part of this change, the `nomodule` bundles generated in production will no longer use `loose` mode, since this opens the possibility of breakage in IE.

### DIFF
--- a/.changeset/breezy-bottles-hammer.md
+++ b/.changeset/breezy-bottles-hammer.md
@@ -1,0 +1,5 @@
+---
+"preact-cli": minor
+---
+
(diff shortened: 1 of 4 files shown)


reznord (approved):

rschristian:
Hm, just running across this now, but we don't actually use https://github.com/preactjs/preact-cli/blob/master/packages/cli/babel/index.js ourselves... It's just there for users to add as a preset.

Probably all the more reason to consolidate this down, to avoid split configs.

PR #1608, opened 2021-11-09
chore: Updating html-webpack-plugin to v4

### LINKED ISSUE #1551: Can't use with storybook due to `html-webpack-plugin` issues
Looks like trying to use this with `storybook` causes builds to stop working, because storybook uses `html-webpack-plugin@4` and `preact-cli` uses version 3, clashing with each other. When I run `preact build` while both are installed, I get:

```

17:53:39.301 | Error: The expected HtmlWebpackPlugin hook was not found! Ensure HtmlWebpackPlugin is installed and was initialized before this plugin.
-- | --
17:53:39.301 | at HtmlWebpackExcludeAssetsPlugin.applyCompilation (/vercel/workpath0/node_modules/html-webpack-exclude-assets-plugin/index.js:25:13)
17:53:39.301 | at SyncHook.eval [as call] (eval at create (/vercel/workpath0/node_modules/tapable/lib/HookCodeFactory.js:19:10), <anonymous>:23:1)
17:53:39.302 | at SyncHook.lazyCompileHook (/vercel/workpath0/node_modules/tapable/lib/Hook.js:154:20)
17:53:39.302 | at Compiler.newCompilation (/vercel/workpath0/node_modules/webpack/lib/Compiler.js:631:26)
17:53:39.302 | at /vercel/workpath0/node_modules/webpack/lib/Compiler.js:667:29
17:53:39.302 | at AsyncSeriesHook.eval [as callAsync] (eval at create (/vercel/workpath0/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:6:1)
17:53:39.302 | at AsyncSeriesHook.lazyCompileHook (/vercel/workpath0/node_modules/tapable/lib/Hook.js:154:20)
17:53:39.302 | at Compiler.compile (/vercel/workpath0/node_modules/webpack/lib/Compiler.js:662:28)
17:53:39.302 | at /vercel/workpath0/node_modules/webpack/lib/Compiler.js:321:11
17:53:39.302 | at Compiler.readRecords (/vercel/workpath0/node_modules/webpack/lib/Compiler.js:529:11)
17:53:39.302 | at /vercel/workpath0/node_modules/webpack/lib/Compiler.js:318:10
17:53:39.302 | at eval (eval at create (/vercel/workpath0/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:11:1)
17:53:39.302 | at run (/vercel/workpath0/node_modules/fork-ts-checker-webpack-plugin/lib/index.js:204:13)
17:53:39.302 | at AsyncSeriesHook.eval [as callAsync] (eval at create (/vercel/workpath0/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:7:1)
17:53:39.302 | at AsyncSeriesHook.lazyCompileHook (/vercel/workpath0/node_modules/tapable/lib/Hook.js:154:20)
17:53:39.302 | at /vercel/workpath0/node_modules/webpack/lib/Compiler.js:315:19

```

and then because of how `preact` works the github action/vercel build i run just goes forever and then times out after 6 hours (expensive!)

Related: https://github.com/jamesjieye/html-webpack-exclude-assets-plugin/issues/18

**Steps to Reproduce**
create a preact app, then install storybook, and then try to run `yarn build`

the build works

i'll try to add a minimal repro soon.

rschristian:

Bumping `html-webpack-plugin` to v4

Existing should cover

**Summary**

Rebase of #1466, with some additional changes to finish things off

Closes #1551

~~Need a new release from critters, now that https://github.com/GoogleChromeLabs/critters/pull/87 has been merged.~~

Possibly need action on/wait for https://github.com/GoogleChromeLabs/critters/issues/88 if my work around isn't viable?


ForsakenHarmony:
Yeah, I think critters is the issue I ran into as well

rschristian:
Yeah, it didn't support v4 until a couple weeks ago.

With that support there was a significant refactor that still has some bugs that need to be fixed, so gotta wait some more :/

packages/cli/lib/lib/webpack/webpack-client-config.js
```diff
@@ -254,7 +254,7 @@ function isProd(config) {
 				preload: 'media',
 				pruneSource: false,
 				logLevel: 'silent',
-				additionalStylesheets: ['*.css'],
+				additionalStylesheets: ['route-*.css'],
```
rschristian, line 257: Critters has always duplicated the styles in `bundle.css` due to it too matching this pattern, but historically has deduped through using `cssnano` as means to compress. `cssnano` is no longer in use, so we're getting duplicate CSS. See [critters/88](https://github.com/GoogleChromeLabs/critters/issues/88).

Not yet sure if this is a regression to be fixed in Critters or not, but we can get around it by only matching chunk split CSS, which I believe is the intention anyhow?

rschristian:
So PNP seems to get broken for unknown reasons, and the build sizes run in opposite directions depending on if ESM is enabled, but close!™

rschristian:
So this is a bit baffling, but it seems Terser will rather inconsistently use absolute or relative paths to the sources in the source maps.

Locally, for me, it's always absolute. There's no deviation, all files are absolute. The CI is where things get fun. `route-home.chunk.*.js.map` uses relative with CJS, but absolute on the ESM build. `route-home.chunk.*.esm.js.map` on the other hand uses relative paths again. This inconsistency results in tests that do not pass on my local machine, though I've somehow got the CI to pass. 

However, even with the CI passing, the source maps that are generated with absolute paths are broken and point to sources that do not exist.

I've set up a test on the CI:

- This is the directory that is output from a test
    `/home/runner/work/preact-cli/preact-cli/packages/cli/tests/output/mhlpx/build`
  - [link](https://github.com/preactjs/preact-cli/runs/4330990953?check_suite_focus=true#step:7:98)
- This is (part of) the contents of `route-home.chunk.js.map` from that build
    ```
    "sources": [
      "webpack:////home/runner/work/preact-cli/preact-cli/packages/cli/tests/output/fleqbvuazw/src/routes/home/style.css",
      "webpack:////home/runner/work/preact-cli/preact-cli/packages/cli/tests/output/fleqbvuazw/src/routes/home/index.js"
    ]
    ```
  - [link](https://github.com/preactjs/preact-cli/runs/4330990953?check_suite_focus=true#step:7:98)

Seems like there's some sort of bizarre caching issues causing Webpack/Terser to point at sources that were used in previous runs.

rschristian:
@merceyz Sorry to tag, but figure you might have an interest in this as you set up the PnP compat in the first place. Getting a build error that seemingly is only reproducible on PnP. I have no idea why this might be and lack enough knowledge on anything Yarn beyond v1 to know what's going wrong here.

[I've set up a demo repo](https://github.com/rschristian/preact-cli-yarn-pnp-bug) that reproduces the error the CI is seeing, with `@rschristian/junk` just being the latest state of this branch here. `yarn link ...` certainly shows the error too, but I wanted to ensure it wasn't a linking issue.

merceyz:
Hey, thanks for the ping and the reproduction 👍 .

The issue is that `critters-webpack-plugin` doesn't declare `html-webpack-plugin` as a dependency. They also [wrapped the `require` call in `try ... catch`](https://github.com/GoogleChromeLabs/critters/blob/5e32531caf7540842ff9af99e9449975bf0ad399/packages/critters-webpack-plugin/src/index.js#L71-L75) so the actual error is hidden but by running the following you can see the error:
```sh
$ PNP_DEBUG_LEVEL=1 yarn build
Error: critters-webpack-plugin tried to access html-webpack-plugin, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.

  Required package: html-webpack-plugin (via "html-webpack-plugin")
  Required by: critters-webpack-plugin@npm:3.0.1 (via C:\cache\critters-webpack-plugin-npm-3.0.1-32c4695844-7c0.zip\node_modules\critters-webpack-plugin\dist\)
```

I made a PR to fix the issue upstream (https://github.com/GoogleChromeLabs/critters/pull/91), you can add the following command to the e2e test to make it pass before that is merged/released.
```sh
yarn config set "packageExtensions.critters-webpack-plugin@*.peerDependencies.html-webpack-plugin" "*"
```

rschristian:
Ah shoot, didn't catch that.

Thanks so much, for the Critters PR and fix for here!

packages/prerender-data-provider/src/hook.js
```diff
@@ -38,8 +38,8 @@ function usePrerenderData(props, doAutomaticFetch = true) {
 		}
 	}
 
-	if (contextValue.CLI_DATA && contextValue.CLI_DATA.preRenderData) {
-		value = contextValue.CLI_DATA.preRenderData;
+	if (contextValue) {
+		value = contextValue;
```
ForsakenHarmony, line 42: oh, why did this change?

packages/cli/tests/subjects/custom-webpack/template.html
```diff
@@ -2,13 +2,11 @@
 <html lang="en">
 	<head>
 		<meta charset="utf-8">
-		<title><%= htmlWebpackPlugin.options.title %></title>
+		<title><%= cli.title %></title>
```
ForsakenHarmony, line 5: Guess this is a bit of a breaking change, do you think we can just publish it as a minor?

ForsakenHarmony (approved):
LGTM

Do you think it's ready to merge?

packages/cli/tests/subjects/custom-webpack/template.html
rschristian: Ah shoot, good catch. Definitely can't get away with that.

I don't think I touched any of this, so not really familiar with how the plugin actually works, but I'll see if I can wrangle this into a non-breaking change. No idea if that's possible, but we'll see.

rschristian:
@ForsakenHarmony 

One thing that is worth bringing up is the file size matching. Rebasing on your work means we have file size checks again (yay!) but it seems Webpack chooses to use absolute or relative file paths in the source maps based on whichever is shorter. This becomes very system dependent and I haven't seen anyway to force one or the other.

For example, while the CI is passing just fine, this entry creates a problem locally:

https://github.com/preactjs/preact-cli/blob/30e17aa958ceeed52aadf4ad4d9cebffb9c9edf6/packages/cli/tests/images/build.js#L23

My local size for this entry is 702b, which is much greater than the allowed 577.5b (+5%).

Not really sure what to do about this.

packages/cli/tests/subjects/custom-webpack/template.html
rschristian: Okay, was able to fix this. It's a tad messy in the source, but should be non-breaking. Tested by adding adding a `<%= JSON.stringify(htmlWebpackPlugin.options) %>` and comparing, more or less.

rschristian: Updated the `custom-template` test to ensure access via `htmlWebpackPlugin.options...` still works as intended.

https://github.com/preactjs/preact-cli/blob/66ab462654c56dcee3ee89a24b7f3e8d2351c96f/packages/cli/tests/subjects/custom-template/template.html#L6-L8

However, **there are still a couple differences on the `htmlWebpackPlugin` object that are breaking**, but I'm tempted to ignore these as they require a very specific use case in order to surface and should be trivial to fix. [preact-www runs into one of these](https://github.com/preactjs/preact-www/blob/5e23416a196929a81ef606beb0c59b7c989d8284/src/template.html#L46), as `htmlWebpackPlugin.files.chunks` no longer exists.

Here's a cleaned up version of the diff between master (left) and this PR (right): https://www.diffchecker.com/E1JDpVpK

packages/cli/tests/build.test.js
```diff
@@ -217,28 +222,8 @@ describe('preact build', () => {
...
-	it('should replace title with <%= preact.title %>', async () => {
-		let dir = await subject('custom-template-3');
-		await build(dir);
-
-		let file = join(dir, 'build/index.html');
-		let html = await readFile(file, 'utf-8');
-
-		looksLike(html, images.templateReplaced);
-	});
-
```
rschristian, line 239: I combined these tests into one. Should check all the same things, but it'll be slightly faster/less busy.

ForsakenHarmony:
@rschristian 
> We can force absolute paths, and this seems to pass my local suite and the CI. Seems like a bad solution though, as any user with a deep enough clone of this repo in their filesystem will likely run into issues.

Why do you think it's an issue?

rschristian:
Because it's entirely system dependent.

If you clone this repo to `/some/really/really/long/path/thats/deep/in/your/file/system/preact-cli`, you will not be able to get the tests to pass locally. The path bloats the size of the source map. Same thing goes for `/short/path/preact-cli`.

The CI and my local setup (`/home/ryan/Projects/preact-cli`) happen to be close enough to get a pass if we force absolute paths for all source maps, but that's just chance.

I suppose we can strip paths out of source maps and then compare. Little bit weird, but I don't imagine that could be problematic. Will try that out.

PR #1612, opened 2021-11-13
Fix/port config precedence

### LINKED ISSUE #1596: $PORT preferred over --port

If I type:
```
▶ PORT=7777 preact watch -p 9999
```
it starts on port :7777. Not :9999

```
Compiled successfully!

You can view the application in browser.

Local:            http://0.0.0.0:7777
On Your Network:  http://192.168.1.192:7777
```

**Steps to Reproduce**
Hopefully clear from the description above. 

I would expect the `-p` option to trump the environment variable. 
Typing `PORT=7777 preact watch -p 9999` is absurd but typing `preact watch -p 9999` is not and then you get negatively surprised that :9999 did not become the port. 

My use case was that I was using [`foreman`](https://www.npmjs.com/package/foreman) to start a bunch of things, including `preact watch` and for some reason it sets `$PORT=5000` in there. I don't know why and nor is it very relevant.

Is it just me? I feel like the `-p` (or `--port`) is a lot more explicit and should be trumped over the environment variable.

rschristian:

Refactor to config precedence

Yes

**Summary**

Closes #1596

Determining the port is a tad different now:

`--port` -> `$PORT` -> `8080` -> `<random>`

Far left is preferred, far right is the fall back.

The other change is that if a user explicitly selects a port with `-p/--port` that is already taken, we throw an error and exit. That scenario is almost certainly a mistake, either in something running on a port it shouldn't or the wrong port passed to preact-cli. We will continue to fall back to random if `$PORT` is already set and already in use, with the idea being that it's probably another service setting and using that value.

I believe this makes a bit more sense, though could just be my opinion. Let me know!

No


packages/cli/lib/util.js
```diff
@@ -54,3 +55,26 @@ exports.normalizeTemplatesResponse = function (repos = []) {
...
+			const server = net.createServer();
+			server.unref();
+			server.on('error', reject);
+			server.listen({ port }, () => {
+				server.close(resolve);
+			});
+		});
+		return true;
+	} catch (err) {
+		if (err.code !== 'EADDRINUSE') throw err;
```
ForsakenHarmony, line 77: this works cross platform?

ForsakenHarmony (approved):

packages/cli/lib/util.js
rschristian: Should be. [Node's `server.listen()` docs](https://nodejs.org/api/net.html#serverlisten) list it as a common thing to check for, which would be odd if it wasn't cross platform.

I was under the impression that Node's error codes were all cross platform anyways. The docs don't seem to list anything on [this page](https://nodejs.org/api/errors.html#common-system-errors), (common system error or no), as being platform-specific. Could be wrong though?

rschristian: Found a random [StackOverflow question of someone getting it on Windows](https://stackoverflow.com/questions/41716833/getting-an-error-listen-eaddrinuse-3000-on-windows-machine/41717870), which I imagine would be the problem child if there was one.

Should be fine to check for that.

PR #1613, opened 2021-11-13
chore: Misc cleaning

rschristian:

Chore

N/A

**Summary**

* ReadMe
  * Corrected required Node version in the ReadMe
  * Removed the deployment section as it hasn't been valid or useful for a while
  * Edited the  `--refresh` flag to match the CLI description
  * Added in missing table of contents entries for env vars & route-based code splitting
* Updated `yarn.lock` as it had some outdated items that we no longer used
* Updated browserslist
* Cleared out a couple TODOs that have sat untouched for years and would never get done.
* Removes `setHtmlTemplate` helper
  * This has been non-functional for a fairly long time, is redundant, and no one has seemed to notice, so a silent removal shouldn't be problematic?

~~Note: Two tests **should** be failing at the moment, due to the `.babelrc` sat in the default template. [I've opened a PR to remove it](https://github.com/preactjs-templates/default/pull/47). The tests here should fail until that's gone.~~ Merged, all good now

N/A


ForsakenHarmony (approved):

ForsakenHarmony:
should this have a changeset?

rschristian:
Yes, yes it should. Sorry about that.

PR #1617, opened 2021-11-18
Enable regenerator in legacy bundles

developit:

Transpiles generators to regenerator in legacy bundles.

**Summary**

In 3.3.0, we stopped using `fast-async` to transpile async functions, however this meant they were transpiled to generators, which we don't transpile at all. The modern bundles should be unaffected by all of these changes, so let's enable regenerator to handle both async and generator functions in legacy bundles.

No, it fixes an unintentionally-breaking change.

### DIFF
--- a/packages/cli/lib/lib/babel-config.js
+++ b/packages/cli/lib/lib/babel-config.js
@@ -10,8 +10,7 @@ module.exports = function (env, options = {}) {
 					modules: options.modules || false,
 					targets: {
 						browsers: options.browsers,
-					},
-					exclude: ['transform-regenerator'],
+					}
 				},
 			],
 		],


JoviDeCroock (approved):

developit:
Not sure how I managed to make the Service Worker tests fail with this PR.

rschristian (approved):

rschristian:
Seems both service workers (esm and legacy) are having regenerator inserted in. I imagine they should be excluded?

Not sure why that's a problem, but seems to be it.

developit:
@rschristian I was thinking that was the case, yeah. Neither should have it, or at least the modern one shouldn't have it (the legacy sw file never gets used, no browsers support SW that don't also support ESM).

developit:
Darn, function `Use` entries are only supported as of Webpack 4

packages/cli/lib/lib/webpack/webpack-base-config.js
```diff
@@ -182,10 +183,14 @@ module.exports = function createBaseConfig(env) {
 					loader: require.resolve('babel-loader'),
 					options: Object.assign(
 						{ babelrc: false },
-						createBabelConfig(env, { browsers }),
+						createBabelConfig(env, {
+							browsers: info.compiler.name === 'InjectManifest'
```
ForsakenHarmony, line 187: what exactly does this do?

ForsakenHarmony (approved):

packages/cli/lib/lib/webpack/webpack-base-config.js
developit: It's supposed to keep modern syntax in both generated `sw.js` outputs, since service worker is only supported by fairly modern browsers with generator/async support anyway.

ForsakenHarmony: ah, right

ForsakenHarmony: maybe it would make sense to add a comment for future reference

packages/cli/lib/lib/webpack/webpack-base-config.js
```diff
@@ -173,7 +174,7 @@ module.exports = function createBaseConfig(env) {
 
 		module: {
 			rules: [
-				{
+				(info) => ({
```
ForsakenHarmony, line 177: The tests seem to be failing because this is a function

developit: Is this only supported in newer Webpack?

rschristian: No, isn't supported in newer Webpack either. Rules have to be objects.

PR #1618, opened 2021-11-19
Fix/sw src

rschristian:

Bug fix

No

**Summary**

With ESM disabled, `InjectManifest` used `join(src, sw.js)` to get the service worker path. If this did not exist (if the user [re]moved the file), the build would error out.

We already had a system set up to catch this and fall back to the one provided from `preact-cli`, just weren't using it on non-esm builds.

No

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -127,7 +127,7 @@
     "stack-trace": "0.0.10",
     "style-loader": "^2.0.0",
     "terser-webpack-plugin": "^4.2.3",
-    "typescript": "^4.2.4",
+    "typescript": "~4.2.4",
     "update-notifier": "^5.1.0",
     "url-loader": "^4.1.1",
     "validate-npm-package-name": "^3.0.0",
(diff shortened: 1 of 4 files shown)


packages/cli/package.json
```diff
@@ -127,7 +127,7 @@
     "stack-trace": "0.0.10",
     "style-loader": "^2.0.0",
     "terser-webpack-plugin": "^4.2.3",
-    "typescript": "^4.2.4",
+    "typescript": "~4.2.4",
```
rschristian, line 130: This is completely unrelated so I could move into a separate PR if desired.

From what I understand, Yarn needs to manually patch some dependencies like TS on every release. TS 4.5 just came out, which apparently caused some issues in the PnP pipeline.

Shouldn't be an issue to just restrict this to a patch/TS minor

developit: eh, if it makes CI pass leave it in.

developit (approved):

PR #1623, opened 2021-11-24
fix: Test suite now uses config files again

rschristian:

Test suite bug fixes & chores

Yes

**Summary**

Found in #1622, as it turns out, none of the config files (`preact.config.js`) have been loaded in our test suite in a fairly long time. [The `custom-webpack` subject calls a method that doesn't exist anymore](https://github.com/preactjs/preact-cli/blob/master/packages/cli/tests/subjects/custom-webpack/preact.config.js) and yet the test isn't throwing. The config files aren't even being used.

This PR corrects that, though the implementation is a bit rough? Basically if `esm` cannot provide the config we simply `require()` it.

The behavior is a bit odd though and I don't quite understand what the issue is. `export default function () {...}` worked with our previous set up, but `export default () => {...}`, `module.exports = function () {...}`, and `module.exports = () => {...}` did not. With this fix, the two `module.exports` formats work in the suite again, but `export default () => {...}` does not. Not sure why this might be.

No

### DIFF
--- a/packages/cli/lib/lib/webpack/transform-config.js
+++ b/packages/cli/lib/lib/webpack/transform-config.js
@@ -112,9 +112,7 @@ module.exports = async function (env, webpackConfig, isServer = false) {
 	// The line above results in an empty object w/ Jest,
 	// so we need to do the following in order to load it:
 	if (Object.keys(m).length === 0) {
-		try {
-			m = require(myConfig);
-		} catch {}
+		m = require(myConfig);
 	}
 
 	const transformers = parseConfig((m && m.default) || m);
--- a/packages/cli/lib/lib/webpack/transform-config.js
+++ b/packages/cli/lib/lib/webpack/transform-config.js
@@ -107,7 +107,15 @@ module.exports = async function (env, webpackConfig, isServer = false) {
 		);
 	}
 
-	const m = require('esm')(module)(myConfig);
+	let m = require('esm')(module)(myConfig);
+
+	// The line above results in an empty object w/ Jest,
+	// so we need to do the following in order to load it:
+	if (Object.keys(m).length === 0) {
+		try {
+			m = require(myConfig);
+		} catch {}
+	}
 
 	const transformers = parseConfig((m && m.default) || m);
 
--- a/packages/cli/tests/lib/output.js
+++ b/packages/cli/tests/lib/output.js
@@ -1,3 +1,4 @@
+const { existsSync, mkdirSync } = require('fs');
 const copy = require('ncp');
 const { resolve } = require('path');
 const { promisify } = require('util');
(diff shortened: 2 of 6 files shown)


packages/cli/tests/lib/output.js
```diff
@@ -16,6 +17,9 @@ function tmpDir() {
 async function subject(name) {
 	let src = resolve(subjects, name);
 	let dest = tmpDir();
+	if (!existsSync(dest)) {
+		mkdirSync(dest, { recursive: true });
+	}
```
rschristian, line 22: Running individual test suites (i.e., `yarn test tests/build.test.js`) results in the output dir not being created and the tests exiting when trying to copy the subjects over. This corrects that.

packages/cli/tests/subjects/preload-chunks/preact.config.js
```diff
@@ -1,6 +0,0 @@
-module.exports = config => {
-	config.optimization.splitChunks = {
-		minSize: 0,
-	};
-	return config;
-};
```
rschristian, line 6: Deleted as unnecessary. We already test that preload is working fine without additional config.

packages/cli/tests/lib/output.js
ForsakenHarmony: wondering if it would make sense to move this into `tmpDir`

packages/cli/lib/lib/webpack/transform-config.js
```diff
@@ -107,7 +107,15 @@ module.exports = async function (env, webpackConfig, isServer = false) {
...
 
-	const m = require('esm')(module)(myConfig);
+	let m = require('esm')(module)(myConfig);
+
+	// The line above results in an empty object w/ Jest,
+	// so we need to do the following in order to load it:
+	if (Object.keys(m).length === 0) {
+		try {
+			m = require(myConfig);
+		} catch {}
```
ForsakenHarmony, line 117: So this is just making sure that it's actually empty? (and we don't care about an error in that case)

ForsakenHarmony (approved):

packages/cli/lib/lib/webpack/transform-config.js
rschristian: That was the idea, yeah. By this point we've already ensured the file exists so if it's imported in some capacity and still empty, that's either a file with no exports or whatever this issue is with `esm` in our test suite. If a user does provide us with an empty config, that shouldn't be an issue.

The `try ... catch` I believe is unnecessary and I'll remove in a second. Left over fragment that I don't believe will do anything ever.

packages/cli/tests/lib/output.js
rschristian: I do think that makes more sense

PR #1624, opened 2021-11-25
Feat/config types

rschristian:

feature (types)

N/A

**Summary**

Provides types for import that can help users with their config files. While this won't provide types for the entire default config, it does help correct a number of mistakes and/or typos.

Side note: I'm partial to moving the `docs/webpack-helpers.md` into the wiki, just to keep all that stuff together. Thoughts?

No


.eslintrc
```diff
@@ -40,5 +40,13 @@
...
-  }
+  },
+  "overrides": [
+    {
+      "files": ["**/*.ts"],
+      "rules": {
+        "no-undef": "off"
+      }
+    }
+  ]
```
rschristian, line 51: Added to correct ESLint claiming `Record` is undefined in the typings

README.md
```diff
@@ -219,35 +219,17 @@ To customize preact-cli create a `preact.config.js` or a `preact.config.json` fi
...
-	 * Supports asynchronous changes when a promise is returned (or it's an async function).
-	 *
-	 * @param {object} config - original webpack config.
-	 * @param {object} env - options passed to the CLI.
-	 * @param {WebpackConfigHelpers} helpers - object with useful helpers for working with the webpack config.
-	 * @param {object} options - this is mainly relevant for plugins (will always be empty in the config), default to an empty object
-	 **/
-	webpack(config, env, helpers, options) {
-		/** you can change the config here **/
-	},
```
rschristian, line 250: I think this format should be moved into the wiki, to keep it documented, but not the first thing people come across. I figure most beginners just want a simple set up to copy/paste, and this is a lot more busy than needs be.

Might just be me though? Would like thoughts on this.

packages/cli/lib/lib/webpack/transform-config.js
```diff
@@ -130,35 +130,15 @@ module.exports = async function (env, webpackConfig, isServer = false) {
 	}
 };
 
-/**
- * WebpackConfigHelpers
- *
- * @class WebpackConfigHelpers
- */
```
rschristian, line 137: I removed this JSDoc comments as they weren't particularly descriptive of types and, as far as I know how to wrangle the TS language server, there's no way they'd be picked up on in any user project.

packages/cli/package.json
```diff
@@ -80,6 +81,7 @@
     "@preact/async-loader": "^3.0.1",
     "@prefresh/babel-plugin": "^0.4.1",
     "@prefresh/webpack": "^3.2.2",
+    "@types/webpack": "^4.38.0",
```
rschristian, line 84: Needed to provide types for `helpers.webpack`. Might be overkill though? Could just `Record<string, unknown>` it.

packages/cli/types.d.ts
```diff
@@ -0,0 +1,127 @@
...
+	dest: string;
+	esm: boolean;
+	sw: boolean;
+	dev: boolean;
+	production: boolean;
+	isProd: boolean;
+	ssr: boolean;
+	prerender: boolean;
+	[key: string]: any;
+};
```
rschristian, line 75: Tried to cover the commonly used ones that I've seen. Did I miss any that should be added?

packages/cli/package.json
ForsakenHarmony: Looking at the lock file I think it's fine to add

README.md
ForsakenHarmony: Yeah, you're probably right

not like we have a lot of plugins

packages/cli/lib/lib/webpack/transform-config.js
ForsakenHarmony: Won't these help if you use typescript to check these JS files?

ForsakenHarmony (approved):

packages/cli/lib/lib/webpack/transform-config.js
rschristian: Hm, well they could help ensure you're returning the correct type, but, at least to my knowledge, without a js/tsconfig set up the TS language server won't type check.

In my editor, I'd need to create a `jsconfig.json` in the project in order for say `getPluginsByType` to show an error when I replace the return value with `'foo'`.

This might be editor-specific though. I'll revert this, and just switch them over to use the types created in this PR (as they're more specific)

ForsakenHarmony: Sounds good

docs/webpack-helpers.md
```diff
@@ -1,154 +0,0 @@
-### Table of Contents
```
rschristian, line 1: (Repeating the comment from above as it seems to have been read over)

I think it would be better to move this content into the wiki to keep everything together a bit more. Currently you have to switch between this doc and the wiki for some config information which is a bit awkward in my experience.

Certainly open to other thoughts, maybe we should move the common config recipes into a doc instead? Not sure.

Issue #1579, opened 2021-06-08
Webpack 5 support


mhauens:
As preact-cli is still using webpack@4.46.0, ist there a plan to upgrade to latest webpack 5 version?

rschristian:
As far as I know there is no current plan in motion for doing so, no. It'd certainly be a nice-to-have, but there's at least a few dependencies that need some work / reevaluation and are holding us back from updating as they're not compatible. And that's not taking into account the work needed to make that upgrade happen, or if anyone is willing to do it.

Most active development is over on [WMR](https://github.com/preactjs/wmr) at the moment.

adroitwhiz:
Webpack v4 is broken on Node 17 and up with no plans to fix it (https://github.com/webpack/webpack/issues/14532), so if preact-cli is in fact unmaintained and there's no work being put into upgrading to Webpack 5, an official deprecation notice would be nice.

rschristian:
Preact-CLI isn't necessarily unmaintained, as you can see via the commit history, work has been done recently. But keeping up with the rapidly changing Webpack ecosystem is extremely time consuming on the small group of volunteer maintainers (see https://github.com/preactjs/wmr/issues/867#issuecomment-916956523 for some great context)

I don't think anyone is ready to deprecate yet (or I haven't heard anything at least) but community contributions are certainly always welcome. If anyone has some experience or wants to put together a PR, I'm sure people would be happy to review.

I believe I had a WIP sitting around, so I might be able to take a look at it over the coming week.

Appreciate the heads-up on that Webpack issue! Good to know that might be a problem.

ForsakenHarmony:
> Webpack v4 is broken on Node 17 and up with no plans to fix it ([webpack/webpack#14532](https://github.com/webpack/webpack/issues/14532)), so if preact-cli is in fact unmaintained and there's no work being put into upgrading to Webpack 5, an official deprecation notice would be nice.

As long as it works with Node 16 I don't see any reason to do so right now, Node 17 is "Current" and will never be LTS. ("Production applications should only use Active LTS or Maintenance LTS releases.").

With Node 18 however, things might change, and it would suck if we can't support it.

FrantzUml:
Sure preact-cli is obsolete with many dependencies package deprecated (webpack, jest, html-webpack-plugin, chokidar, even preact itself, and so on).
For example, using preact-router@3.0.1 instead of @4.1.0 makes error as vnode.attributes is replaced by vnode.props, and so on.

[...]

PR #1645, opened 2022-01-30
feat: webpack v5

rschristian:
**Summary**

Major changes (so far):

- Webpack v5
- Switches to [`optimize-plugin`](https://github.com/developit/optimize-plugin) for dual output
  - `polyfills.js` is now `dom-polyfills.js` to differentiate as `optimize-plugin` creates es polyfills
- Removes `--esm` flag
- `.babelrc` now no longer overrides matching keys in the default config
- Increases default browserslist config to `> 0.5%, last 2 versions, Firefox ESR, not dead` (which is browserslist's default)


rschristian:
`import()` segfaults in Jest at the moment, which means any of our tests that try to use `copy-webpack-plugin` segfault. This is only a few, and most are actually passing. Might be worth skipping some tests/editing them to avoid this issue for now, seems like something that can be resolved upstream without our involvement.

https://github.com/nodejs/node/issues/35889

developit:
@rschristian I think it's okay to drop Preact v8 support from the next major release of Preact CLI.

Regarding `babel-esm-plugin`, I wonder if we could switch to [optimize-plugin](https://npm.im/optimize-plugin)?

```js
new OptimizePlugin({
    modernize: false
})
```

rschristian:
@developit I only took a quick look, but ran into a couple potential issues:

1) I couldn't get it to pick up a babel config for the life of me. Would error out on JSX even with a JSX transform plugin. 

Not really sure what's going on with that, but even if it were working, it seems like all users would need a babel config in their project (even if it just points at a default one we distribute)? That doesn't sound ideal, would much rather pass a babel config directly to the plugin. Fewer config files for the users and it'd be easier for us to conditionally add plugins and whatnot.

2) We'd probably want to skip certain assets like the service workers, but there doesn't seem to be any blacklist of plugins/assets.

Will try to take a better look tomorrow. `babel-esm` actually doesn't look too far off if we want to keep with it, the only removed plugin is `MultiEntries` which we don't make use of here.

packages/cli/lib/lib/webpack/webpack-client-config.js
```diff
@@ -146,7 +146,12 @@ async function clientConfig(env) {
 			}),
 			new PushManifestPlugin(env),
 			...(await renderHTMLPlugin(env)),
-			...getBabelEsmPlugin(env),
+			env.esm &&
+				new OptimizePlugin({
+					modernize: false,
+					minify: isProd,
+					exclude: [/^sw.*\.js/],
+				}),
```
rschristian, line 154: I was thinking, do we want to keep having users opt in/out with the `--esm` flag, or should this instead be something they can disable with their `preact.config.js` files?

The only real argument against outputting ESM too is, AFAIK, a desire to have fewer files/extreme disk space constraints. I don't think that's a common enough of a problem to continue to have this as a CLI flag.

rschristian: Hmm, so it looks like `optimize-plugin` is quite slow and doesn't play nice in dev. We'd have to remove the flag for `preact watch`, but prod is up in the air.

rschristian: Ended up removing, it's honestly simpler to have users cut out `optimize-plugin` in their config and adjust Terser's file matching. We can also push towards better practices.

packages/cli/lib/lib/babel-config.js
```diff
@@ -3,7 +3,7 @@ module.exports = function (env, options = {}) {
```
rschristian, line 3: Someone who really knows what they're doing probably needs to take a deep look at this. I think the general idea is to remove non-syntax stuff for `optimize-plugin`, but I'm not confident I didn't totally bugger up someone's app.

.github/workflows/ci.yml
```diff
@@ -44,36 +44,37 @@ jobs:
...
-        run: |
-          git clone https://github.com/preactjs-templates/default.git default
-          cd default/template
-          touch yarn.lock
-          echo $(cat package.json | jq '.name = "pnp-test"') > package.json
-          yarn set version 2
-          yarn config set pnpFallbackMode none
-          yarn config set compressionLevel 0
-          yarn link -A -p ../..
-          yarn build
```
rschristian, line 69: I counted 6-8 upstream dependencies that throw issues in strict PnP. There's no reason to let that hold things up.

packages/cli/tests/build.test.js
```diff
@@ -1,7 +1,8 @@
 const { join } = require('path');
```
rschristian, line 1: This test suite was not instilling much confidence in me, hence all these changes.

I will be extracting this all out into a separate test suite PR, to ensure this one doesn't get out of hand / bogged down in irrelevant work, but I wanted to stage it here so I get a better idea of where we're at while I build it out.

...and I want to delay rebasing this monster for as along as I can.

packages/cli/lib/lib/webpack/render-html-plugin.js
```diff
@@ -112,7 +112,6 @@ module.exports = async function renderHTMLPlugin(config) {
 				};
 			},
 			inject: true,
-			scriptLoading: 'defer',
```
rschristian, line 115: Is default val: https://github.com/jantimon/html-webpack-plugin#options

rschristian:
We do have to wait for https://github.com/developit/optimize-plugin/pull/25 to be published (using local patch right now) but thought I'd mark this open for review anyhow as that's not really impeding a merge, only a publish of a new version of CLI.

**This is just for Webpack v5 upgrade** and some changes that are necessitated by the upgrade or most convenient to do at the same time.

I've tried to keep this as minimal as I could to that upgrade, apologies for it still being a rather large diff.

developit:
@rschristian I just published [optimize-plugin 1.3.0](https://github.com/developit/optimize-plugin/releases/tag/1.3.0) with the fix(es). Sorry for leaving that open forever!

rschristian:
Thanks, and no worries! Certainly didn't block any work here

ForsakenHarmony (approved):

rluvaton:
Great PR! any idea when it should be in stable npm version?

rschristian:
> Great PR! any idea when it should be in stable npm version?

A fair way off at best. Would need to go into a new major, and there's some other things I'd like to include in that major.

rluvaton:
Thanks man!

kay-is:
@rschristian  Can I use this right now in some way?

rschristian:
@kay-is https://github.com/preactjs/preact-cli/issues/1579#issuecomment-1362313817

PR #1648, opened 2022-02-01
Chore: Backport recent refactorings

rschristian:

Minor refactors

Existing cover

**Summary**

Backport of misc changes made in #1466/#1608 and #1645. I don't want those PRs to get completely out-of-hand and impossible to review, so I'm extracting out some of the more less-focused changes.

 - Added a `jsconfig.json` for better intellisense. Seems like a pretty handy addition, caught a bunch of minor issues already.
 - Re-enables the build output checking from #1466 
 - Removes non-functional missing node modules check (read below)
 - Public path is ensured to be normalized for service workers
   - `publicPath = http://some-url.com/foo` -> `http://some-url.com/foosw.js` without this, which is no good.

No


packages/cli/lib/lib/webpack/run-webpack.js
```diff
@@ -17,22 +17,6 @@ async function devBuild(env) {
...
-				Array.from(missingDeps).some(
-					file => file.indexOf(nodeModulesPath) !== -1
-				)
-			) {
-				compilation.contextDependencies.push(nodeModulesPath);
-			}
-
-			callback();
-		});
-
```
rschristian, line 35: This was adopted from CRA's `WatchMissingNodeModulesPlugin`, however, it seems to have been broken for a few years. This path that it watches resolves to `<project-name>/node_modules/preact-cli/node_modules` which is pretty pointless, as nothing will ever exist/be added to that.

Rather than correct it, I've elected to remove it after seeing [it was recently cut from CRA in v5](https://github.com/facebook/create-react-app/pull/11201). It apparently was having a rather large performance impact which they either couldn't solve or decided it wasn't worth solving. Not entirely sure which.

I figure there's probably no reason to doubt the performance issues claims, so I cut it.

ForsakenHarmony (approved):

PR #1658, opened 2022-02-05
Chore/docs

rschristian:
Just minor revisions to docs, in ReadMe and CLI


packages/cli/lib/commands/build.js
```diff
@@ -42,7 +42,7 @@ const options = [
 	},
 	{
 		name: '--template',
-		description: 'Path to custom HTML template',
+		description: 'Path to custom HTML template (default "src/template.html")',
```
rschristian, line 45: We don't set this as the default value in `sade`, only in the message, to keep this detection intact:

https://github.com/preactjs/preact-cli/blob/84dba1a78d98dc10f8ce7d2114726f62b0fba2f2/packages/cli/lib/lib/webpack/render-html-plugin.js#L21-L33

ForsakenHarmony (approved):

PR #1667, opened 2022-02-25
Fix/prerender urls esm

### LINKED ISSUE #1662: Unexpected token export when using prerenderUrls
Hello, today I tried using the `--prerenderUrls` tag with a `.js` file in a TypeScript project.. I need some dynamic props in my `prerender-urls` file, which means I also imported some other files in it, and then using `export default` of the function with the array of props that I need, but I get `Unexpected token export`.. When I try using `module.exports = ...` I then again have issues about my imports since they're `.ts` files.. Is there any way to accomplish what I am doing?

rschristian:

Bugfix

Yes

**Summary**

Closes #1662 

As the linked issue mentions, we `require()` the user's `prerender-urls.js` file which disallows `import/export` keywords. I'm guessing this used to work due to ESM not really being solidified in Node? Not entirely sure.

Fixed this by mirroring the process we already use for the user's `preact.config.js` files.

No


packages/cli/package.json
```diff
@@ -34,9 +34,9 @@
   "homepage": "https://github.com/preactjs/preact-cli",
   "devDependencies": {
     "@types/express": "^4.17.13",
-    "@types/jest": "^27.4.0",
+    "@types/jest": "^24.9.1",
     "html-looks-like": "^1.0.2",
-    "jest": "^27.0.1",
+    "jest": "^24.9.0",
```
rschristian, line 39: Chased down the [weird config file hack](https://github.com/preactjs/preact-cli/blob/aea6fedbc0fa363f94296c5c3912c349f5312527/packages/cli/lib/lib/webpack/transform-config.js#L110-L116) I introduced in #1623, turns out to be an issue in `esm` (the package, not the format) in Jest > v25.

Issue filed at https://github.com/facebook/jest/issues/12493, but I needed to use ESM within Jest and that hack only worked for CJS config files in our test suite. Downgrading shouldn't be problematic, so I went this route.

packages/cli/lib/lib/webpack/render-html-plugin.js
```diff
@@ -109,19 +108,18 @@ module.exports = async function (config) {
...
 				if (typeof result === 'function') {
-					info(`Fetching URLs from ${config.prerenderUrls}`);
 					result = await result();
-					info(`Fetched URLs from ${config.prerenderUrls}`);
 				}
 				if (typeof result === 'string') {
 					result = JSON.parse(result);
 				}
-				if (result instanceof Array) {
+				if (Array.isArray(result)) {
```
rschristian, line 122: Not sure why this stopped working, but probably shouldn't be using `instanceof` anyhow

packages/cli/lib/lib/webpack/render-html-plugin.js
```diff
@@ -109,19 +108,18 @@ module.exports = async function (config) {
...
+				let result = esmImport(resolve(cwd, config.prerenderUrls));
+
 				if (typeof result.default !== 'undefined') {
-					result = result.default();
+					result = result.default;
 				}
 				if (typeof result === 'function') {
-					info(`Fetching URLs from ${config.prerenderUrls}`);
 					result = await result();
-					info(`Fetched URLs from ${config.prerenderUrls}`);
```
rschristian, line 119: I don't think this info messages really provided any useful information.

rschristian (commented):
One thing that might also be worth bringing up: I noticed that if a user provides a `prerender-urls` file, and it cannot be found or is malformed, we give a warning and move on with the default values. If the same happens for the `preact.config`, we throw an error and exit immediately.

Do we want to change this behavior so they match? I can't say I really have a preference for which way we go.

ForsakenHarmony (approved):

rschristian:
I was looking to add similar file handling for the babelConfig in v4 and the difference in preact config vs prerender urls was bothering me so I fixed it up here by extracting out a reusable handler for such situations.

No longer will a missing `preact.config.js` file result in a broken build -- instead, it will continue on using the default (but printing a warning).

rschristian:
I reverted the test suite changes as I'm not happy with them and am planning to do a larger PR for tests in the next day or so.

PR #1671, opened 2022-03-14
Feat: Add 'PREACT_APP_' prefixed env vars automatically & pickup .env file

rschristian:

feature

Yes

**Summary**

This is something that's asked about fairly often and I figure it should become better supported for how low maintenance it is for us.

Any env vars prefixed with `PREACT_APP_` will now automatically be added to `DefinePlugin`. This matches what [CRA does](https://github.com/facebook/create-react-app/blob/bb64e31a81eb12d688c14713dce812143688750a/packages/react-scripts/config/env.js#L67-L69) and also (mostly) matches what [`preact-cli-plugin-env-vars` does](https://github.com/robinvdvleuten/preact-cli-plugin-env-vars/blob/a8436c8a3ed2da5e924af2dbf89330b906a35da1/index.js#L7-L8). This is still a fairly popular plugin, getting ~1/10 of the downloads per week that preact-cli itself is.

`.env` files in the user's project root are also now automatically consumed. Added `dotenv` for handling this.

Shouldn't be, no.

### DIFF
--- a/packages/cli/lib/commands/watch.js
+++ b/packages/cli/lib/commands/watch.js
@@ -2,6 +2,7 @@ const runWebpack = require('../lib/webpack/run-webpack');
 const { isPortFree, toBool, warn } = require('../util');
 const { validateArgs } = require('./validate-args');
 const getPort = require('get-port');
+const { resolve } = require('path');
 
 const options = [
 	{
@@ -104,6 +105,9 @@ async function command(src, argv) {
 	if (argv.sw) {
 		argv.sw = toBool(argv.sw);
 	}
+
+	let cwd = resolve(argv.cwd);
+	require('dotenv').config({ path: resolve(cwd, '.env') });
 
 	argv.port = await determinePort(argv.port);
 
--- a/.changeset/stupid-suns-crash.md
+++ b/.changeset/stupid-suns-crash.md
@@ -1,0 +1,7 @@
+---
+'preact-cli': minor
+---
+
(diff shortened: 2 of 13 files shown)


rschristian:
One thing that came up is the question of whether the `.env` file path should be a flag or not. Can't say I have much of an opinion either way.

packages/cli/lib/commands/watch.js
```diff
@@ -105,6 +106,9 @@ async function command(src, argv) {
 		argv.sw = toBool(argv.sw);
 	}
 
+	let cwd = resolve(argv.cwd);
+	require('dotenv').config({ path: resolve(cwd, '.env') });
```
ForsakenHarmony, line 110:
```suggestion
	require('dotenv').config();
```

Wouldn't this work as well?
https://github.com/motdotla/dotenv#path

ForsakenHarmony (approved):

ForsakenHarmony:
> One thing that came up is the question of whether the `.env` file path should be a flag or not.

Could always add that later on

packages/cli/lib/commands/watch.js
rschristian: In normal circumstances, sure.

However, that would break in our test suite (and probably mono-repos too): `cwd` would differ from the project root.

I ran into that issues and moved the dotenv setup here accordingly.

rschristian:
Hm something in that watch test is badly misbehaving and getting the CI stuck in a loop. Will need to look into that tomorrow, though nothing stands out at the moment.

PR #1674, opened 2022-03-15
feat: support CRA-style proxy config

### LINKED ISSUE #1672: Feature: Support supplying proxy info from `package.json`
Setting up a proxy needs to be done in the `preact.config.js`

**What is the motivation or use case for changing this behaviour?**
CRA users can get tripped up on this 

temp

**Describe the solution you'd like**
Should be simple for us to support consuming proxy info in the same way. Would allow for easier migration for CRA users.

rschristian:

feature

Yes

**Summary**

Closes #1672

Some users copy/paste from CRA thinking their config will work out of the box. While that won't work everywhere, we can support simple cases like this pretty easily.

This implementation is pretty close to [CRA's](https://github.com/facebook/create-react-app/blob/fd8c5f7b1b1d19d10d24cc2f9fdfc110585dc030/packages/react-dev-utils/WebpackDevServerUtils.js#L302-L389), though with a couple changes around the public folder and error handling.

I'm leaning towards not documenting this though, would rather people use their `preact.config.js` to better control this themselves. I think the defaults I've set are probably fine, but still not a "one size fits all" situation.

Technically, it could be problematic if a user has a (previously) extraneous key sitting around in their `package.json`. Don't know if that counts and might be a bit of a reach.

### DIFF
--- a/packages/cli/lib/lib/webpack/webpack-client-config.js
+++ b/packages/cli/lib/lib/webpack/webpack-client-config.js
@@ -278,6 +278,43 @@ function isProd(env) {
 	return prodConfig;
 }
 
+function setupProxy(target) {
+	if (!target) {
+		return;
+	}
+
+	const errorTemplate = warn =>
+		`Invalid proxy configuration in package.json. ${warn} Skipping and using default.`;
+
+	if (typeof target !== 'string') {
+		warn(errorTemplate('If provided, "proxy" must be a string.'));
+		return;
+	} else if (!/https?:\/\//.test(target)) {
+		warn(
+			errorTemplate(
+				'If provided, "proxy" must start with "http://" or "https://".'
+			)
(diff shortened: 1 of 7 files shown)


packages/cli/lib/lib/webpack/webpack-client-config.js
```diff
@@ -278,6 +278,43 @@ function isProd(env) {
...
+			errorTemplate(
+				'If provided, "proxy" must start with "http://" or "https://".'
+			)
+		);
+		return;
+	}
+
+	return {
+		target,
+		logLevel: 'warn',
```
rschristian, line 303: This will result in proxy issues appearing in the user's terminal like the following:

temp

I tend to think that's handy, but we certainly could disable it and leave those errors to the browser console too.

PR #1680, opened 2022-03-23
fix: push manifest

### LINKED ISSUE #1675: Push manifest does not adhere custom changes in preact config file

When I change output paths via `preact.config.js` file, generated `push-manifest.json` file is wrong.

**Steps to reproduce:**

1. Change output paths via `preact.config.js` file:
```javascript

	config.output.filename = "scripts/" + config.output.filename;
	config.output.chunkFilename = "scripts/" + config.output.chunkFilename;

	const { plugin: babelEsmPlugin } = helpers.getPluginsByName(config, "BabelEsmPlugin")[0];
	babelEsmPlugin.options_.filename = "scripts/" + babelEsmPlugin.options_.filename;
	babelEsmPlugin.options_.chunkFilename = "scripts/" + babelEsmPlugin.options_.chunkFilename;

	const { plugin: miniCssExtractPlugin } = helpers.getPluginsByName(config, "MiniCssExtractPlugin")[0];
	miniCssExtractPlugin.options.filename = "styles/" + miniCssExtractPlugin.options.filename;
	miniCssExtractPlugin.options.chunkFilename = "styles/" + miniCssExtractPlugin.options.chunkFilename;

};
```

2. Build:
```bash
$ npm run build
```

3. Check generated `push-manifest.json` file:
```bash
$ cat build/push-manifest.json
```
```json
{"/":{"undefined":{"type":"script","weight":1}}}
```

Generate correct `push-manifest.json` file.

**Please mention any other relevant information:**

https://github.com/preactjs/preact-cli/blob/aea6fedbc0fa363f94296c5c3912c349f5312527/packages/cli/lib/lib/webpack/create-load-manifest.js#L9

In `filename` variable are not basenames of assets, but whole path, therefore regular expressions won't match.

```
scripts/bundle.945f9.js
scripts/bundle.d7579.esm.js
scripts/polyfills.60e68.js
scripts/polyfills.aae10.esm.js
styles/bundle.fa67f.css
sw-esm.js
sw.js
```

https://github.com/preactjs/preact-cli/blob/aea6fedbc0fa363f94296c5c3912c349f5312527/packages/cli/lib/lib/webpack/create-load-manifest.js#L16

https://github.com/preactjs/preact-cli/blob/aea6fedbc0fa363f94296c5c3912c349f5312527/packages/cli/lib/lib/webpack/create-load-manifest.js#L18

Please paste the results of `npx preact-cli info` here.

```
Environment Info:
  System:
    OS: Linux 4.15 Ubuntu 16.04.7 LTS (Xenial Xerus)
    CPU: (4) x64 Intel(R) Core(TM) i5-5200U CPU @ 2.20GHz
  Binaries:
    Node: 17.6.0 - /usr/bin/node
    Yarn: 1.22.5 - /usr/bin/yarn
    npm: 8.5.1 - /usr/bin/npm
  Browsers:
    Chrome: 99.0.4844.51
    Firefox: 88.0
  npmPackages:
    preact: ^10.3.1 => 10.6.6
    preact-cli: ^3.0.0 => 3.3.5
    preact-render-to-string: ^5.1.20 => 5.1.20
```

rschristian:

Bug fix / refactoring for future use

Yes

**Summary**

Closes #1675

In short, the `push-manifest.json` output had hard-coded expectations that caused two problems in the linked issue:
  - If the asset could not be found, there would be an `"undefined"` entry in the `push-manifest`
  - If the asset output paths were altered, they could not be added to the manifest due to the strict filename matching

This fixes both issues by using `webpack-manifest-plugin` (Webpack really should make this part of core, extremely useful to expose this info to other plugins).

This plugin provides a map output assets that allows us to keep our expectation about a `bundle.css` or `route-home.js` existing, but allows users to still customize the actual output. For example, the following config:

```js
// preact.config.js
	config.output.publicPath = '/foo/';

	const { plugin: miniCssExtractPlugin } = helpers.getPluginsByName(config, "MiniCssExtractPlugin")[0];
	miniCssExtractPlugin.options.filename = "styles/foobar.css";
};
```

...would create the following manifest:

```json
{
  "/bundle.css": "/foo/styles/foobar.css",
  "/bundle.js": "/foo/bundle.29bec.esm.js",
  "/bundle.js.map": "/foo/bundle.29bec.esm.js.map",
}
```

Should allow users to customize output names as they wish without it interfering with any assumptions we've made about output.

In the future we might be able to make greater use of this and move away from asset path concatenation like the following:

https://github.com/preactjs/preact-cli/blob/a56d904ee0a213109a0e6090ab4e8e575b382001/packages/cli/lib/resources/head-end.ejs#L1

However, for now, I'd hesitate to do that as this might break some configs. Better to wait for a major.

Still, it should allow for much more robust targeting of output assets as it won't fall to pieces if a user makes configuration changes.

No


packages/cli/lib/lib/webpack/create-load-manifest.js
```diff
@@ -1,74 +1,67 @@
-const isESM = filename => /\.esm\.js$/.test(filename);
-const isMatch = (filename, condition) => isESM(filename) === condition;
```
rschristian, line 2: In case this raises an eyebrow, the asset manifest will always map to ESM output if enabled, CJS if it's not. We don't need to deal with that ourselves anymore.

There's a new test to ensure the `push-manifest.json` is correct and referring to ESM scripts when ESM is enabled.

PR #1690, opened 2022-05-05
fix: silently failing test in watch.test.js

zubhav:
Fixes a silently failing test in `watch.test.js`

The fix was a change to a test itself

**Summary**
The test in concern was using `.catch()` to assert that an error was thrown. However, if the logic of the function was changed and a Promise did not reject, this test would pass as it wouldn't invoke `.catch()`. 

By using Jest's `.reject`, we are able to assert that the Promise rejected as well as checking the error itself.

No

### DIFF
--- a/packages/cli/tests/watch.test.js
+++ b/packages/cli/tests/watch.test.js
@@ -89,7 +89,7 @@ describe('should determine the correct port', () => {
 		expect(await determinePort('3999')).toBe(3999);
 	});
 
-	it('should use $PORT in the abscence of --port', async () => {
+	it('should use $PORT in the absence of --port', async () => {
 		process.env.PORT = '4001';
 		expect(await determinePort()).toBe(4001);
 	});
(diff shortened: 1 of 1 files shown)


rschristian (approved):
Cheers, good spot.

Thanks for fixing my awful spelling too

zubhav:
👍  Noticed there were some lint warnings in the demo app section. Is it worth me raising a PR to add eslint-ignore comment for these lines?

Screenshot 2022-05-05 at 09 45 29

rschristian:
If you're willing, sure. Definitely happy to merge improvements.

Could also just convert that to a variable assignment instead (and update the test).

zubhav:
@rschristian Cool, I'll prob raise a PR later for those warnings. 

Had no issue with port assignment. Cloned this repo to browse around and look for anywhere I could help. Let me know if you need a hand on anything else, I'll do my best to help :)

PR #1691, opened 2022-05-09
fix: remove lint warnings, update tests and rename file

zubhav:
Tidy-up/fixes

**Summary**
- Removes ESLint warnings that appear in CI
- Rename consants to constants in filename
- Remove unused file (TBC)

166890080-38cc2aa7-b70e-4897-ba87-ceab322b4fb2


packages/cli/babel/index.js
```diff
@@ -1,3 +1,5 @@
+// Is this file used/required? If not, can it be removed?
```
zubhav, line 1: Probably a question for @rschristian 

I couldn't find any instance of this file being used. Unless it's used implicitly by Babel, could we remove it entirely?

rschristian: Not used by us, no. However, users can add it as a preset to their own babel config if they so wish. 

It's not a great setup tbh, so I'm going to try to remove it in the next major, but there are some users who depend on it for now.

packages/cli/tests/build.test.js
```diff
@@ -182,8 +182,10 @@ describe('preact build', () => {
 			`${dir}/build/${bundleFile}`,
 			'utf8'
 		);
+
 		// "Hello World!" should replace 'process.env.PREACT_APP_MY_VARIABLE'
-		expect(transpiledChunk.includes('console.log("Hello World!")')).toBe(true);
+		expect(transpiledChunk.includes('Hello World!')).toBe(true);
```
rschristian, line 187: nit: We do need to check for quotes when replacing env vars, else if we fail to stringify we'd end up with `var foo = Hello World!`

```suggestion
		expect(transpiledChunk.includes('"Hello World!"')).toBe(true);
```

packages/cli/babel/index.js
```diff
@@ -1,3 +1,5 @@
...
+
```
rschristian, line 2:
```suggestion
```

rschristian (approved):
Thanks for doing this!

PR #1699, opened 2022-06-07
refactor: Simplify CLI flag validation

rschristian:

Refactor

Adjusted one, yes.

Larger test changes were extracted to #1738 because this diff was rather large and I wanted to use base other work off of both of these independently. 

**Summary**

Partial revert of #1467

`sade` has a [builtin mechanism for catching unknown options being passed to it](https://github.com/lukeed/sade#optsunknown), allowing us to simplify things a bit.

There should be no functional changes here at all.

No


packages/cli/package.json
```diff
@@ -45,7 +45,7 @@
     "preact": "^10.5.13",
     "preact-render-to-string": "^5.1.19",
     "preact-router": "^3.0.1",
-    "puppeteer": "^9.1.1",
+    "puppeteer": "^14.4.0",
```
rschristian, line 48: Older version had some issues with SWs and upgrading also allows for top-level await in the evals which is convenient.

packages/cli/lib/lib/webpack/webpack-client-config.js
```diff
@@ -327,6 +316,7 @@ function isDev(env) {
 		},
 		plugins: [
 			new webpack.NamedModulesPlugin(),
+			env.esm && new webpack.HotModuleReplacementPlugin(),
```
rschristian, line 319: This is a weird one that I don't fully understand. `webpack-dev-server` v4 will immediately throw with the following error if the `--esm` flag is used:

```
Uncaught Error: [HMR] Hot Module Replacement is disabled.
```

Just points to Webpack itself throwing, so no useful stack trace there. You can see this on the latest release of `preact-cli`.

Users will get warnings that this plugin is automatically added for them if the devServer has `hot` enabled ([as we do](https://github.com/preactjs/preact-cli/blob/a60f8df8e87935f3b3e03a2bb4aecda801568c83/packages/cli/lib/lib/webpack/webpack-client-config.js#L333-L334)), though for whatever reason this does fix the ESM issue. To reduce noise I've therefore put this behind the esm flag check.

Only found it from looking at Prefresh's docs to see if we were missing something. Not sure why it works, but I imagine it's something with `babel-plugin-esm`.

packages/cli/lib/index.js
```diff
@@ -28,41 +27,100 @@ process.on('unhandledRejection', err => {
...
 	.command('build [src]')
 	.describe(
 		'Create a production build. You can disable "default: true" flags by prefixing them with --no-<option>'
-	);
-commands.buildOptions.forEach(option => {
-	buildCommand.option(option.name, option.description, option.default);
-});
-buildCommand.action(commands.build);
+	)
+	.option('--src', 'Specify source directory', 'src')
```
ForsakenHarmony, line 37: I think I remember it originally being done this way, but I don't know why it was changed to the version it currently is

Probably older than the new `sade` option

ForsakenHarmony (approved):

packages/cli/lib/index.js
rschristian: Indeed, we had the commands in this format but I think `sade`'s option was just missed, so validation was done the way it was. It's a pretty long ReadMe without a table of contents, easy enough to do.

Alternatively, maybe there was an idea for better type checking on args down the road? `sade`/`mri` are a bit "loose" with coercing input types, leading to a few odd situations. Just guessing though; if we wanted to do that there's probably simpler ways to validate the schema than this setup.

PR #1700, opened 2022-06-08
chore: Removes 'material' template from 'list' cmd output

rschristian:
*What kind of change does this PR introduce?**

Chore

N/A

**Summary**

Removes the 'material' template from output of `preact list`. This template is archived and we've already removed it from the docs. 

No. Users can still select it if they want, we just won't display/advertise it.

### DIFF
--- a/packages/cli/lib/commands/list.js
+++ b/packages/cli/lib/commands/list.js
@@ -11,13 +11,14 @@ module.exports = async function () {
 		process.stdout.write('\n');
 		info('Available official templates: \n');
 
-		repos.forEach(repo => {
-			if (repo.name === 'material') return;
-			const description = repo.description ? ` - ${repo.description}` : '';
-			process.stdout.write(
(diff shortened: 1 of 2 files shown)


packages/cli/lib/commands/list.js
```diff
@@ -11,7 +11,8 @@ module.exports = async function () {
 		process.stdout.write('\n');
 		info('Available official templates: \n');
 
-		repos.map(repo => {
+		repos.forEach(repo => {
```
JoviDeCroock, line 14: Can we do `repos.filter(repo => !repo.archived)` instead to have this scale with us archiving more of them?

rschristian: Whoops, can indeed! Forgot that wasn't just a list of repo names.

Thanks

JoviDeCroock (approved):

PR #1707, opened 2022-07-07
Update critters-webpack-plugin to 3.x to address CVE-2021-20066

nocive:

Updates `critters-webpack-plugin` dependency to 3.x.

No.

**Summary**

The critters-webpack-plugin depends on jsdom which has a known CVE vulnerability in versions <16.5.0.
The latest version of critters-webpack-plugin (3.x) no longer depends on jsdom and therefore solves this issue.

See https://github.com/advisories/GHSA-f4c9-cqv8-9v98 for more information on the CVE.

Looking at the critters-webpack-plugin changelog, none that I'm aware of.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -94,7 +94,7 @@
     "compression-webpack-plugin": "^6.1.1",
     "console-clear": "^1.0.0",
     "copy-webpack-plugin": "^6.4.0",
-    "critters-webpack-plugin": "^2.5.0",
+    "critters-webpack-plugin": "^3.0.0",
     "cross-spawn-promise": "^0.10.1",
     "css-loader": "^5.2.7",
     "dotenv": "^16.0.0",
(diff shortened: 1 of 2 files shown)


nocive:
@preact/cli can someone please approve the workflow so that the tests can run? :pray:

rschristian:
Sorry, can't do.

`critters` has a peer dependency on `html-webpack-plugin`, and upgrading from critters v2 -> v3 bumps that peer dep requirement from v3 -> v4.

Bumping `html-webpack-plugin` requires another major due to it being a part of the public API.

#1608 already handles this, though it can't be released into `preact-cli` 3.x

nocive:
@rschristian thanks for the context, wasn't aware of any of this.

Just for my own reference, so you're saying we can only count on this being patched in a future preact-cli 4.x version?

rschristian:
Just to clarify a bit, the "malicious web app" from that text would have to be your own. So the vulnerability here is that you could write **your app** to maliciously manipulate **your files** at **build time only**.

I have many problems with the current state of JS security warnings, and this is why.

Issue #1706, opened 2022-07-07
Environment variables doesn't seem to work.


drager:
I have a `.env` file in my root like so:

```
PREACT_APP_API_URL="https://api.example.com/"
```

Without any config I get `process is not defined` and then I did this to my `preact.config.js`:

```js
import dotenv from "dotenv";

dotenv.config();

export default (config, env, helpers) => {
  config.devServer.liveReload = false;
  config.node.process = true;

  const { plugin } = helpers.getPluginsByName(config, "DefinePlugin")[0];
  plugin.definitions["process.env.MY_VARIABLE"] = JSON.stringify("my-value");
};
```

and `console.log(process.env)` in `index.tsx` and it's an empty object.

I'm using `preact-cli@3.4.0`

rschristian:
Can you show you're using your value? We have a couple tests that confirm it's acting as it should. You can see the test subject here: https://github.com/preactjs/preact-cli/tree/master/packages/cli/tests/subjects/custom-dotenv

> and `console.log(process.env)` in `index.tsx` and it's an empty object.

This is correct. `process` doesn't exist outside of Node, so what `DefinePlugin` does is simple find & replace, i.e., `console.log(process.env.MY_VARIABLE)` -> `console.log('foo')`. This doesn't make `process` usable, just literally swaps out any usage. This is why `process.env` is empty, and that's correct.

drager:
> Can you show how you're using your value? We have a couple tests that confirm it's acting as it should. You can see the test subject here: https://github.com/preactjs/preact-cli/tree/master/packages/cli/tests/subjects/custom-dotenv
> 
> You should being doing something like:
> 
> ```js
> console.log(process.env.PREACT_APP_API_URL);
> ```
> 
> > and `console.log(process.env)` in `index.tsx` and it's an empty object.
> 
> This is correct, and maybe where understanding what `DefinePlugin` actually does becomes important.
> 
> `DefinePlugin` acts as a find & replace, it doesn't make `process` usable in browser bundles. It scans your code, and upon seeing any strings you've defined (`process.env.MY_VARIABLE`), it replaces it for the value you've provided for that definition (`'my-value'`). This really is just basic string matching, it's not accessing `MY_VARIABLE` upon `process.env`.
> 
> This is why `process.env` is empty, which is correct behavior.
> 

Ah alright! `console.log(process.env.PREACT_APP_API_URL);` works. I thought logging `process.env` would give me something. Then this is a non-issue 😊 

[...]

PR #1709, opened 2022-07-08
docs: Expounding upon how DefinePlugin actually works

rschristian:
https://github.com/preactjs/preact-cli/issues/1706

Hopefully will catch users who try to log/use `process.env` expecting their environment variables to show up.


rschristian:
Think I close this on accident.

developit (approved):

PR #1714, opened 2022-07-24
refactor: Determine CSS module by filename, not directory

rschristian:

Refactor

Edited existing

**Summary**

Changes CSS Module handling to be based upon filename, rather than directory.

The existing setup is rather unintuitive, as identical files are handled very differently depending upon where they are, and require different user code in order to take advantage of. Using file names to distinguish is far more transparent to the end user.

Yes

Users using CSS Modules in the `routes` or `components` directory will need to update their file names to match the pattern `*.module.css`.


packages/cli/lib/lib/webpack/run-webpack.js
```diff
@@ -143,7 +143,7 @@ function writeJsonStats(cwd, stats) {
 function allFields(stats, field, fields = [], name = null) {
 	const info = stats.toJson({
 		errors: true,
-		warnings: false,
+		warnings: true,
```
rschristian, line 146: Noticed when debugging.

We attempt to log warnings if `stats.hasWarnings()` is true:

https://github.com/preactjs/preact-cli/blob/3f389c2486a3790ae9e18b67ac959f87de19c973/packages/cli/lib/lib/webpack/run-webpack.js#L167-L169

However, `info.warnings` will always be empty w/out this change.

packages/cli/lib/lib/webpack/webpack-base-config.js
```diff
@@ -63,7 +63,7 @@ function getSassConfiguration(...includePaths) {
  * @returns {import('webpack').Configuration}
  */
 module.exports = function createBaseConfig(env) {
-	const { cwd, isProd, isWatch, src, source } = env;
+	const { cwd, isProd, src, source } = env;
```
rschristian, line 66: Just a consistency thing. Easier to scan if we're not switching between `isProd` and `isWatch`.

PR #1716, opened 2022-07-25
Refactor/remove preact 8 support

rschristian:

Refactor

Removed a test, actually

**Summary**

Drops support for Preact v8.

Yes, users need to upgrade to v10+


packages/async-loader/index.js
```diff
@@ -1,11 +1,9 @@
 const path = require('path');
-const { getOptions, stringifyRequest } = require('loader-utils');
```
rschristian, line 2: Should've been included in the Webpack v5 PR, but I evidently missed it.

PR #1722, opened 2022-08-01
chore: repository clean

rschristian:

chores

Existing cover

**Summary**

Just cleaning up some things that have been on my mind

- Removed lerna
  - AFAIK we aren't really using it. Yarn workspaces cover our use pretty well, and Lerna is quite heavy; it made up ~1/8th of the lock file.
- Formatted `package.json` files
  - Consistency-wise, they were all over the place. Different ordering of values, different entries, etc. [I ran a formatter](https://www.npmjs.com/package/sort-package-json) over them to at least get consistent ordering, then manually tried to make sure values like `author`, `repository`, etc. were corrected and matched.
- Moved `cli/lib` -> `cli/src`, again for consistency (and because `cli/lib/lib/...` was never great).
  - Would be nice for `async-loader` to match but I think that would require properly setting up package exports to _maybe_ avoid breaking changes. Likely would still break something, so I'm avoiding that for now.


packages/prerender-data-provider/package.json
```diff
@@ -1,29 +1,24 @@
...
+  "description": "Prerender data provider for preact-cli apps",
   "keywords": [
     "preact",
     "preact-cli"
   ],
-  "author": "Preact Authors <core@preactjs.com>",
-  "homepage": "https://github.com/developit/preact-cli#readme",
-  "license": "MIT",
-  "main": "src/index.js",
-  "directories": {
```
rschristian, line 13: It doesn't seem like this serves much purpose?

https://docs.npmjs.com/cli/v8/configuring-npm/package-json#directories

packages/async-loader/package.json
```diff
@@ -2,26 +2,26 @@
...
+    "directory": "packages/async-loader"
   },
-  "author": "Jason Miller <jason@developit.ca> (http://jasonformat.com)",
   "license": "MIT",
-  "bugs": "https://github.com/preactjs/preact-cli/issues",
-  "homepage": "https://github.com/preactjs/preact-cli",
-  "engines": {
-    "node": ">=8"
-  },
+  "author": "The Preact Authors (https://preactjs.com)",
```
rschristian, line 13: We have quite a few variations of this across our repos. I copied from one, but open to suggestions/recommendations!

ForsakenHarmony, line 13:
```suggestion
  "author": "The Preact CLI Authors (https://github.com/preactjs/preact-cli/contributors)",
```

Maybe like in the preact core repo https://github.com/preactjs/preact/blob/master/package.json#L247

packages/prerender-data-provider/package.json
```diff
@@ -1,29 +1,24 @@
...
-    "url": "git+https://github.com/developit/preact-cli.git"
-  },
-  "bugs": {
-    "url": "https://github.com/developit/preact-cli/issues"
+    "url": "https://github.com/preactjs/preact-cli.git",
+    "directory": "packages/prerender-data-provider"
   },
+  "license": "MIT",
+  "author": "The Preact Authors (https://preactjs.com)",
+  "main": "lib/index.js",
```
ForsakenHarmony, line 18: I personally really dislike `lib` in the root dir, would kinda prefer moving the CLI code to src

ForsakenHarmony (approved):

PR #1731, opened 2022-08-13
experimental: Using preact-iso to lazy load components

rschristian:

Experimental refactor

Yes

**Summary**

Maybe deprecate the async loader for instead using `preact-iso`'s `lazy()`.

Code splitting, as it works now, is rather opaque as it depends on magic directories rather than user code.

Yes


packages/cli/lib/lib/webpack/webpack-base-config.js
```diff
@@ -292,9 +292,12 @@ module.exports = function createBaseConfig(env) {
...
 					isProd && !env.isServer ? '[name].[contenthash:5].css' : '[name].css',
-				chunkFilename: isProd
-					? '[name].chunk.[contenthash:5].css'
-					: '[name].chunk.css',
+				chunkFilename: pathData => {
+					const chunkName = pathData.chunk.id.split('_').slice(0, -1).join('-');
+					return isProd
+						? `${chunkName}.chunk.[chunkhash:5].css`
+						: `${chunkName}.chunk.css`;
+				},
```
rschristian, line 300: TODO: Maybe try to make these chunk names better match previous setup

packages/cli/lib/lib/webpack/preact-iso-prerender.js
```diff
@@ -0,0 +1,60 @@
+const { h, options, cloneElement } = require('preact');
```
rschristian, line 1: Vendoring `preact-iso`'s `prerender.js`: https://github.com/preactjs/wmr/blob/main/packages/preact-iso/prerender.js

`preact-iso` only offers ESM output and [dynamic imports cause v8's `vm` subsystem to segfault](https://github.com/nodejs/node/issues/35889). Probably need to have babel transpile `preact-iso` or something, at least for prerendering purposes.

packages/cli/lib/lib/webpack/prerender.js
```diff
@@ -2,17 +2,18 @@ const { red, yellow } = require('kleur');
...
-module.exports = function (env, params) {
+module.exports = async function (env, params) {
 	params = params || {};
 
 	let entry = resolve(env.dest, './ssr-build/ssr-bundle.js');
 	let url = params.url || '/';
 
 	global.history = {};
-	global.location = { ...URL.parse(url) };
+	global.location = new URL(url, 'http://localhost');
```
rschristian, line 16: Note: Moving to the WHATWG URL API is probably a breaking change. This does rather notably set things like `location.hostname` where previously they were null. Hopefully no one is relying on that to determine ssr vs browser envs.

rschristian:
Supporting automatic population of preload tags is gonna be pretty tricky now (at least, doing so in a flexible manner) as we no longer have magic directory names to match up against:

https://github.com/preactjs/preact-cli/blob/6aba5d8e4c165c81f258ca9f367fea6c9a08cf17/packages/cli/lib/lib/webpack/create-load-manifest.js#L34-L35

I suppose we do still set the chunk names, so we could input something there, but there's really no way for us to tell a lazily loaded chunk apart from lazily loaded chunks that are routes. Maybe we set `routes/` and say `pages/` as magic directory names for preload generation. If a chunk is prefixed with either, it & corresponding assets will go in the manifest to generate preload tags. Else, we'll just skip it.

Dunno how we'd differentiate otherwise, but maybe an ideal will come to me.

rschristian:
Just now realizing our method of connecting split chunks to routes is totally junk:

https://github.com/preactjs/preact-cli/blob/1d53f2659ed833bdfbb0a8cb3c8167925a7aa89c/packages/cli/lib/lib/webpack/create-load-manifest.js#L46

There's no guarantee at all that `route-foo` (which comes from the file structure `routes/foo`) corresponds to `/foo`. We're incorrectly assuming directory/file names directly correspond to the route paths they're served from which isn't remotely true.

Think we should rip our preloading out entirely and maybe switch to https://webpack.js.org/guides/code-splitting/#prefetchingpreloading-modules

packages/cli/tests/images/build.js
```diff
@@ -9,40 +9,40 @@ exports.default = {
...
 	'ssr-build/asset-manifest.json': 76,
 
-	'bundle.79d07.js': 21560,
-	'bundle.79d07.js.map': 85822,
-	'bundle.79d07.legacy.js': 22549,
-	'bundle.79d07.legacy.js.map': 106841,
+	'bundle.528e9.js': 26608,
+	'bundle.528e9.js.map': 107005,
+	'bundle.528e9.legacy.js': 28182,
+	'bundle.528e9.legacy.js.map': 133666,
```
rschristian, line 19: Note to self: this is a rather sizeable jump. Split chunks are no longer paying the cost of code splitting (as seen by `route-profile`'s change from 3198b -> 743b), but this is still rather extreme. ~5kb difference in main bundle for ~2.5kb in chunks doesn't seem right.

packages/cli/tests/images/build.js
```diff
@@ -9,40 +9,40 @@ exports.default = {
...
 	'bundle.354c3.css': 945,
 	'bundle.354c3.css.map': 1758,
 
-	'dom-polyfills.8a933.legacy.js': 5221,
-	'dom-polyfills.8a933.legacy.js.map': 18676,
-	'es-polyfills.legacy.js': 42690,
+	'dom-polyfills.8a933.legacy.js': 5252,
+	'dom-polyfills.8a933.legacy.js.map': 18836,
+	'es-polyfills.legacy.js': 61323,
```
rschristian, line 25: 20kb jump is also really heavy. Thinking something isn't right.

packages/cli/src/lib/webpack/preact-iso-prerender.js
```diff
@@ -0,0 +1,41 @@
+// Vendored CJS version of https://github.com/preactjs/wmr/blob/main/packages/preact-iso/prerender.js
+// The CJS Webpack SSR output combined with the ESM-only preact-iso resulted in dual-package hazard woes
```
rschristian, line 2: We should maybe add dual output to `preact-iso` but this'll work for now.

packages/cli/src/lib/webpack/webpack-base-config.js
```diff
@@ -45,19 +45,6 @@ function resolveTsconfig(cwd, isProd) {
...
-		sassOptions: {
-			includePaths,
-		},
-	};
-
-	Object.defineProperty(config, 'includePaths', { value: includePaths });
-
-	return config;
-}
-
```
rschristian, line 60: Old hack that's no longer needed

packages/cli/src/lib/webpack/prerender.js
```diff
@@ -11,25 +11,42 @@ module.exports = async function (config, params) {
 	let entry = resolve(config.dest, './ssr-build/ssr-bundle.js');
 	let url = params.url || '/';
 
-	global.history = {};
```
A170885, line 14: global.history = /** @type {object} */ ({});

rschristian: ?

packages/cli/src/lib/webpack/webpack-client-config.js
```diff
@@ -196,7 +154,7 @@ function prodBuild(config) {
 				preload: 'media',
 				pruneSource: false,
 				logLevel: 'silent',
-				additionalStylesheets: ['route-*.css'],
+				additionalStylesheets: ['routes-*.css'],
```
rschristian, line 157: Note: Probably not sound, only used to avoid duping issue w/ Critters (https://github.com/GoogleChromeLabs/critters/issues/88)

Probably need to figure something better out here

PR #1735, opened 2022-08-13
types: Restricts `preact.config.js`'s `env` types to a more idiomatic subset

rschristian:

refactor of `preact.config.js` types

N/A

**Summary**

Goes along with #1734 

`env` in `preact.config.js` is quite overloaded and has numerous semi-duplicated values: `production` and `isProd`, `ssr` and `isServer`, etc. I'd like to cut this down to just `isProd`, `isDev`, and `isServer` in v4, but for now, a good start would be to restrict the types & update the wiki.

At best, the types/values beyond these three (and their duplicates) are just convenience (not having to get the path for `src` yourself), and at worst, confuse new users.

No


developit:
Is `isServer` the one for prerendering? I regret calling it "SSR" here haha

developit:
Yeah, SSR is a bit of a misnomer there. Unless folks are still abusing it for SSR? haha

rschristian:
Some still are or at least want to, but I've been dissuading.

I'm somewhat still partial to `isServer` as it's not really meant as SSR, but as the server _Webpack build_. Not that there's much differentiating in the way we use it, but one could theoretically remove the prerendering and still have a server build for some reason.

That might be me doing mental gymnastics though.

ForsakenHarmony (approved):

PR #1752, opened 2022-12-13
refactor: prune flags

rschristian:

Refactor

Modified existing

**Summary**

##### Removed
- `--json`
  - While we don't do too much to support this, allowing a user to use alternative bundle analyzers than the one included I think is rather niche and unnecessary. I'm not set on removing it, but I don't think it sees much use in all honesty.
- `--brotli`
  - This appears to have been broken for a while with no complaints. I think it's pretty niche and better left up to users to figure out if they need it.
- `--rhl`
  - Was an undocumented alias for `--refresh`.   

##### Altered
- `--inline-css`
  - All other flags are camelCased, this is the odd one out. Switched it in the docs & CLI, but I did add an alias so `--inline-css` will continue to work without issues. No breaking changes here.  

Yes, will require stripping out the removed flags if in use

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,13 +123,11 @@ $ [npm run / yarn] preact build
     --cwd              A directory to use instead of $PWD  (default .)
     --sw               Generate and attach a Service Worker  (default true)
     --babelConfig      Path to custom Babel config (default .babelrc)
-    --json             Generate build stats for bundle analysis
-    --template         Path to custom HTML template (default 'src/template.html')
-    --analyze          Launch interactive Analyzer to inspect production bundle(s)
     --prerender        Renders route(s) into generated static HTML  (default true)
     --prerenderUrls    Path to pre-rendered routes config  (default prerender-urls.json)
-    --brotli           Adds brotli redirects to the service worker  (default false)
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -123,13 +123,11 @@ $ [npm run / yarn] preact build
...
-    --json             Generate build stats for bundle analysis
-    --template         Path to custom HTML template (default 'src/template.html')
-    --analyze          Launch interactive Analyzer to inspect production bundle(s)
     --prerender        Renders route(s) into generated static HTML  (default true)
     --prerenderUrls    Path to pre-rendered routes config  (default prerender-urls.json)
-    --brotli           Adds brotli redirects to the service worker  (default false)
-    --inline-css       Adds critical css to the prerendered markup  (default true)
+    --template         Path to custom HTML template (default 'src/template.html')
+    --inlineCss        Adds critical css to the prerendered markup  (default true)
+    --analyze          Launch interactive Analyzer to inspect production bundle(s) (default false)
```
rschristian, line 130: `--template` and `--analyze` were just moved to match the order for `preact watch`'s flags -- ordering consistency is just a nice-to-have and I felt dev's order was better.

PR #1753, opened 2022-12-15
refactor: Bump Node version, remove `src` arg in build & watch, and misc housekeeping

rschristian:

Refactor + a few misc chores

Existing should cover

**Summary**

- Bumps minimum Node version to v14.14.0 (to support `recursive` arg in fs.rm)
- Drops `src` argument in build & watch commands
  - Both support `--src`, so this was duplicated and rather unnecessary (IMO)
- Empties output directory rather than deleting it
  - Nice for containerized envs where a stable reference to mounted directories is necessary
- Removes `update-notifier`
  - Not necessary for a non-globally installed tool really. It also has a [fair number of dependencies](http://npm.anvaka.com/#/view/2d/update-notifier) that we can cut which is nice.   

No

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -72,7 +72,6 @@
     "progress-bar-webpack-plugin": "^2.1.0",
     "promise-polyfill": "^8.2.3",
     "react-refresh": "0.11.0",
-    "rimraf": "^3.0.2",
     "sade": "^1.8.1",
     "size-plugin": "^2.0.2",
     "source-map": "^0.7.2",
@@ -109,6 +108,7 @@
     "preact-render-to-string": "^5.2.6",
     "preact-router": "^3.0.1",
     "puppeteer": "^17.1.3",
+    "rimraf": "^3.0.2",
     "sass": "^1.56.1",
     "sass-loader": "^12.4.0",
     "shelljs": "^0.8.5",
@@ -135,6 +135,6 @@
     }
   },
   "engines": {
-    "node": ">=12"
+    "node": ">=14.14.0"
   }
 }
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -80,7 +80,6 @@
     "style-loader": "^3.3.1",
     "terser-webpack-plugin": "^5.3.0",
     "typescript": "^4.6.4",
-    "update-notifier": "^5.1.0",
     "webpack": "^5.67.0",
     "webpack-bundle-analyzer": "^4.5.0",
     "webpack-dev-server": "^4.9.0",
--- a/packages/cli/src/index.js
+++ b/packages/cli/src/index.js
@@ -1,6 +1,8 @@
 #!/usr/bin/env node
-const envinfo = require('envinfo');
 const sade = require('sade');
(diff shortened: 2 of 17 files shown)


packages/cli/src/lib/webpack/webpack-base-config.js
```diff
@@ -329,7 +321,6 @@ module.exports = function createBaseConfig(config, env) {
 						},
 					},
 				}),
-			isProd && new webpack.LoaderOptionsPlugin({ minimize: true }),
```
rschristian, line 332: Super legacy, [exists to smooth over Webpack 1 -> 2 migration woes](https://v4.webpack.js.org/plugins/loader-options-plugin/)

We've just had it sat around forever.

packages/cli/src/lib/webpack/webpack-base-config.js
```diff
@@ -45,19 +45,6 @@ function resolveTsconfig(cwd, isProd) {
...
-		sassOptions: {
-			includePaths,
-		},
-	};
-
-	Object.defineProperty(config, 'includePaths', { value: includePaths });
-
-	return config;
-}
-
```
rschristian, line 60: [Exists to smooth over `sass-loader` 7 -> 8 migration](https://github.com/preactjs/preact-cli/pull/930#discussion_r362248387) (current version is v12)

packages/cli/src/index.js
```diff
@@ -118,3 +94,16 @@ prog.parse(process.argv, {
 		);
 	},
 });
+
+function exec(cmd) {
+	cmd.catch(catchExceptions);
+}
```
rschristian, line 100: Making an honest attempt to catch errors, rather than relying entirely on `process.on('unhandledRejection')` to catch everything.

packages/cli/src/index.js
```diff
@@ -18,19 +16,12 @@ if (
...
-process.on('unhandledRejection', err => {
-	error(err.stack || err.message);
-});
+const { build, info, watch } = require('./commands');
 
 const prog = sade('preact').version(pkg.version);
 
 prog
-	.command('build [src]')
+	.command('build')
```
rschristian, line 24: I think it's better to push users towards the flag; avoids conflicts and (potentially) obscure errors.

packages/cli/package.json
```diff
@@ -138,6 +135,6 @@
     }
   },
   "engines": {
-    "node": ">=12"
+    "node": ">=14.14.0"
```
rschristian, line 138: Need it for `rm(..., { recursive: true })` (to replace `rimraf`)

packages/cli/src/lib/webpack/prerender.js
```diff
@@ -37,7 +37,7 @@ module.exports = function (env, params) {
 			throw err;
 		}
 
-		handlePrerenderError(err, env, stack, entry);
+		await handlePrerenderError(err, env, stack, entry);
```
rschristian, line 40: Prerender errors result in a good ol' `process.exit(1)`, so it's better if we wait than have Webpack continue on even after we know the build is doomed.

PR #1754, opened 2022-12-15
fix: Exit status code for fatal Webpack exceptions

### LINKED ISSUE #1744: Failed builds return Exit Code 0
### Discussed in https://github.com/preactjs/preact-cli/discussions/1743

<div type='discussions-op-text'>

<sup>Originally posted by **jasonhilldm** October  2, 2022</sup>
I am using CircleCI to deploy an app when changes are pushed to git. One of the steps is the command "npm run build" which in turn runs "preact build".

When there is an error with the build, I am seeing a "Build failed!" message in the CircleCI step but it also says "CircleCI received exit code 0" and so continues with the next step of the deployment.

Any ideas why CircleCI is receiving an exit code 0 from the build step even though it has failed?

I see a very old issue about this same problem but it was fixed: https://github.com/preactjs/preact-cli/issues/61</div>

rschristian:

bugfix

No

**Summary**

Closes #1744 

We don't correctly handle fatal Webpack exceptions at the moment; the `err` in the callback for `compiler.run()` relates to fatal Webpack exceptions, like configuration errors. When one is hit, we only reject the promise, which actually allows the build to "complete" with a status code of `0`. We instead want to exit immediately with a status code `1` (as that matches what we do for compilation errors).

No

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,7 @@ jobs:
   test:
     name: Test
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     strategy:
       matrix:
         node-version: [12.x, 14.x]
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -127,7 +127,7 @@
     "preact": "^10.5.13",
     "preact-render-to-string": "^5.1.19",
     "preact-router": "^3.0.1",
-    "puppeteer": "^9.1.1",
+    "puppeteer": "^13.7.0",
     "sass": "^1.34.0",
     "sass-loader": "^10.2.0",
     "shelljs": "^0.8.3",
--- a/packages/cli/src/lib/webpack/run-webpack.js
+++ b/packages/cli/src/lib/webpack/run-webpack.js
@@ -78,13 +78,13 @@ async function prodBuild(env) {
 }
 
 function runCompiler(compiler) {
-	return new Promise((res, rej) => {
+	return new Promise(res => {
 		compiler.run((err, stats) => {
-			showStats(stats, true);
-
-			if (err || (stats && stats.hasErrors())) {
-				rej(`${red('\n\nBuild failed! \n\n')} ${err || ''}`);
+			if (err) {
+				error(err, 1);
 			}
+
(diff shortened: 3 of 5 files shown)


packages/cli/src/lib/webpack/run-webpack.js
```diff
@@ -78,14 +78,14 @@ async function prodBuild(env) {
...
-			showStats(stats, true);
-
-			if (err || (stats && stats.hasErrors())) {
-				rej(`${red('\n\nBuild failed! \n\n')} ${err || ''}`);
+			if (err) {
+				error(err, 1);
 			}
 
+			showStats(stats, true);
+
```
rschristian, line 88: `showStats()`, in production builds, will also immediately exit upon coming across any errors; as such, there's no reason to ever reject this promise as we've already exited.

.github/workflows/ci.yml
```diff
@@ -16,6 +16,7 @@ jobs:
   test:
     name: Test
     runs-on: ubuntu-latest
+    timeout-minutes: 10
```
rschristian, line 19: Unrelated, but for some (godforsaken reason), puppeteer randomly never exits on the CI. I've seen no pattern for this, just that it occurs.

GitHub's default is 360 minutes, or 6 hours, which we definitely don't need. I just caught a run still going after an hour and had to manually kill it. 10 minutes should be plenty of time.

rschristian:
Ugh, this works but I don't love that we're mixing throwing & `process.exit()` to signify errors. For end users it probably doesn't make too much of a difference but it does make our test suite a bit of a headache.

packages/cli/package.json
```diff
@@ -127,7 +127,7 @@
     "preact": "^10.5.13",
     "preact-render-to-string": "^5.1.19",
     "preact-router": "^3.0.1",
-    "puppeteer": "^9.1.1",
+    "puppeteer": "^13.7.0",
```
rschristian, line 130: This shouldn't be necessary, but our version of Puppeteer has turned *super* flakey recently. Granted it probably is a good few years old, so maybe that's unavoidable.

PR #1426, opened 2020-09-15
Better prerender crash message when error originates in node_modules/

VanTanev:
**What kind of change does this PR introduce?**

Improves the error message when prerender fails because of code that ran in `node_modules/`

Before:
image (17)

After:
image

**Did you add tests for your changes?**

No

**Summary**

This improves the prerender error message in cases like the following:
```jsx
import { createBrowserHistory } from 'history'

<App>
  <Router history={createBrowserHistory()}> // breaks prerendering
  // [...]
```

Fixed code:
```jsx
import { createBrowserHistory, createMemoryHistory } from 'history'

<App>
  <Router history={
    typeof window !== 'undefined' 
        ? createBrowserHistory()
        : createMemoryHistory()
  }> 
  // [...]
```

**Does this PR introduce a breaking change?**

No

**Other information**

```
Environment Info:
  System:
    OS: Linux 4.15 Ubuntu 16.04.7 LTS (Xenial Xerus)
    CPU: (16) x64 Intel(R) Core(TM) i9-9900KS CPU @ 4.00GHz
  Binaries:
    Node: 14.8.0 - ~/.nvm/versions/node/v14.8.0/bin/node
    Yarn: 1.21.1 - /usr/bin/yarn
    npm: 6.14.7 - ~/.nvm/versions/node/v14.8.0/bin/npm
  Browsers:
    Chrome: 85.0.4183.102
    Firefox: 80.0.1
  npmPackages:
    preact: ^10.4.8 => 10.4.8 
    preact-cli: ^3.0.1 => 3.0.1 
    preact-render-to-string: ^5.1.10 => 5.1.10 
    preact-router: ^3.2.1 => 3.2.1 
```

PR #1756, opened 2022-12-17
fix: Improve prerender errors

rschristian:

Bugfix / revert

N/A

**Summary**

TLDR:

Before PR               | Improved Error Logic | Improved Error Logic & Disabled SSR Build Minification
:-------------------------:|:-------------------------:|:-------------------------:
Before this PR | With Error Logic Altered | Disabled SSR Build Minification

Partial revert of #1426

Having used it for a bit, I don't find it to be particularly compelling, especially w/ Webpack v5 -- `methodName` is often null and not correct enough for determining user vs lib code.

Additionally, I've found the stack trace locations (from Node / v8, to be clear) to be exceptionally poor on Webpack's minimized output, and so have elected to not minimize our SSR build.

Take the following:

> src/routes/home/index.js

```jsx
import { h } from 'preact';
import style from './style.css';

console.log(foo.bar);

const Home = () => (
	<div class={style.home}>
		<h1>Home</h1>
		<p>This is the Home component.</p>
	</div>
);

export default Home;
```

As `foo` is not defined, we should get a (reference) error here (and we do). However, the stack trace of `ssr-bundle.js` actually points to the `h("div", { class: ... }, ...)`.

Why this is, I don't know. Mix of Webpack's output and Node maybe? Regardless, it's not great and reduces the quality of our error reporting (as it highlights the wrong line). But what we can do is just not minimize the SSR output. As it's only ran in Node, we don't have any particular reason _to minimize_, and a couple against, such as:
  - faster builds of the SSR bundles as minification takes time
  - better error positions if prerender errors occur

No


packages/cli/src/lib/webpack/prerender.js
```diff
@@ -45,112 +45,98 @@ async function handlePrerenderError(err, env, stack, entry) {
...
-					let line = position.line + i;
-					let sourceLine = sourceLines[line - 1];
-					sourceCodeHighlight += sourceLine ? `${color(sourceLine)}\n` : '';
-				}
+		if (sourceLines) {
+			for (var i = -4; i <= 4; i++) {
+				let color = i === 0 ? red : yellow;
+				let line = position.line + i;
+				let sourceLine = sourceLines[line - 1];
+				sourceCodeHighlight += sourceLine ? `${color(sourceLine)}\n` : '\n';
```
rschristian, line 95: Rather importantly, we weren't handling blank lines correctly as the ternary was falling back to `''`, instead of the new line as it does above. This caused further issues in line positioning.

PR #1758, opened 2022-12-20
feat: Improved errors and warnings

rschristian:

Refactor

Existing should cover

**Summary**

Refactors our error handling for better internal consistency and improves error/warning messages for users.

Previously, we quite heavily used `process.exit()` to react to any compilation errors from Webpack. While this is perfectly usable for end users, it creates a bit of a problem internally for our test suite; if we're mixing throwing errors and `process.exit()`, it becomes a bit of a challenge to ensure our builds are completing as they should. At best, it's extra boilerplate for each test that relies on completion w/out errors (test needs to now throw and we need to spy on `process.exit()`), at worst, tests are passing that shouldn't.

So, this PR looks to remove those `.exit()` calls and instead simply throw errors; these errors are still caught at the CLI root (which is how our users will interact with `preact-cli`) but we can detect from the commands themselves (`build()`, `watch()`, etc.)

Additionally, with Webpack v5 and Dev Server v4, Webpack's messages in dev are _much_ improved, removing the need for use to handle them ourselves. Here's an example from before this PR:

 Warning             |  Error
:-------------------------:|:-------------------------:
Warning in Dev | Error in Dev

Our messages are first in both cases, with Webpack's own coming second. As you can see, while the warnings are interchangeable, in the case of errors, our messages are a bit lacking; the in-built error messaging conveniently returns the error position to the end user.

As such, I see no reason for us to continue handling warning/errors in dev, instead, let's use the (greatly improved) built-ins. I've also adopted this structure for our own messages, so in `build`, if a Webpack compilation error is thrown, it will resemble the error structure from above (module name + line of code, with the full message on the following line).

I don't believe so

### DIFF
--- a/packages/cli/src/lib/webpack/prerender.js
+++ b/packages/cli/src/lib/webpack/prerender.js
@@ -5,7 +5,7 @@ const stackTrace = require('stack-trace');
 const URL = require('url');
 const { SourceMapConsumer } = require('source-map');
 
-module.exports = function (env, params) {
+module.exports = async function (env, params) {
 	params = params || {};
 
 	let entry = resolve(env.dest, './ssr-build/ssr-bundle.js');
@@ -37,7 +37,7 @@ module.exports = function (env, params) {
 			throw err;
 		}
 
-		handlePrerenderError(err, env, stack, entry);
+		await handlePrerenderError(err, env, stack, entry);
 	}
 };
 
--- a/packages/cli/src/lib/webpack/run-webpack.js
+++ b/packages/cli/src/lib/webpack/run-webpack.js
@@ -2,12 +2,12 @@ const ip = require('ip');
 const webpack = require('webpack');
 const { resolve } = require('path');
 const clear = require('console-clear');
-const { bold, red, green, magenta } = require('kleur');
+const { bold, green, magenta } = require('kleur');
 const DevServer = require('webpack-dev-server');
 const clientConfig = require('./webpack-client-config');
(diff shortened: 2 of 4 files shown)


packages/cli/src/lib/webpack/prerender.js
```diff
@@ -37,7 +37,7 @@ module.exports = function (env, params) {
 			throw err;
 		}
 
-		handlePrerenderError(err, env, stack, entry);
+		await handlePrerenderError(err, env, stack, entry);
```
rschristian, line 40: The end result of a prerender error is a good ol' `process.exit(1)`, so we want to block Webpack if we're at this point; there's no point in allowing it to continue whilst we track down the error position.

packages/cli/src/lib/webpack/run-webpack.js
```diff
@@ -69,100 +65,81 @@ async function prodBuild(config, env) {
...
 		}
 
-		if (stats.hasWarnings()) {
-			allFields(stats, 'warnings')
-				.map(stripLoaderPrefix)
-				.forEach(({ message }) => warn(message));
+		if (stats.hasErrors()) {
+			allFields(stats, 'errors').forEach(err => {
+				throw err;
+			});
```
rschristian, line 104: I switched the order of warning & error reporting here; as we bail on errors, a user would need an error-less build in order to be alerted of any warnings which I think is less than ideal.

If they're alerted to warnings first, they can fix those alongside whatever error they're experiencing.

PR #1761, opened 2022-12-22
feat: Support proxied loaders w/ `getLoadersByName()` is user config

### LINKED ISSUE #929: getLoadersByName fails to resolve certain loaders
A bug I guess, or if this is an intended behavior a feature.

I'm using the `sass-loader` in combination with `node-sass` and want to modify the loader-config.
Using `helper.getLoadersByName(config, 'sass-loader')` returns an empty array.

```bash
$ preact create simple [name]
```
... install and update dependencies which at the end looks like this (snipped):
```json
{
    "devDependencies": {
        "eslint": "^6.7.2",
        "eslint-config-synacor": "^3.0.5",
        "if-env": "^1.0.4",
        "node-sass": "^4.13.0",
        "preact-cli": "^2.2.1",
        "sass-loader": "^7.3.1"
    },
    "dependencies": {
        "preact": "^10.1.1",
        "preact-render-to-string": "^5.1.2"
    }
}
```
Create a `preact.config.js` with the following content:
```js
    const sassLoader = helpers.getLoadersByName(config, 'sass-loader');
    console.log(sassLoader);
};
```
When I start it via `npm run dev` (`preact watch`) I'll receive an empty array:
image

I'll get an array with the `sass-loader`-object inside of it.

I've managed to find a work-around:
```js
const findLoaderByName = (rules, name) => {

    for (const {loaders} of rules) {
        for (const loader of loaders) {
            const loaderName = loader.loader;

            if (loaderName === 'proxy-loader') {
                if(loader.options.loader === name) {
                    return loader.options;
                }
            } else if (loaderName === name) {
                return loader;
            }
        }
    }

    return null;
};

    const sassLoader = findLoaderByName(helpers.getLoaders(config), 'sass-loader');
    // sassLoader is now resolved
    console.log(sassLoader);
};
```

```
Unknown argument: info
```
> The Preact-CLI version is `2.2.1` if you want that.

rschristian:

Feature / bugfix

No

**Summary**

Closes #929

If the loader is being proxied (like say `sass-loader`), we need to look a bit deeper then we normally would to get the "correct" loader.

The interface here is a bit messy though, might be a bit confusing to users *why* there's this proxy loader:

```json
[
    {
        "rule": {
            "enforce": "pre",
            "test": "\.s[ac]ss$",
            "use": [
                {
                    "loader": "/home/ryun/Projects/foobar/node_modules/preact-cli/src/lib/webpack/proxy-loader.js",
                    "options": {
                        "cwd": "/home/ryun/Projects/foobar",
                        "loader": "sass-loader",
                        "options": {
                            "sourceMap": true,
                            "sassOptions": {
                                "includePaths": [
                                    "/home/ryun/Projects/foobar/node_modules",
                                    "/home/ryun/Projects/foobar/node_modules/preact-cli/node_modules"
                                ]
                            }
                        }
                    }
                }
            ]
        },
        "ruleIndex": 2,
        "loader": {
            "loader": "/home/ryun/Projects/foobar/node_modules/preact-cli/src/lib/webpack/proxy-loader.js",
            "options": {
                "cwd": "/home/ryun/Projects/foobar",
                "loader": "sass-loader",
                "options": {
                    "sourceMap": true,
                    "sassOptions": {
                        "includePaths": [
                            "/home/ryun/Projects/foobar/node_modules",
                            "/home/ryun/Projects/foobar/node_modules/preact-cli/node_modules"
                        ]
                    }
                }
            }
        },
        "loaderIndex": 0
    }
]
```

No

### DIFF
--- a/packages/cli/src/lib/webpack/transform-config.js
+++ b/packages/cli/src/lib/webpack/transform-config.js
@@ -239,11 +239,13 @@ class WebpackConfigHelpers {
 					: [{ rule, ruleIndex, loader: loaders, loaderIndex: -1 }]
 			)
 			.reduce((arr, loaders) => arr.concat(loaders), [])
-			.filter(
-				({ loader }) =>
-					(typeof loader === 'string' && loader.includes(name)) ||
-					(typeof loader.loader === 'string' && loader.loader.includes(name))
-			);
+			.filter(({ loader }) => {
+				if (typeof loader === 'string') return loader.includes(name);
+				return typeof loader.loader === 'string' &&
+					loader.loader.includes('proxy-loader')
+					? loader.options.loader.includes(name)
+					: loader.loader.includes(name);
+			});
 	}
 
 	/**
--- a/.changeset/flat-ducks-invent.md
+++ b/.changeset/flat-ducks-invent.md
@@ -1,0 +1,5 @@
+---
+'preact-cli': patch
+---
+
+Support proxied loaders w/ `getLoadersByName()` method in users' `preact.config.js`


packages/cli/src/lib/webpack/transform-config.js
```diff
@@ -239,11 +239,13 @@ class WebpackConfigHelpers {
...
 			)
 			.reduce((arr, loaders) => arr.concat(loaders), [])
-			.filter(
-				({ loader }) =>
-					(typeof loader === 'string' && loader.includes(name)) ||
-					(typeof loader.loader === 'string' && loader.loader.includes(name))
-			);
+			.filter(({ loader }) => {
+				if (typeof loader === 'string') return loader.includes(name);
+				return typeof loader.loader === 'string' &&
```
rschristian, line 244: Not 100% that we need this type check here, but Webpack's types are quite loose and I don't want to break anyone's setup by changing behavior.

Issue #1561, opened 2021-05-17
Lack of documentation around adding meta tags info in template.html


sharat9211:
My requirement was to add meta tags, so that when sharing links on twitter/fb/whatsapp a link preview should appear. I needed to add meta tags for this purpose. There is not a lot of documentation around that. Two things which are mentioned are preact-cli uses `ejs` and template.html has `preact.title` for dynamically showing title. Other than that, I couldn't find anything. I was able to figure out a solution by looking at the source code of the preact documentation site, where something like this is being used `<%= htmlWebpackPlugin.options.CLI_DATA.preRenderData.description %>`. 
So my questions are :
- Is this the recommended way to add meta tags in a prect-cli project?
- If no then what is the recommended way?
- If yes then, is there any other data is useful which is being exposed like this and where should I look for that(this is for my own learning)

Also, I am willing to add the necessary documentation for this if  but advance warning I don't have a lot of knowledge on html-webpack-plugin.

PS: 
Apologies if I have missed some obvious place which has the documentation
And thanks for making this really useful framework

rschristian:
Sorry, could use some extra docs there, but feel free to add.

If you look, preact-www uses `preact build --prerenderUrls src/prerender.js ...` as its build command, which leads [here}(https://github.com/preactjs/preact-www/blob/master/src/prerender.js). 

`preRenderData.description` isn't something that is exposed automatically, as it's coming entirely from that prerender script. 

You can make your own script to generate whatever values you need.

PR #1768, opened 2023-01-05
refactor: Move EJS into user templates (`template.html`)

rschristian:
### What kind of change does this PR introduce?

Refactor

### Did you add tests for your changes?

Existing cover

### Summary

Moves `head-end.ejs` & `body-end.ejs` into user templates.

Originally, these were abstracted out in #862 to allow us to make alterations to the templates without breaking changes, but I don't think we ever really leaned into this. #1608, for example, was put behind a major as the API in the templates changed so drastically; almost every user with a custom template would've needed to make an adjustment (or a few). As such, the API contract never really shifted from `html-webpack-plugin`'s template parameters.

Therefore, I believe continuing to have these abstracted out is doing more harm than good. We've received numerous issues & discussions both here and on Slack from users asking how to make simple alterations, like [adding a meta tag](https://github.com/preactjs/preact-cli/issues/1561) or a [attribute on a script tag](https://github.com/preactjs/preact-cli/discussions/1766). In the case of adding an attribute to script tags, the user would need to know to hop into `preact-cli`'s source and find `body-end.ejs` to copy it into their template. That's not an ideal experience, even if we had documentation covering that.

By moving the EJS templates into users' projects (and switching the file extension to `.ejs`), my hope is that it'll be much clearer to users how they can customize this behavior. We'll still need a wiki page (reminder to myself) going over some things like the `cli` and `htmlWebpackPlugin` keys that are accessible for the power users, but immediate transparency should improve (I hope).

Additionally, the work done in #1466 went a long way towards making this change possible and palatable. The EJS is far simpler now and won't give new users a scare, which is awesome.

### Does this PR introduce a breaking change?

Yes, `<% preact.headEnd %>` and `<% preact.bodyEnd %>` will no longer be supported.

If they still exist in the user's template, the compilation will error out and they'll be provided with a link for the new default template. While we could replace with empty strings, that's hardly a successful compilation as it'll be unusable; there will be no links to any resources in the HTML.

They'll need to copy from the default template that will ship with v4 or delete theirs if they have no customization made, as `preact-cli` will use the default if a user's cannot be found.


packages/cli/src/lib/webpack/render-html-plugin.js
```diff
@@ -78,29 +98,44 @@ module.exports = async function renderHTMLPlugin(config, env) {
...
+				if (compilationAssets.includes('es-polyfills.js')) {
+					entrypoints['es-polyfills'] = assets.publicPath + 'es-polyfills.js';
+				}
+
 				return {
 					cli: {
 						title,
 						url,
 						manifest: config.manifest,
-						inlineCss: config['inlineCss'],
```
rschristian, line 96: Already available as `config.inlineCss`, no point in having a duplicated top level key for it.

PR #1770, opened 2023-01-07
fix: Ensure devServer info message uses actual devServer config

rschristian:

Refactor

No

**Summary**

Essentially, we are normalizing CLI options way too late, causing inconsistencies.

As the output message didn't consume `webpack.devServer`'s config, there were some situations in which the two wouldn't match up: an example of this being `HTTPS=true yarn preact watch`. We consumed `process.env.HTTPS` in the informational text, but not for the actual dev server config. The user would see an error if they tried to open the URL we provided in their browser.

**Other information**

### DIFF
--- a/packages/cli/src/commands/build.js
+++ b/packages/cli/src/commands/build.js
@@ -8,7 +8,6 @@ exports.build = async function buildCommand(src, argv) {
 	argv.src = src || argv.src;
 	// add `default:true`s, `--no-*` disables
 	argv.prerender = toBool(argv.prerender);
-	argv.production = toBool(argv.production);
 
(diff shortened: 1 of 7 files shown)


packages/cli/src/commands/build.js
```diff
@@ -8,7 +8,6 @@ exports.build = async function buildCommand(src, argv) {
 	argv.src = src || argv.src;
 	// add `default:true`s, `--no-*` disables
 	argv.prerender = toBool(argv.prerender);
-	argv.production = toBool(argv.production);
```
rschristian, line 11: `production` is not a valid CLI flag, so it's been moved into `run-webpack.js` with the other derived env values.

PR #1772, opened 2023-01-11
Refactor: Enable Prefresh by Default

rschristian:

Refactor

No

**Summary**

Prefresh is pretty neat and HMR is now the standard for build tooling. We should enable it by default else some might assume we don't support it.

Yes, enables `--refresh` by default now.

Users can still disable with `--no-refresh` or `--refresh false` if they'd prefer live reload.

### DIFF
--- a/.changeset/lucky-lizards-drive.md
+++ b/.changeset/lucky-lizards-drive.md
@@ -1,0 +1,5 @@
+---
+'preact-cli': major
+---
+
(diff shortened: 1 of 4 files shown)


rschristian:
The changes to the test suite aren't related to this change -- I had previously and separately altered the templates which means our tests here needed to be adjusted.

PR #1783, opened 2023-02-05
test: Refactor snapshot testing

rschristian:

Testing refactor

Yes

**Summary**

The existing snapshot tests were quite time consuming to update, as they weren't making use of Jest's snapshots. Every time a file hash changed or output deviated beyond the +/- 5% bound, the contributor would need to manually update. When a few hashes or output sizes changed, this was a slow and annoying process.

This PR switches over to using actual snapshots, though there is a downside: Jest has no "Snapshot sorta matches" functionality, so if there's any deviation at all in a file size, we need to update. I don't see this as too much of an issue though, as it's so quick to update (simply append `-u` to the test command, and Jest itself informs you of this). Also, it's a nice canary for changes that shouldn't have changed output sizes at all -- in the past, anything within +/- 5% would be hidden.

This will greatly reduce the time I spend updating tests after I've determined a change should be made to the stored images.

N/A


rschristian:
Test failure is due to differing paths in comments in SSR bundle. Need to think on that, don't really want to do too much fiddling with bundles to normalize, but we might need to.

Issue #1793, opened 2023-06-29
CSS source maps missing in build output


teodragovic:
**What is the current behaviour?**
- Running `preact build` produces build with source maps for JS files but no source map for CSS
- Behavior is same for the project I'm working on and clean install and using either CSS or SASS
- I see `sourceMap` option set to `true` in webpack config for all loaders but also `css.map` missing in tests (otherwise they would fail) so not sure if bug or per design?
- CSS source map is included in `ssr-build` though

**What is the motivation or use case for changing this behaviour?**
We pushed a big redesign to production and some users are reporting page being broken due to CSS not loading. We suspect parsing error so having source map would be helpful to run down the issue.

**Describe the solution you'd like**
Output CSS source maps in `build` by default or provide options/recipe how to enable

Please paste the results of `npx preact-cli info` here.
```
Environment Info:
  System:
    OS: Windows 10 10.0.19044
    CPU: (16) x64 Intel(R) Core(TM) i7-10700F CPU @ 2.90GHz
  Binaries:
    Node: 16.18.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - C:\Program Files\nodejs\yarn.CMD
    npm: 8.19.3 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (114.0.1823.58)
  npmPackages:
    preact: 10.15.1 => 10.15.1
    preact-cli: ^3.3.0 => 3.4.1
    preact-render-to-string: ^5.1.19 => 5.2.6
    preact-router: ^3.2.1 => 3.2.1
```

rschristian:
This looks to be fixed in `@next`, but will take a look a bit later for v3

teodragovic:
@rschristian thanks, good to know. I poked around some and found the issue in `OptimizeCssAssetsPlugin` setting which should be set as described [here](https://github.com/NMFR/optimize-css-assets-webpack-plugin/issues/81).

rschristian:
Released as v3.4.6

PR #1794, opened 2023-06-30
fix: output css source map

teodragovic:

bugfix

no

**Summary**

https://github.com/preactjs/preact-cli/issues/1793

no


teodragovic:
Is filesize some random value or should I produce build to get it? If latter, how do I do it? Just run clean installation?

rschristian:
You'll need to run a build to get it. The test ensures that file is +/5% of the listed value.

Unfortunately it's a bit of a pain in the ass to generate, could never figure out a way to automate it. Easiest way is to only run the one build test and get the file size from the directory that's created.

rschristian:
I'll take care of it in a few minutes, it's a faff

rschristian (approved):
Thanks for catching this, tracking down the solution, AND implementing! I really appreciate it!

PR #1818, opened 2024-01-27
feat: update ForkTsCheckerWebpackPlugin which presented random lockin…

### LINKED ISSUE #1813: build gets stuck at "after chunk asset optimization"
Recently I have updated to the latest (stable) 3.5.0 version of Preact CLI and I started facing an issue where build would randomly (i mean not on every build) get stuck at this phase:

image

then, if I stop the build process and start it again, it would most likely finish without any issues.

So far I haven't been able to find any clues on when this happens, except for the fact that it seems more likely to happen when the ide has been opened for several hours, and less prone to happen on first builds. But let's not count on this, it could also be a suggestion.

Anyways, is there anything I could do just to get more detailed information which might help to pinpoint or debug more thoroughly the issue?

```
npx preact-cli info

Environment Info:
  System:
    OS: Windows 10 10.0.19045
    CPU: (16) x64 AMD Ryzen 7 3700X 8-Core Processor
  Binaries:
    Node: 20.11.0 - ~\AppData\Local\nvs\default\node.EXE
    npm: 10.2.4 - ~\AppData\Local\nvs\default\npm.CMD
  Browsers:
    Chrome: 120.0.6099.217
    Edge: Chromium (120.0.2210.133)
  npmPackages:
    preact: ^10.3.1 => 10.6.6
    preact-cli: ^3.5.0 => 3.5.0
```

krskrs:
update ForkTsCheckerWebpackPlugin which presented random locking issues with newest 4.x TS versions (fixes #1813 )

* update ForkTsChecker configuration format
* add ForkTsChecker plugin to the excluded plugins list of BabelEsmPlugin (the latter is directly calling apply on all the other plugins, causing unexpected behavior to ForkTsChecker)

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -64,7 +64,7 @@
     "envinfo": "^7.8.1",
     "esm": "^3.2.25",
     "file-loader": "^6.2.0",
-    "fork-ts-checker-webpack-plugin": "^4.0.4",
+    "fork-ts-checker-webpack-plugin": "^6.5.3",
     "get-port": "^5.0.0",
     "gittar": "^0.1.0",
     "glob": "^8.0.3",
--- a/packages/cli/src/lib/webpack/webpack-base-config.js
+++ b/packages/cli/src/lib/webpack/webpack-base-config.js
@@ -358,7 +358,13 @@ module.exports = function createBaseConfig(env) {
 			...(tsconfig
 				? [
 					new ForkTsCheckerWebpackPlugin({
-						typescript: {configFile: tsconfig, diagnosticOptions: {semantic: true, syntactic: true} },
+						typescript: {
+							configFile: tsconfig,
+							diagnosticOptions: {
+								semantic: true,
+								syntactic: true
+							}
+						},
 						async: !isProd
 					}),
 				  ]
--- a/packages/cli/src/lib/webpack/webpack-base-config.js
+++ b/packages/cli/src/lib/webpack/webpack-base-config.js
@@ -357,12 +357,10 @@ module.exports = function createBaseConfig(env) {
 			}),
 			...(tsconfig
 				? [
-						new ForkTsCheckerWebpackPlugin({
-							checkSyntacticErrors: true,
-							async: !isProd,
-							tsconfig: tsconfig,
-							silent: !isWatch,
-						}),
+					new ForkTsCheckerWebpackPlugin({
+						typescript: {configFile: tsconfig, diagnosticOptions: {semantic: true, syntactic: true} },
+						async: !isProd
+					}),
 				  ]
 				: []),
 			...(isProd
--- a/packages/cli/src/lib/webpack/webpack-client-config.js
+++ b/packages/cli/src/lib/webpack/webpack-client-config.js
@@ -170,7 +170,7 @@ function getBabelEsmPlugin(env) {
 			new BabelEsmPlugin({
 				filename: env.isProd ? '[name].[chunkhash:5].esm.js' : '[name].esm.js',
 				chunkFilename: '[name].chunk.[chunkhash:5].esm.js',
-				excludedPlugins: ['BabelEsmPlugin', 'InjectManifest'],
+				excludedPlugins: ['BabelEsmPlugin', 'InjectManifest', 'ForkTsCheckerWebpackPlugin'],
 				beforeStartExecution: plugins => {
 					plugins.forEach(plugin => {
 						if (
--- a/yarn.lock
+++ b/yarn.lock
@@ -23,7 +23,7 @@
   dependencies:
     "@babel/highlight" "^7.10.4"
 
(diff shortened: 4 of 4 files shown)


rschristian:
Sorry, going to close & reopen real quick to see if I can trigger the CI

packages/cli/src/lib/webpack/webpack-base-config.js
```diff
@@ -357,12 +357,10 @@ module.exports = function createBaseConfig(env) {
...
 			...(tsconfig
 				? [
-						new ForkTsCheckerWebpackPlugin({
-							checkSyntacticErrors: true,
-							async: !isProd,
-							tsconfig: tsconfig,
-							silent: !isWatch,
-						}),
+					new ForkTsCheckerWebpackPlugin({
+						typescript: {configFile: tsconfig, diagnosticOptions: {semantic: true, syntactic: true} },
```
rschristian, line 361: `semantic: true` is not something we want to set. 

Additionally, this should be formatted. We use precommit hooks via Husky to do this, generally.

packages/cli/package.json
```diff
@@ -64,7 +64,7 @@
     "envinfo": "^7.8.1",
     "esm": "^3.2.25",
     "file-loader": "^6.2.0",
-    "fork-ts-checker-webpack-plugin": "^4.0.4",
+    "fork-ts-checker-webpack-plugin": "^6.5.3",
```
rschristian, line 67: There should likely be a change in the repo's lock file to accompany this.

packages/cli/src/lib/webpack/webpack-base-config.js
krskrs: > `semantic: true` is not something we want to set.
> 
I just made it explicit, but `semantic: true` is already the default also in the forkstsc 7.x branch used in pract-cli 4.0.
Just wanted to make sure you are aware of this, and you're just meaning that it's not needed to make it explicit, and we can leave it to the default value.


> Additionally, this should be formatted. We use precommit hooks via Husky to do this, generally.
I'm not familiar with Husky, but I will look into it 😄

rschristian: > I just made it explicit, but semantic: true is already the default also in the forkstsc 7.x branch used in pract-cli 4.0.

Well, no, you've changed the option there. While I decided to change that in the new major, that should not be backported (IMO) to 3.x

>  I'm not familiar with Husky, but I will look into it 😄

It's just a precommit hook. I imagine you didn't clone the repo locally to make these changes? You can also format manually using our prettier config.

packages/cli/package.json
krskrs: > There should likely be a change in the repo's lock file to accompany this.

uhm, are you referring to yarn.lock? in such case, my bad! 😄 I only use npm and I totally missed it! Could you briefly tell me how such file it's supposed to be updated and I will look into it as well. (it looks some sort of autogenerated file and not just a matter of updating the version like it's for package.json)

rschristian: Just run `yarn` / `yarn install`. It's the same as a `package-lock.json`.

packages/cli/src/lib/webpack/webpack-base-config.js
krskrs: > Well, no, you've changed the option there. While I decided to change that in the new major, that should not be backported (IMO) to 3.x
> 

uhm, I get your point, but we should already be ok then, please see here https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/blob/v4.1.6/README.md (this is the version you were using before). From what I can understand, the old version is already enabling semantic checks by default? 
I quote from their README (```checkSyntacticErrors boolean: This option is useful if you're using ts-loader in happyPackMode with [HappyPack](https://github.com/amireh/happypack) or [thread-loader](https://github.com/webpack-contrib/thread-loader) to parallelise your builds. If true it will ensure that the plugin checks for both syntactic errors (eg const array = [{} {}];) and semantic errors (eg const x: number = '1';). By default the plugin only checks for semantic errors. This is because when ts-loader is used in transpileOnly mode, ts-loader will still report syntactic errors. When used in happyPackMode it does not. Default: false.``` It's a bit tricky to districate between changes in other plugins, defaults, etc.etc. What do you think? 

> > I'm not familiar with Husky, but I will look into it 😄
> 
> It's just a precommit hook. I imagine you didn't clone the repo locally to make these changes? You can also format manually using our prettier config.
Well, I cloned it, but didn't install anything. Now it's all clear (should be 😄 )

rschristian: >  From what I can understand, the old version is already enabling semantic checks by default?

You're right, my bad!

rschristian:
Thanks!

rschristian:
Published as `3.5.1`

krskrs:
Thank to you for considering my pull request. But are there some issues with tests? I looked at the test output here on github, but I couldn't clearly find what's exactly failing.

rschristian:
The tests for this PR passed, not sure what's up with the push to 3.x.x but it's probably nothing.

